PR #1, opened 2018-11-18
Fix docstring typos

jekbradbury:


jrevels:
Thanks! Too much copy-pasting between commits 😛

PR #3, opened 2019-03-21
Implemented conjugate and multiply for Wirtinger

simeonschaub:
My attempt at implementing `conj` for Wirtinger and `mul` for two Wirtinger derivatives. Please let me know if I was missing something as I'm still not too familiar with all the internal structure.

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -76,7 +76,7 @@ eval(defs)
 #####
 ##### `Wirtinger`
 #####
-# TODO does this represent (∂f/∂z, ∂f/∂z̄) or (∂f/∂z, ∂f̄/∂z) or what?
+# represents (∂f/∂z, ∂f/∂z̄)
 
 struct Wirtinger{P,C} <: AbstractDifferential
     primal::P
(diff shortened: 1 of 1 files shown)


ararslan:
Note that `*(::Wirtinger, ::Wirtinger)` is deliberately unsupported. See https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/differentials.jl#L130-L146.

simeonschaub:
I can't quite understand the rationale behind this. Composition of two Wirtinger derivatives seems like such a common thing to want to do. Non-commutative multiplication doesn't seem unreasonable to me at all, what issues exactly would this cause? At least theoretically this should definitely be realizable, as everything is still associative.

jrevels:
Thanks for the PR; sorry I didn't have most stuff up on master before you spent time on the multiplication thing.

> Composition of two Wirtinger derivatives seems like such a common thing to want to do. 

Yes, of course. ChainRules supports this, it just doesn't use `*` to do so (it does so via the `WirtingerRule`).

> Non-commutative multiplication doesn't seem unreasonable to me at all, what issues exactly would this cause?

Silently incorrect derivatives, unfortunately. 

The actual problem is more of a programming interface problem than a mathematical problem - you can always "pick a convention" for the local `*` definition, the question is "is it easy/possible for generic upstream code (i.e. code that may not know Wirtingers exist) to know + respect the convention I've chosen." If generic code assumes commutativity for `*` (which is does for real-valued scalar code), and you defined `*` in this manner, then you'd also need some extra contextual information at every `*` callsite that tells you the order in which the derivatives should be composed. We could just define a `compose` function that documents a generic ordering to the arguments, regardless of type, but that gets tricky to extend to non-scalar stuff.

It's an interesting issue, I have some notes I can post later if it helps. Would definitely be interested in any thoughts/solutions you might have.

PR #6, opened 2019-04-02
Add basic documentation

ararslan:
This is mostly copying what's in comments in the code into actual doc files along with the existing docstrings. I've also added a couple of basic docstrings.


src/rules.jl
```diff
@@ -121,8 +185,28 @@ end
...
+"""
+    frule(f, xs...)
+
+Apply the forward-mode differentiation rule to `f` with the given arguments `xs`,
+returning a tuple of `f(xs...)` and an [`AbstractRule`](@ref) object which can be
+called to evaluate the rule. If no forward-mode rule has been defined for `f`,
+`nothing` is returned.
+
+See also: [`rrule`](@ref)
+"""
```
jrevels: `frule` returns a tuple of `AbstractRule`s - one for each of the function's outputs. 

Similarly, `rrule` returns a tuple of `AbstractRule`s, one for each of the function's inputs.

If the returned tuple would only contain a single `AbstractRule`, the single `AbstractRule` may be returned directly (basically just matching Julia's `return` behavior).

docs/src/rules.md
```diff
@@ -0,0 +1,74 @@
...
+    r = frule(f, x)
+    if r === nothing
+        fx = f(x)
+        dx = Rule(Δx -> ForwardDiff.derivative(f, x) * Δx)
+    else
+        fx, dx = r
+    end
+    return fx, dx
+end
+```
```
jrevels: This kind of thing definitely belongs in an advanced usage section :)

jrevels:
Thanks for kicking the tires here! I'll add some commits to this branch as well; I've settled on a notation for certain parts of these, so I'll update `frule`/`rrule` docstrings to use that

PR #14, opened 2019-04-12
Initial finite differencing testing

willtebbutt:
This is some initial work to set up systematic testing for `frule`s and `rrule`s. This initial work covers:

- Some utility functionality to make running finite-differencing straightforward. This depends on FDM.jl
- I've extended `accumulate!` very slightly to accomodate scalars. @jrevels I could use your input here, this might be totally inconsistent with what you have in mind.
- I've moved some tests around and established a nested testset with 1-1 mapping between `src` and `test` files.
- All `frule`s and `rrule`s in `linalg` are now covered as a demonstration of the functionality.

There are a few things that need to be addressed before I go about adding extra finite-differencing test-coverage (in a later PR) / before this could be merged:

- What have I got structurally wrong in this PR? i.e. have I missed the point with everything.
- What is missing? i.e. if we extended the types of tests in `test/linalg.jl` to cover the entire codebase, what still wouldn't be covered? The `test_adjoint!` function was already here when I started, so I've kept that. Does it need extending? Do we need a forwards-mode equivalent?
- There are some tests left in `runtests.jl` that I've wrapped in a `Misc Tests` testset. These would ideally go somewhere else. I'm not entirely sure where the optimal location is though and could do with some advice.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ julia = "^1.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
 
 [targets]
-test = ["Test"]
+test = ["Test", "FDM"]
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,6 @@ uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 
 [deps]
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
-FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 
 [deps]
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
+FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
@@ -12,8 +13,8 @@ Cassette = "^0.2"
 julia = "^1.0"
 
 [extras]
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["Test", "FDM"]
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 
 [compat]
 Cassette = "^0.2"
+FDM = "^0.4.0"
 julia = "^1.0"
 
 [extras]
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -65,7 +65,8 @@ TODO
 function accumulate!(Δ, rule::AbstractRule, args...)
     return materialize!(Δ, broadcastable(add(cast(Δ), rule(args...))))
 end
-accumulate!(Δ::Real, rule::AbstractRule, args...) = accumulate(Δ, rule, args...)
+
+accumulate!(Δ::Number, rule::AbstractRule, args...) = accumulate(Δ, rule, args...)
 
 """
 TODO
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -245,7 +245,7 @@ as `Ω`, return the tuple:
 
 where each returned propagation rule `rule_for_Δxᵢ` can be invoked as
 
-    rule_for_Δxᵢ(previous_Δxᵢ, ΔΩ₁, ΔΩ₂, ...)
+    rule_for_Δxᵢ(ΔΩ₁, ΔΩ₂, ...)
 
 to yield `xᵢ`'s corresponding differential `Δxᵢ`. To illustrate, if all involved
 values are real-valued scalars, this differential can be written as:
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -62,7 +62,10 @@ accumulate(Δ, rule::AbstractRule, args...) = add(Δ, rule(args...))
 """
 TODO
 """
-accumulate!(Δ, rule::AbstractRule, args...) = materialize!(Δ, broadcastable(add(cast(Δ), rule(args...))))
+function accumulate!(Δ, rule::AbstractRule, args...)
+    return materialize!(Δ, broadcastable(add(cast(Δ), rule(args...))))
+end
+accumulate!(Δ::Real, rule::AbstractRule, args...) = accumulate(Δ, rule, args...)
 
 """
 TODO
--- a/src/rules/linalg.jl
+++ b/src/rules/linalg.jl
@@ -30,8 +30,8 @@ end
 
 function rrule(::typeof(inv), x::AbstractArray)
     Ω = inv(x)
-    m = @thunk(-Ω)
-    return Ω, Rule(ΔΩ -> extern(m)' * ΔΩ * Ω')
+    m = @thunk(-Ω')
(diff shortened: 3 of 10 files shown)


test/test_util.jl
```diff
@@ -0,0 +1,188 @@
...
+to_vec(X::Diagonal) = vec(Matrix(X)), x_vec->Diagonal(reshape(x_vec, size(X)...))
+
+# Non-array data structures.
+function to_vec(x::Tuple)
+    x_vecs, x_backs = zip(map(to_vec, x)...)
+    sz = cumsum([map(length, x_vecs)...])
+    return vcat(x_vecs...), v->([x_backs[n](v[sz[n]-length(x[n])+1:sz[n]]) for n in 1:length(x)]...,)
+end
+
+function FDM.jvp(fdm, f, x, ẋ)
```
ararslan: Seems like this should be defined in FDM instead, especially since here we're extending a function from FDM without providing types defined in this package (actually no types at all), which is Type Piracy™.

test/test_util.jl
```diff
@@ -0,0 +1,188 @@
+using LinearAlgebra: AbstractTriangular
+
+#####
+##### We need to be able to convert everything into into a vector for use with FDM.jl
```
ararslan: It actually seems to me that all of this could go into FDM, since other packages may also want to use it. (Or alternatively FDM itself could use this machinery in order to be more permissive in what it accepts, then convert things to vectors internally as needed.)

ararslan:
I can't speak to Jarrett's overarching vision, so hopefully he can chime in on those parts, but overall this looks really good to me. I was just thinking about integrating FDM for the tests the other day, as it's proved immensely useful for Nabla. As I said in some in-line comments, I think much of the machinery here could actually be moved to FDM.

willtebbutt:
> As I said in some in-line comments, I think much of the machinery here could actually be moved to FDM.

I'm in total agreement with you here. Will open a PR. Would definitely simplify a lot of stuff here and various other bits of work, as you point out.

src/rules/linalg.jl
```diff
@@ -31,7 +31,7 @@ end
 function rrule(::typeof(inv), x::AbstractArray)
     Ω = inv(x)
     m = @thunk(-Ω)
-    return Ω, Rule(ΔΩ -> m' * ΔΩ * Ω')
+    return Ω, Rule(ΔΩ -> extern(m)' * ΔΩ * Ω')
```
jrevels: Ah good catch. IIRC, this was from back when `adjoint` was defined on `t::Thunk` to be something like `@thunk(extern(t)')`.

Perhaps we should just move the adjoint into the `@thunk`, e.g.:

```julia
m = @thunk(-Ω')
return Ω, Rule(ΔΩ -> m * ΔΩ * Ω')
```

src/rules.jl
```diff
@@ -62,7 +62,10 @@ accumulate(Δ, rule::AbstractRule, args...) = add(Δ, rule(args...))
 """
 TODO
 """
-accumulate!(Δ, rule::AbstractRule, args...) = materialize!(Δ, broadcastable(add(cast(Δ), rule(args...))))
+function accumulate!(Δ, rule::AbstractRule, args...)
+    return materialize!(Δ, broadcastable(add(cast(Δ), rule(args...))))
+end
+accumulate!(Δ::Real, rule::AbstractRule, args...) = accumulate(Δ, rule, args...)
```
jrevels: Seems reasonable. Could probably expand this to `Δ::Number`, even. Could we [add a space between these two definitions](https://github.com/jrevels/YASGuide#linealignmentspacing-guidelines)? 🙂

I guess one place where this might yield unintuitive behavior is if it causes users to think that passing in a non-`materialize!`-able `Δ` always "just" works, e.g. if you call `accumulate!(::SArray, ...)` a user might expect it to fallback to `accumulate` instead of hitting the `setindex!` error it would currently hit when `materialize!(::SArray, ...)` is called. It's not immediately clear how to implement something that ensures that. However, I think we can justify/explain this fallback by saying that it only exists so that callers can handle things generically without checking for the scalar case. That way, it makes sense that numbers are special here, and avoids the impression that a similar fallback exists for immutable containers in general. 

Downstream packages are allowed to special-case these methods as well, which makes it even more okay for us to do so. I'm writing docstrings for these methods today which will hopefully make things clearer.

jrevels:
This is awesome, thanks so much!

Made some comments but everything here generally LGTM, agree with moving the relevant functionality to FDM.jl.

> What is missing? i.e. if we extended the types of tests in test/linalg.jl to cover the entire codebase, what still wouldn't be covered? The test_adjoint! function was already here when I started, so I've kept that. Does it need extending? Do we need a forwards-mode equivalent?

Yup, a forward-mode equivalent will be necessary at some point, though it could probably eventually be merged with the current `test_adjoint!` into a single test harness. 

There are quite a few permutations in terms of features to test; for any given rule, we could presumably test both modes vs. all relevant differential types vs. all relevant rule types vs. a sample of possible input shapes vs. different levels of materialization, etc. In the long run, we'd want to cover as much of that space as possible in shared test harnesses, and leave the rest to ad hoc per-rule tests. As time goes on and more per-rule ad hoc tests are added, we can refactor whenever we find similar tests/common functionality to pull out into the shared harnesses.
 
> There are some tests left in runtests.jl that I've wrapped in a Misc Tests testset. These would ideally go somewhere else. I'm not entirely sure where the optimal location is though and could do with some advice.

I think these could be considered some of those ad hoc per-rule tests; seems like each test file could have its own `Misc Tests` testset for such tests. For example, the `*` and `hypot` test would fall under `test/rule/base.jl`'s miscellaneous tests, while the `broadcast(sin, ...)` one would fall under `test/rule/broadcast.jl`. 

Anyway, this PR is already a huge improvement, thanks again 🙂

src/rules/linalg.jl
willtebbutt: Do we even need a `Thunk` here? What do we gain in this particular case?

src/rules.jl
willtebbutt: > Could we add a space between these two definitions?

ofc :)

> Could probably expand this to Δ::Number

Sounds reasonable.

I'm on board with the rest of what you've suggested as well

willtebbutt:
> Yup, a forward-mode equivalent will be necessary at some point, though it could probably eventually be merged with the current test_adjoint! into a single test harness.

Are you happy for this particular PR to be merged before this happens?

> I think these could be considered some of those ad hoc per-rule tests; seems like each test file could have its own Misc Tests testset for such tests.

Sounds reasonable to me.

src/rules/linalg.jl
jrevels: Good point. The general reason to use a `Thunk` for closed-over computations is to avoid performing the computation if it ends up being unnecessary; e.g. if a `Zero` is passed in for the differential. However, that seems pretty unlikely for a unary rule.

willtebbutt: Ohhh I see. `Zero` has higher precedence than `Thunk` in your monad-y dispatch list, meaning that if the stuff on the RHS of the `Thunk` is `Zero` then the `Thunk` is never materialised and we avoid the computation entirely. Have I understood this correctly? (I'm still getting too grips with this...)

willtebbutt:
Latest push requires [this FDM.jl PR](https://github.com/invenia/FDM.jl/pull/11) to be merged and a new version tagged before tests stand a chance of passing.

ararslan:
This can now use FDM 0.4.0.

Project.toml
```diff
@@ -3,16 +3,19 @@ uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
...
 [deps]
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
+FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 
 [compat]
 Cassette = "^0.2"
+FDM = "^0.4.0"
```
ararslan:
```suggestion
FDM = "^0.4"
```

Project.toml
```diff
@@ -3,16 +3,19 @@ uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 
 [deps]
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
+FDM = "e25cca7e-83ef-51fa-be6c-dfe2a3123128"
```
ararslan: Couldn't FDM be a test-only dependency?

willtebbutt: Good point.

willtebbutt:
@jrevels anything else that you want done before merging?

willtebbutt:
> Should definitely be squashed on merge.

Haha yes, for sure.

jrevels:
LGTM too 👍 thanks again! 

Gave you commit bit so feel free to merge after rebase (sorry for the conflict, I can add the note about `Δ::Real` to the `accumulate!` docstring in a follow-up PR) 🙂

ararslan:
I see no merge conflict...

jrevels:
Huh, I guess GitHub served me an outdated version of the page or something? weird...

PR #18, opened 2019-04-20
A few rrules

willtebbutt:
A few basic rrules for non-dense matrices and related.


src/rules/linear_algebra/diagonal.jl
```diff
@@ -0,0 +1,2 @@
+rrule(::typeof(Diagonal), d::AbstractVector) = Diagonal(d), Rule(ΔΩ->diag(ΔΩ))
```
ararslan:
```suggestion
rrule(::typeof(Diagonal), d::AbstractVector) = Diagonal(d), Rule(diag)
```

src/rules/linear_algebra/diagonal.jl
```diff
@@ -0,0 +1,2 @@
...
+rrule(::typeof(diag), A::AbstractMatrix) = diag(A), Rule(ΔΩ->Diagonal(ΔΩ))
```
ararslan:
```suggestion
rrule(::typeof(diag), A::AbstractMatrix) = diag(A), Rule(Diagonal)
```

willtebbutt:
> Doesn't matter much, but why not keep calling the directory linalg?

Figured it matched the stdlib name a bit better, but I also agree that it really doesn't matter.

ararslan:
Personally I prefer linalg since it's concise and not particularly ambiguous, plus linear_algebra similarly doesn't fully coincide with the stdlib name because of the underscore.

willtebbutt:
Have reverted to `linalg` -- I agree it's a better choice I think.

PR #23, opened 2019-04-22
Fix rrules for Symmetric and Diagonal constructors

ararslan:
Currently these definitions are extending `rrule(::typeof(T), x)` where `T` is a type. However, `typeof(Diagonal) == UnionAll`, which means this is not defining the method it looks like it might be defining. The only reason this worked when originally implemented was that one of the `rrule` definitions was for `rrule(UnionAll, Matrix)` and the other for `rrule(UnionAll, Vector)`, so dispatch still worked.

This replaces these problematic `::typeof(T)`s with `::Type{<:T}`.

### DIFF
--- a/src/rules/linalg/diagonal.jl
+++ b/src/rules/linalg/diagonal.jl
@@ -1,2 +1,2 @@
-rrule(::typeof(Diagonal), d::AbstractVector) = Diagonal(d), Rule(diag)
+rrule(::Type{<:Diagonal}, d::AbstractVector) = Diagonal(d), Rule(diag)
 rrule(::typeof(diag), A::AbstractMatrix) = diag(A), Rule(Diagonal)
--- a/src/rules/linalg/symmetric.jl
+++ b/src/rules/linalg/symmetric.jl
@@ -1,4 +1,4 @@
-rrule(::typeof(Symmetric), A::AbstractMatrix) = Symmetric(A), Rule(_symmetric_back)
+rrule(::Type{<:Symmetric}, A::AbstractMatrix) = Symmetric(A), Rule(_symmetric_back)
 
 _symmetric_back(ΔΩ) = UpperTriangular(ΔΩ) + LowerTriangular(ΔΩ)' - Diagonal(ΔΩ)
 _symmetric_back(ΔΩ::Union{Diagonal, UpperTriangular}) = ΔΩ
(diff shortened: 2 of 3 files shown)


ararslan:
I had thought of doing so but wasn't able to determine how to formulate such a test. The only way I can think to properly catch this case would be to add more matrix constructor sensitivities, which would show method overwriting warnings for `rrule(::Type{UnionAll}, ::Matrix)`. Any suggestions?

willtebbutt:
I was just thinking something simple like checking that
```julia
rrule(Diagonal{Float64}, x)
rrule(Diagonal, x)
```
should do sensible things. Wouldn't generalise to new types of course, but would at least prevent these particular ones from reverting.

ararslan:
> perhaps update the `rrule_test` invocations in the existing tests

Update them in what way?

willtebbutt:
> Update them in what way?

Sorry, I was just thinking something along the lines of changing [this](https://github.com/JuliaDiff/ChainRules.jl/blob/f783dff4c5416d600101882582fd9362d19ea57e/test/rules/linalg/diagonal.jl#L4) and related to
```julia
rrule_test(typeof(Diagonal(randn(rng, N))), randn(rng, N, N), (randn(rng, N), randn(rng, N)))
```

ararslan:
Okay, I've implemented that for `Diagonal`, but `Symmetric` is considerably more difficult, as no one-argument constructor exists for `Symmetric` concrete subtypes.

willtebbutt:
>  but Symmetric is considerably more difficult, as no one-argument constructor exists for Symmetric concrete subtypes.

Is `Symmetric(randn(N, N))` not viable?

ararslan:
Well it is, but that's covered by the existing `Symmetric` tests. If you look at my recent addition to the `Diagonal` tests, you'll see that you can pass the concrete type easily as a constructor to `rrule_test`, and all is well. But `rrule_test` calls `f(x)`, and no such method exists if `f` is a concrete subtype of `Symmetric`; it requires a second argument specifying which is the triangle to consider "primary."

willtebbutt:
Ahh I see what you mean. Good point -- we should probably sort this out in another PR as we'll probably have more of these things to check in the future. Would you mind raising an issue regarding this?

I'm happy with the PR now though. Thanks for sorting this out.

ararslan:
Issue in https://github.com/JuliaDiff/ChainRules.jl/issues/24.

PR #25, opened 2019-04-26
Implement sensitivities for `BLAS.gemm`

ararslan:
These are ported from Nabla.

### DIFF
--- a/src/rules/blas.jl
+++ b/src/rules/blas.jl
@@ -3,6 +3,9 @@ These implementations were ported from the wonderful DiffLinearAlgebra
 package (https://github.com/invenia/DiffLinearAlgebra.jl).
 =#
 
+using LinearAlgebra: BlasFloat
+using LinearAlgebra.BLAS: gemm
+
 _zeros(x) = fill!(similar(x), zero(eltype(x)))
 
 _rule_via(∂) = Rule(ΔΩ -> isa(ΔΩ, Zero) ? ΔΩ : ∂(extern(ΔΩ)))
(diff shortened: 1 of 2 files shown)


src/rules/blas.jl
```diff
@@ -72,3 +74,37 @@ function rrule(f::typeof(BLAS.gemv), tA, A, x)
...
 end
+
+#####
+##### `BLAS.gemm`
+#####
+
+function rrule(::typeof(gemm), tA::Char, tB::Char, α::Number,
+               A::AbstractMatrix, B::AbstractMatrix)
+    C = gemm(tA, tB, α, A, B)
+    #∂α = C̄ -> sum(C̄ .* C) / α  TODO: This is wrong and idk why
```
willtebbutt: Should this be
```julia
∂α = C̄ -> sum(C̄ .* C)
```

ararslan: Unfortunately that doesn't seem to do it:
```
gemm: Test Failed at /home/alex/.julia/dev/ChainRules/test/rules/blas.jl:22
  Expression: ≈(x̄_ad, x̄_fd, rtol=1.0e-9, atol=1.0e-9)
   Evaluated: 4.6207100689748355 ≈ -1.8353251336788157 (rtol=1.0e-9, atol=1.0e-9)
```

ararslan: Dividing by `α` is what Nabla does; I have no idea why it doesn't work in this instance.

willtebbutt: Hmm yeah, this is really weird. Frankly, I have no idea why Nabla does it...

willtebbutt:
I've also just realised that the typing is probably a bit loose. I think that BLAS routines are usually restricted to strided arrays with particular types of floating-point elements. Perhaps we should do the same here?

src/rules/blas.jl
ararslan: Perhaps we should add a `NotImplementedRule` and punt it down the road :stuck_out_tongue:

ararslan:
`gemm` is a case where the methods in LinearAlgebra are, for whatever reason, specified as `AbstractMatrix` instead of `StridedMatrix`. That is a good observation about the element type; it's restricted to `BlasFloat`, and with `alpha`'s type matching the element type of the matrices.

src/rules/blas.jl
ararslan: > Frankly, I have no idea why Nabla does it

Tbh I don't either. I don't see how it would arise from the math.

ararslan: > Perhaps we should add a NotImplementedRule and punt it down the road 😛

I couldn't figure out how to deal with alpha so I actually did this in #28.

ararslan:
To do:
* Add tests for non-square matrices

ararslan:
Rectangular tests added. Good to go?

willtebbutt:
LGTM

PR #26, opened 2019-04-28
Extend Base.isapprox instead of defining cr_isapprox

ararslan:
Currently in the testing utilities we define an internal `cr_isapprox` function that mimics `isapprox` except we only accept a subset of the keyword arguments accepted by `isapprox`. Instead, we can just define methods for `isapprox` on ChainRules-defined types for the purposes of testing. That way we can more easily take arbitrary keyword arguments and get Test's special-cased pretty printing for `@test a ≈ b`.

### DIFF
--- a/test/test_util.jl
+++ b/test/test_util.jl
@@ -3,50 +3,54 @@ using FDM: jvp, j′vp
 const _fdm = central_fdm(5, 1)
 
 """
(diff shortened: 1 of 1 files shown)


ararslan:
Coverage change is bogus; I didn't touch anything in `src/`.

ararslan:
Well we can always revert if we find that this wrecks things horribly :smile:

PR #27, opened 2019-05-01
Reformat rrule_test for more informative output on failure

ararslan:
Currently, if the check for all rules comparing approximately equal to their finite differencing counterparts fails, the test failure just tells you that `all(...)` was `false`. If we restructure it as a loop
over the results, a failure will display the actual values that were compared, which is much more informative.

While I was in here, I also changed the `map` directly below to a loop, since that's what the `map` was doing; the resulting vector of `nothing`s was not being used. It could just have well have been a `foreach`, but an explicit loop seems clearer.

I ran into this because, as it turns out, I'm real good at getting tests to fail when adding new stuff.


ararslan:
I'm just going to merge this since it's a pretty small change without any real downside, and it will be useful for debugging the test failures in #29.

Issue #21, opened 2019-04-22
Dealing with non-array/scalar structured values


ararslan:
Consider the case of an eigenvalue decomposition from the `eigen` function, which produces both eigenvalues and eigenvectors in an `Eigen` object. [Giles](https://people.maths.ox.ac.uk/gilesm/files/NA-08-01.pdf) provides forward- and reverse-mode sensitivities for the decomposition, which depend on both the eigenvalues and vectors. That begs the question of how this should be expressed in ChainRules.

In a conversation with @jrevels, he said that his vision for this was to use named tuples in cases such as this, which involve structures other than arrays and scalars. However, we weren't able to come to a concrete conclusion on whether it makes more sense for a `Rule` for e.g. `eigen` to produce a named tuple upon application of the rule, or if `frule`/`rrule` should themselves should produce a named tuple of `Rule`s.

It seems that the author of a rule can reuse computations for the eigenvalues and vectors if applying the rule yields a single named tuple. This would look something along the lines of (untested):
```julia
function frule(::typeof(eigen), X::AbstractMatrix)
    E = eigen(X)
    λ, U = E
    n = size(X, 1)
    R = Rule() do ΔX
        Y = U' * ΔX * U
        values = copy(diag(Y))
        @inbounds for j = 1:n, i = 1:n
            if i == j
                Y[i,j] = 0
            else
                Y[i,j] /= λ[j] - λ[i]
            end
        end
        vectors = U * Y
        (values=values, vectors=vectors)
    end
    return E, R
end
```
However, if `frule`/`rrule` returns a named tuple of rules, one can specify accessor functions much more conveniently, something like:
```julia
function frule(::typeof(eigvals), X::AbstractMatrix)
    E, R = frule(eigen, X)
    return E.values, R.values
end
```
To quote Jarrett directly, at what stage do we want the caller to know that they should deal with a named tuple?

willtebbutt:
I'm also not entirely sure what the correct answer is here.

The other thing we should sort out while making these choices is whether, when dealing with non-`Array` `AbstractArray` types (e.g. `Diagonal`, `Symmetric` etc), the rules should also be `AbstractArray` types, named tuples (in whatever way we decide), or whether it's acceptable to have both conventions coexist.

ararslan:
[...]

PR #31, opened 2019-05-11
Add SVD factorization rrule

ararslan:
This adds `rrule`s for the SVD factorization as well as an accompanying `rrule` for `getproperty` on `SVD` objects. The definitions are ported from Nabla.

This uses the scheme agreed upon in #21, in which evaluating a `Rule` from `rrule` returns a `NamedTuple`, rather than `rrule` returning a `NamedTuple` of `Rule`s. I believe this PR shows that it is the most consistent and sensible choice.


GiggleLiu:
Does it support complex values? If not, here is a version with tested complex valued AD
https://github.com/GiggleLiu/LinalgBackwards.jl/blob/master/src/svd.jl

Hope it can be helpful to ChainRules.

ararslan:
Cool, thanks! We don't currently have a mechanism in place to automatically test complex-valued derivatives, since we use [FDM](https://github.com/invenia/FDM.jl) for comparison against finite differencing, and I don't believe FDM supports complex values. But I'll look into incorporating your definition (I actually think it's basically the same as what's implemented here) once I figure out an appropriate way to test it within the ChainRules framework.

Issue #20, opened 2019-04-22
add(::Real, ::Wirtinger) throws StackOverflowError


simeonschaub:
When trying to add a Wirtinger derivative to either `Zero()`, `One()`, a real number or `DNE()`, currently a StackOverflowError is thrown. It would probably make sense for real numbers to be handled as `Wirtinger(x, Zero())`, although one would  still have to decide how to handle `DNE()`.

simeonschaub:
The problem here seems to be, that the constructor for `Wirtinger` falls back to just adding primal and conjugate if both are real, so the implementation for `add_wirtinger` fails to change any types and gets stuck in an infinite loop. To have this fall-through behavior in the constructor seems problematic in the first place – I'm not even sure this behavior is actually valid in all cases. If this behavior is valid, one should probably have an extra method, that behaves like the constructor does now and have the constructor always return a `Wirtinger` type.

simeonschaub:
The problem here seems to be, that the constructor for `Wirtinger` falls back to just adding primal and conjugate if both are real, so the implementation for `add_wirtinger` fails to change any types and gets stuck in an infinite loop. To have this fall-through behavior in the constructor seems problematic in the first place – I'm not even sure this behavior is actually valid in all cases. If this behavior is valid, one should probably have an extra method, that behaves like the constructor does now and have the constructor always return a `Wirtinger` type.

Issue #19, opened 2019-04-22
Wrong derivative for `conj`


simeonschaub:
Shouldn't the derivative of `conj` with a complex argument be `Wirtinger(Zero(), One())`? Instead I get this:
```julia
julia> w, dw = frule(conj, 1+im)
(1 - 1im, ChainRules.WirtingerRule{ChainRules.Rule{getfield(ChainRules, Symbol("##304#308"))},ChainRules.Rule{getfield(ChainRules, Symbol("##305#309"))}}(ChainRules.Rule{getfield(ChainRules, Symbol("##304#308"))}(getfield(ChainRules, Symbol("##304#308"))(Core.Box(One()))), ChainRules.Rule{getfield(ChainRules, Symbol("##305#309"))}(getfield(ChainRules, Symbol("##305#309"))(Core.Box(One())))))

julia> dw(One())
One()

julia> dw(1 + 0im)
ChainRules.Wirtinger{Complex{Int64},ChainRules.Zero}(1 + 0im, ChainRules.Zero())
```

simeonschaub:
closed by JuliaDiff/ChainRulesCore.jl#29

PR #34, opened 2019-05-15
removed fall-through for `Wirtinger`

simeonschaub:
See #20. For now, I just removed the fall-through, but one could add another method with this old behavior and then use that, whenever the optimization should occur. Also, the behavior for `DNE` with respect to `add` now seems weird to me, but this might be desired?
```julia
julia> add(DNE(), Zero())
DNE()

julia> add(DNE(), One())
One()

julia> add(DNE(), Wirtinger(1,2))
Wirtinger{Int64,Int64}(1, 2)

julia> add(DNE(), 2)
2
```
Please let me now, what you think should be done here.

### DIFF
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -290,26 +290,3 @@ add_thunk(a, b::Thunk) = add(a, extern(b))
 mul_thunk(a::Thunk, b::Thunk) = mul(extern(a), extern(b))
 mul_thunk(a::Thunk, b) = mul(extern(a), b)
 mul_thunk(a, b::Thunk) = mul(a, extern(b))
-
-#####
-##### misc.
-#####
-
-"""
-    Wirtinger(primal::Real, conjugate::Real)
-
-Return `add(primal, conjugate)`.
-
-Actually implementing the Wirtinger calculus generally requires that the
-summed terms of the Wirtinger differential (`∂f/∂z * dz` and `∂f/∂z̄ * dz̄`) be
-stored individually. However, if both of these terms are real-valued, then
-downstream Wirtinger propagation mechanisms resolve to the same mechanisms as
-real-valued calculus, so that the terms' sum can be eagerly computed and
-propagated without requiring a special `Wirtinger` representation
-
-This method primarily exists as an optimization.
-"""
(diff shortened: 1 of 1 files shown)


simeonschaub:
I've been thinking a bit about what optimization still could be done here and I think the right way to decide wether we need to store a Wirtinger derivative, or wether the Wirtinger conjugate is unnecessary, is depending on wether the function itself is of type ℝ → U ⊂ ℂ or ℂ → U ⊂ ℂ, accordingly. Making this mathematically rigorous is difficult, since the formal definition for Wirtinger derivatives makes no sense in the real case, but for f(x): ℝ → ℂ, and g(x): ℂ → ℝ (neither of them have to be holomorphic) one can find that when applying the chain rule to f ∘ g, one gets the same results as storing the full wirtinger derivative for f, when just storing Wirtinger primal + Wirtinger conjugate and handling f like a holomorphic function. If one would want to implement this, `frule` and `rrule` would have to tell each `WirtingerRule`, whether the corresponding input variable is real or complex. One could then just return a lazy `Rule` where `f = Δx -> add(wirtinger_primal(Δx), wirtinger_conjugate(Δx))` or store this information inside the `WirtingerRule` and just adding the two, when the rule is actually evaluated. I could make this into a PR if this is deemed useful.

jrevels:
Hmm, I guess there's still the problem of `@scalar_rule(conj(x), Wirtinger(Zero(), One()))` with this optimization though (#19)...

Maybe a new constructor is the right approach? We definitely want some nice way to reduce to `Real`s in cases like `@scalar_rule(abs2(x), Wirtinger(x', x))`.

simeonschaub:
Thanks for the review, @jrevels! Could you maybe give some more of the original justification of the current optimization? The more I think about it, the more I am convinced that the current behavior is not correct. I don't think the types of the Wirtinger primal and conjugate have anything to do with whether the inputs of the function are real or complex. The way I previously proposed is definitely a bit uglier and less straight forward but it seems necessary to me. Maybe you have an idea to implement this more elegantly?

simeonschaub:
Succeded by #53

PR #36, opened 2019-05-28
Remove identity overdubs

oxinabox:
These don't do anything, do they?


ararslan:
Good question. I think practically they _probably_ don't, but are they hurting anything by existing?

jrevels:
IIRC they're necessary for correct nested overdubbing of add/mul. Memory's fuzzy though. Would've been good if past Jarrett left a comment 😛

Issue #35, opened 2019-05-27
Evaluating rules is unreasonably slow on first call


ararslan:
Currently, evaluating rules can be pretty slow (and in some cases _extremely_ slow) on the first call. It seems to be due to JIT compiling the anonymous functions, the use of which is a central design point of ChainRules.

Here's a basic example, where it takes 2.6 seconds(!!!) to evaluate the derivative for `svd`:
```
julia> using ChainRules, LinearAlgebra

julia> F, dX = rrule(svd, randn(4, 4));

julia> nt = (U=F.U, S=F.S, V=F.V);

julia> @time dX(nt)
  2.644154 seconds (11.32 M allocations: 589.599 MiB, 6.32% gc time)
4×4 Array{Float64,2}:
  0.189976   0.714807   0.456714    2.18944 
 -0.313175  -0.760914  -0.0347824   0.465126
 -0.53937    1.4229    -1.20063    -1.46798 
 -0.517822   1.03534   -1.66866     0.263543

julia> @time dX(nt)
  0.017788 seconds (2.15 k allocations: 146.201 KiB)
4×4 Array{Float64,2}:
  0.189976   0.714807   0.456714    2.18944 
 -0.313175  -0.760914  -0.0347824   0.465126
 -0.53937    1.4229    -1.20063    -1.46798 
 -0.517822   1.03534   -1.66866     0.263543
```
Compare this to Nabla:
```
julia> using Nabla, LinearAlgebra

julia> X = randn(4, 4); F = svd(X); nt = (U=F.U, S=F.S, V=F.V);

julia> @time ∇(svd, Arg{1}, (), F, nt, X)
  0.631797 seconds (2.37 M allocations: 114.680 MiB, 3.77% gc time)
4×4 Array{Float64,2}:
 -1.09959    1.16123    2.27612    -0.97398 
  1.42299   -0.17832   -2.52324     1.48229 
 -3.32982   -0.746237   1.1226      2.98457 
  0.346265   0.402947  -0.0502055   0.661208

julia> @time dX(nt)
  0.006206 seconds (2.15 k allocations: 146.201 KiB)
4×4 Array{Float64,2}:
  0.189976   0.714807   0.456714    2.18944 
 -0.313175  -0.760914  -0.0347824   0.465126
 -0.53937    1.4229    -1.20063    -1.46798 
 -0.517822   1.03534   -1.66866     0.263543
```

[...]

PR #37, opened 2019-05-28
Make work on IRTools

oxinabox:
This swaps out Cassette for IRTools `@dynamo`.

Which was an attempt to solve #35  which it does.
Time of first call for SVD is down to about 1/4 what it was;

## Original, Cassette
3.128563 seconds (11.08 M allocations: 577.564 MiB
0.005609 seconds (2.19 k allocations: 152.170 KiB)

## No Overdubs (Best Case Senario)
0.794261 seconds (2.49 M allocations: 120.038 MiB, 7.86% gc time)
0.000015 seconds (23 allocations: 3.641 KiB)

## IRTools, with precompile.jl (This branch)
0.881375 seconds (2.98 M allocations: 146.369 MiB, 6.39% gc time)
0.000031 seconds (33 allocations: 4.188 KiB)

Note however I added the precompile.jl which makes sure that IRTools itself has been compiled before anything else. Without that it is much worse than anything else.

Next I will try that for Cassette.

Commented out is a sketch of what it should looklike without usiing `IRTools.IR`
which is close to what it should looklike without IRTools at all.
Doesn't quiet work.
Which might be nice since no IRTools dep, no need to precompile IRTools.
This is a **really** simple transform.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 
 [deps]
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
+IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
--- a/src/ChainRules.jl
+++ b/src/ChainRules.jl
@@ -1,6 +1,7 @@
 module ChainRules
 
-using Cassette
+using IRTools: IRTools, IR, @dynamo, isexpr, xcall
+using IRTools.MacroTools: prewalk
 using LinearAlgebra
 using Base.Broadcast: materialize, materialize!, broadcasted, Broadcasted, broadcastable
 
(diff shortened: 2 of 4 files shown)


oxinabox:
Nope, Cassette does not do so well when I add the `precompile`

### Cassette with precompile.jl
 2.838656 seconds (11.13 M allocations: 580.525 MiB, 7.72% gc time)
 0.004810 seconds (2.19 k allocations: 152.170 KiB)

ararslan:
According to @jrevels, the use of Cassette in this package is purely an implementation detail and could be replaced with regular ol' multiple dispatch. The reason why Jarrett didn't do that to begin with is it requires defining a fair number of methods to resolve ambiguities between the various differential types, rule types, etc. I think going that route will be a cleaner solution overall than replacing the dependency on Cassette with on one IRTools.

oxinabox:
We might want to do this first, which makes things faster.
Since it will take/is a whole ton less code than  #38 will be

jrevels:
Do we know what the reason for the overhead here is and why switching away from Cassette fixes it

oxinabox:
I do not know.

jrevels:
Would probably be good to debug the actual perf issue before churning through a lot of code. Would also be interested if performing the transform via a Cassette pass vs. using dispatch improved things.

oxinabox:
Debugging Cassette performance issues is really really annoying.

willtebbutt:
Should we anticipate that a custom Cassette `@pass` that performs the same transformation as `@dynamo` would result in the same performance gains? i.e. is doing contextual dispatch (and leaving it up to the compiler to figure out that `overdub` does nothing both of the time), rather than manual IR re-writing, the thing that's degrading performance here?

oxinabox:
resolved by https://github.com/JuliaDiff/ChainRulesCore.jl/pull/35

PR #38, opened 2019-05-28
[WIP] Replace all uses of add and mul with + and * and define promotions etc

oxinabox:
From @ararslan https://github.com/JuliaDiff/ChainRules.jl/pull/37#issuecomment-496666401

> According to @jrevels, the use of Cassette in this package is purely an implementation detail and could be replaced with regular ol' multiple dispatch. The reason why Jarrett didn't do that to begin with is it requires defining a fair number of methods to resolve ambiguities between the various differential types, rule types, etc. I think going that route will be a cleaner solution overall than replacing the dependency on Cassette with on one IRTools.

This is going to require so much code.
I've just done the build find and replace so far,
next comes fixing all the ambiguities.
by rewriting a bunch of it to use promotion rules etc
And deleting redundant code after that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ notifications:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("ChainRules"); Pkg.test("ChainRules"; coverage=true)'
 after_success:
   # push coverage results to Coveralls
-  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'using Pkg; Pkg.+(Coveralls.process_folder())'
 jobs:
   include:
     - state: "Documentation"
--- a/src/differentials.jl
+++ b/src/differentials.jl
@@ -9,9 +9,9 @@ support, broadcast fusion, zero-elision, etc. into nicely separated parts.
 
 All subtypes of `AbstractDifferential` implement the following operations:
 
-`add(a, b)`: linearly combine differential `a` and differential `b`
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ notifications:
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("ChainRules"); Pkg.test("ChainRules"; coverage=true)'
 after_success:
   # push coverage results to Coveralls
-  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'using Pkg; Pkg.+(Coveralls.process_folder())'
```
ararslan: Looks like a `sed` issue? :stuck_out_tongue:

src/differentials.jl
```diff
@@ -45,12 +45,12 @@ each step handles an element of the algebra before dispatching to the next step.
...
 of `AbstractDifferential` to resolve potential ambiguities.
 =#
-
+#==
 const PRECEDENCE_LIST = [:wirtinger, :casted, :zero, :dne, :one, :thunk, :fallback]
 
 global defs = Expr(:block)
 
-let previous_add_name = :add, previous_mul_name = :mul
+let previous_add_name = :(Base.:+), previous_mul_name = :(Base.:*)
```
ararslan: This block here could likely be repurposed to define the method hierarchy that resolves ambiguities

oxinabox: Yes

.travis.yml
oxinabox: Yes, I was ment to not commit that one

PR #42, opened 2019-05-30
Make Rules store a second function for updating

### LINKED ISSUE #39: Customizing in-place accumulation
In porting Nabla's reverse mode sensitivity definitions to ChainRules, I've found that it's not particularly straightforward to port those which directly update a value on the tape. The equivalent operation in ChainRules appears to be `accumulate!(value, rule, args...)`, which accumulates the result of `rule(args...)` to `value`. However, most rules are defined in terms of anonymous functions which are themselves defined within `frule`/`rrule`, which means we can't specialize `accumulate!` on particular rules in the outer scope.

I see a few options here:
1. Move rule definitions which have updating counterparts to separate functions, so that one can write
   ```julia
   dfdx(x) = something
   rrule(f, x) = (f(x), Rule(dfdx))
   accumulate!(value, ::Rule{typeof(dfdx)}, x) = whatever
   ```
   One downside to this is that evaluation of the rule and accumulating the result to an existing value can't share any intermediate steps, which could mean using more memory. Another downside is that it would require reshuffling all of the current definitions outside of `frule` and `rrule` into a bunch of top-level named functions.

   On the other hand, an upside to this is that we could precompile the functions defined at the top level, which would likely help with #35.

2. Have `Rule`s store a second function alongside the sensitivity propagation function which can be used for in-place updating. Then we would have
   ```julia
   struct Rule{F<:Function,U<:Union{Function,Nothing}} <: AbstractRule
       f::F
       u::U
   end
   Rule(f) = Rule{typeof(f),Nothing}(f, nothing)
   ```
   `accumulate!` would then use the `u` function for updating if it isn't `nothing`, otherwise it would fall back to the current definition in terms of `materialize!`. The function `u` would have a signature like `u(value, args...)` which accumulates `f(args...)` to `value`.

3. Something else entirely.

I kind of like option 2, though 1 seems fine as well, if a bit suboptimal in terms of the work required to implement that change up front. Thoughts?

ararslan:
Now `Rule`s store two things: one callable thing, which is used for evaluating the rule, and one other thing, which can be `nothing` or a function with the signature `u(value, args...)` that evaluates the rule for the given arguments and adds the result to `value`, doing so in place when possible.

The advantage of this is that we can more easily define custom ways of accumulating the results of rules, and we can even share intermediate steps between the regular rule evaluation and custom updating. As a test/proof of concept, the `rrule` for `svd` now also has an updating function that handles `NamedTuple`s appropriately.

Fixes #39, but implemented before Lyndon made his suggestion.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -108,6 +108,32 @@ See also: [`accumulate`](@ref), [`accumulate!`](@ref), [`AbstractRule`](@ref)
 """
 store!(Δ, rule::AbstractRule, args...) = materialize!(Δ, broadcastable(rule(args...)))
 
+# Special purpose updating for operations which can be done in-place. This function is
+# just internal and free-form; it is not a method of `accumulate!` directly as it does
+# not adhere to the expected method signature form, i.e. `accumulate!(value, rule, args)`.
+# Instead it's `_update!(old, new, extrastuff...)` and is not specific to any particular
+# rule.
+
+_update!(x, y) = x + y
+_update!(x::Array{T,N}, y::AbstractArray{T,N}) where {T,N} = x .+= y
+
+_update!(x, ::Zero) = x
+_update!(::Zero, y) = y
+_update!(::Zero, ::Zero) = Zero()
+
+function _update!(x::NamedTuple{Ns}, y::NamedTuple{Ns}) where Ns
+    return NamedTuple{Ns}(map(p->_update!(getproperty(x, p), getproperty(y, p)), Ns))
+end
+
+function _update!(x::NamedTuple, y, p::Symbol)
+    new = NamedTuple{(p,)}((_update!(getproperty(x, p), y),))
+    return merge(x, new)
+end
+
+function _update!(x::NamedTuple{Ns}, y::NamedTuple{Ns}, p::Symbol) where Ns
+    return _update!(x, getproperty(y, p), p)
+end
+
 #####
 ##### `Rule`
 #####
@@ -123,13 +149,17 @@ Cassette.overdub(::RuleContext, ::typeof(add), a, b) = add(a, b)
 Cassette.overdub(::RuleContext, ::typeof(mul), a, b) = mul(a, b)
 
 """
-    Rule(propation_function)
+    Rule(propation_function[, updating_function])
 
 Return a `Rule` that wraps the given `propation_function`. It is assumed that
 `propation_function` is a callable object whose arguments are differential
 values, and whose output is a single differential value calculated by applying
 internally stored/computed partial derivatives to the input differential
 values.
+
+If an updating function is provided, it is assumed to have the signature `u(Δ, xs...)`
+and to store the result of the propagation function applied to the arguments `xs` into
+`Δ` in-place, returning `Δ`.
 
 For example:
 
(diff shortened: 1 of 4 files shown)


src/rules.jl
```diff
@@ -108,6 +108,28 @@ See also: [`accumulate`](@ref), [`accumulate!`](@ref), [`AbstractRule`](@ref)
 """
 store!(Δ, rule::AbstractRule, args...) = materialize!(Δ, broadcastable(rule(args...)))
 
+# Special purpose updating for operations which can be done in-place
```
willtebbutt: Are these just helper functions for `accumulate!`?

src/rules.jl
```diff
@@ -123,14 +145,18 @@ Cassette.overdub(::RuleContext, ::typeof(add), a, b) = add(a, b)
...
 
 Return a `Rule` that wraps the given `propation_function`. It is assumed that
 `propation_function` is a callable object whose arguments are differential
 values, and whose output is a single differential value calculated by applying
 internally stored/computed partial derivatives to the input differential
 values.
 
+If an updating function is provided, it is assumed to have the signature `u(y, xs...)`
+and to store the result of the propagation function applied to the arguments `xs` into
+`y` in-place, returning `y`.
```
willtebbutt: Nitpicking: `y`->`Δ` for consistency with other code in this file?

src/rules/linalg/factorization.jl
```diff
@@ -12,14 +12,17 @@ end
 
 function rrule(::typeof(getproperty), F::SVD, x::Symbol)
     if x === :U
-        return F.U, (Rule(Ȳ->(U=Ȳ, S=zero(F.S), V=zero(F.V))), DNERule())
+        rule = Ȳ->(U=Ȳ, S=zero(F.S), V=zero(F.V))
```
willtebbutt: For efficiency's sake, should we be using `Zero()` here, as opposed to `zero(F.S)` and `zero(F.V)`? (Or better still, `Zeros` from `FillArrays`, since that retains shape information)

src/rules.jl
ararslan: Yes, I didn't make them methods of `accumulate!` because they don't share the same expected method signature form, i.e. `accumulate!(value, rule, args)`; this is more free-form.

src/rules/linalg/factorization.jl
ararslan: Yeah I thought about `Zero()` but decided not to pursue it for now because it loses size information. I wonder if we could be a bit fancy with the definition of `Zero` such that it retains information when applicable as type parameters.

The issue with using FillArrays, which I ran into when exploring their use in Nabla for a similar purpose, is that they're `AbstractArray`s for which `setindex!` is not defined, which means a lot of internal code has to be audited in order to not make assumptions about being able to assign indices for arbitrary array inputs.

willtebbutt: Hmm okay, fair enough. I do think that we should go down the `FillArrays` route at some point. I'll raise an issue about this so that we don't forget, and revisit later.

I don't think I'm pro encoding size information as a type parameter as I imagine it would play badly with the compiler if we had a lot of differently-sized `Zero` things, which I could well imagine happening.

src/rules.jl
willtebbutt: Fair enough. Could you please extend the comment to this effect? i.e. the reason for `_update!` existing / its intended use?

src/rules/linalg/factorization.jl
ararslan: I'm only against it insofar as I'm not sure that it's a significant enough gain to be worth the effort. Can you elaborate on how it would play poorly with the compiler? If we have size information as type parameters, say as a tuple, we could allow e.g. `Zero{(2,3)}() * Zero{(3,4)}` which would produce a `Zero{(2,4)}`.

willtebbutt: Sure. My concern with the compiler is that a proliferation of types (which, as I understand it, is what having loads of instances of parametric types essentially does) will increase compile times substantially, because you have lots of types. I'm far from expert here though, so I could be completely wrong.

The other thing is that I doubt that we really don't lose much performance by using dynamic sizing. i.e. `Zeros(2, 3) * Zeros(3, 4)` is still _really_ fast.

oxinabox: Appropriate `@nospecialize` solves

willtebbutt: Sure. But then we might as well just use `FillArrays` and be done with it, since it's already defined all of these things and is something I think we should be supporting anyway.

ararslan: This is a bit out of scope for the current PR so we should continue this discussion in an issue.

PR #44, opened 2019-06-06
Add an rrule for the Cholesky decomposition

ararslan:
This is a (mostly) direct port of the code from Nabla.

### DIFF
--- a/src/rules/linalg/factorization.jl
+++ b/src/rules/linalg/factorization.jl
@@ -1,3 +1,6 @@
+using LinearAlgebra: checksquare
+using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
+
 #####
 ##### `svd`
 #####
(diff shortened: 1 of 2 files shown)


src/rules/linalg/factorization.jl
```diff
@@ -79,3 +82,192 @@ function _add!(X::AbstractMatrix{T}, Y::AbstractMatrix{T}) where T<:Real
...
+
+# TODO: The comment below about row- versus column-major is incorrect; Julia's arrays are
+# column-major. This comment along with the below implementation was copied from Nabla and
+# works, but the implementation should perhaps should be revisited to ensure we're actually
+# being cache friendly.
+#
+# See [1] for implementation details: pages 5-9 in particular. The derivations presented in
+# [1] assume column-major layout, whereas Julia primarily uses row-major. We therefore
+# implement both the derivations in [1] and their transpose, which is more appropriate to
+# Julia.
```
oxinabox: Canwe **not** tranpose comments we know are wrong?
Can we just open an issue about checking this for cache friendlyness and delete 80% of this comment?

src/rules/linalg/factorization.jl
```diff
@@ -79,3 +82,192 @@ function _add!(X::AbstractMatrix{T}, Y::AbstractMatrix{T}) where T<:Real
...
+#
+# [1] - "Differentiation of the Cholesky decomposition", Murray 2016
+
+"""
+    level2partition(A::AbstractMatrix, j::Integer, upper::Bool)
+
+Returns views to various bits of the lower triangle of `A` according to the
+`level2partition` procedure defined in [1] if `upper` is `false`. If `upper` is `true` then
+the transposed views are returned from the upper triangle of `A`.
```
oxinabox:
```suggestion
the transposed views are returned from the upper triangle of `A`.

 [1] - "Differentiation of the Cholesky decomposition", Murray 2016
```

src/rules/linalg/factorization.jl
```diff
@@ -79,3 +82,192 @@ function _add!(X::AbstractMatrix{T}, Y::AbstractMatrix{T}) where T<:Real
...
+        r = view(A, 1:j-1, j)
+        d = view(A, j, j)
+        B = view(A, 1:j-1, j+1:n)
+        c = view(A, j, j+1:n)
+    else
+        r = view(A, j, 1:j-1)
+        d = view(A, j, j)
+        B = view(A, j+1:n, 1:j-1)
+        c = view(A, j+1:n, j)
+    end
```
oxinabox: This mighjt be one of the rare times i suggest wrapping the block in `@views`

Also didn't we just do out bounds check so we can now `@inbounds` it.
(again rare time i am going to suggest doing that to a whole l block)

src/rules/linalg/factorization.jl
ararslan: I didn't analyze the algorithm in depth to determine whether the code itself is the transpose of what it should be, but I figured that comment would be a good reminder to go back and fact check. An issue would work as well I suppose.

@willtebbutt, would you be willing to take a look at this, since you're the original author of the comment?

willtebbutt: Happy to follow up with this at some point if @oxinabox 's suggestion is followed. Certainly it would be useful to get rid of the comment for now.

test/rules/linalg/factorization.jl
```diff
@@ -21,4 +23,65 @@
             @test ChainRules._add!(copy(X), Y) ≈ X + Y
         end
     end
+    @testset "cholesky" begin
+        rng = MersenneTwister(4)
+        @testset "the thing" begin
+            X = generate_well_conditioned_matrix(rng, 10)
+            V = generate_well_conditioned_matrix(rng, 10)
+            F, dX = rrule(cholesky, X)
+            for p in [:U, :L]
```
oxinabox:
```suggestion
            for p in (:U, :L)
```

ararslan: Meh, it's just in the tests.

oxinabox: indeed meh

src/rules/linalg/factorization.jl
```diff
@@ -79,3 +82,192 @@ function _add!(X::AbstractMatrix{T}, Y::AbstractMatrix{T}) where T<:Real
...
+    return r, d, B, c
+end
+
+"""
+    level3partition(A::AbstractMatrix, j::Integer, k::Integer, upper::Bool)
+
+Returns views to various bits of the lower triangle of `A` according to the
+`level3partition` procedure defined in [1] if `upper` is `false`. If `upper` is `true` then
+the transposed views are returned from the upper triangle of `A`.
+"""
```
oxinabox:
```suggestion
[1] - "Differentiation of the Cholesky decomposition", Murray 2016
"""
```

src/rules/linalg/factorization.jl
oxinabox: --


 
--
# [1] - "Differentiation of the Cholesky decomposition", Murray 2016

src/rules/linalg/factorization.jl
ararslan: The only problem with `@views` is that AFAICT you aren't guaranteed to actually get a view, since it uses `Base.maybeview` rather than replacing `getindex` with `view`.

oxinabox: Isn't that ok though?
Or does this need a `view`?

ararslan: Not sure, I just wanted to make sure this was the same as Nabla's implementation.

PR #45, opened 2019-06-06
Port more array-related derivatives from Nabla

ararslan:
The stuff in Nabla's `src/sensitivities/array.jl`, which includes `reshape`, `hcat`, `vcat`, and `fill`.

### DIFF
--- a/src/rules/array.jl
+++ b/src/rules/array.jl
@@ -1,0 +1,12 @@
+#####
+##### `reshape`
+#####
+
+function rrule(::typeof(reshape), A::AbstractArray, dims::Tuple{Vararg{Int}})
+    return reshape(A, dims), (Rule(Ȳ->reshape(Ȳ, dims)), DNERule())
+end
+
+function rrule(::typeof(reshape), A::AbstractArray, dims::Int...)
+    Y, (rule, _) = rrule(reshape, A, dims)
+    return Y, (rule, fill(DNERule(), length(dims))...)
+end
--- a/src/rules/array.jl
+++ b/src/rules/array.jl
@@ -22,6 +22,8 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
         l = mapreduce(j->size(Xs[j], 2), Base.add_sum, 1:i-1; init=0)
         u = l + size(Xs[i], 2)
         dim = u > l + 1 ? (l+1:u) : u
+        # NOTE: The copy here is defensive, since `selectdim` returns a view which we can
+        # materialize with `copy`
         Rule(Ȳ->copy(selectdim(Ȳ, 2, dim)))
     end
     return Y, rules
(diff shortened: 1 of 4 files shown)


src/rules/array.jl
```diff
@@ -0,0 +1,56 @@
...
+#####
+
+function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
+    Y = hcat(A, Bs...)
+    Xs = (A, Bs...)
+    rules = ntuple(length(Bs) + 1) do i
+        l = mapreduce(j->size(Xs[j], 2), Base.add_sum, 1:i-1; init=0)
+        u = l + size(Xs[i], 2)
+        dim = u > l + 1 ? (l+1:u) : u
+        Rule(Ȳ->copy(selectdim(Ȳ, 2, dim)))
```
oxinabox: why `copy`?
For my education on how ChainRules thinks of defensive copies

src/rules/array.jl
```diff
@@ -0,0 +1,56 @@
...
+#####
+##### `fill`
+#####
+
+function rrule(::typeof(fill), value::Any, dims::Tuple{Vararg{Int}})
+    return fill(value, dims), (Rule(sum), DNERule())
+end
+
+function rrule(::typeof(fill), value::Any, dims::Int...)
+    return fill(value, dims), (Rule(sum), fill(DNERule(), length(dims))...)
```
oxinabox:
```suggestion
    return fill(value, dims), (Rule(sum), Fill(DNERule(), length(dims))...)
```

I thought we added a FillArrays dependency?
He haven't but we should?

willtebbutt: I wouldn't be pro- introducing this here until we've addressed #46 . Also, it doesn't look like `Fill`s have been optimised for splatting into a `Tuple` yet, so there's no real performance gain here.

oxinabox: Oh, I missed that it was splatted
Fair enough

oxinabox: We can make this compile time, if we use `ntuple`
```
julia> @btime (()->tuple(ntuple(_->Nothing(), 10)...))()
  0.027 ns (0 allocations: 0 bytes)
(nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing)

julia> @btime (()->tuple(fill(Nothing(), 10)...))()
  319.146 ns (1 allocation: 80 bytes)
(nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing)
```


```suggestion 
    return fill(value, dims), (Rule(sum), ntuple(_->DNERule(), length(dims))...)
```

src/rules/array.jl
ararslan: Definitely a defensive copy in this case; `selectdim` returns a view, which `copy` materializes. This is the same thing that Nabla does.

oxinabox: Right, I see why that is needed.

Thanks

ararslan: I should add a comment to that effect, so thanks for asking. :slightly_smiling_face:

PR #48, opened 2019-06-07
Set version to 0.0.1

ararslan:
In preparation for an initial release


oxinabox:
I don't think you are allowed to  register below `0.1.0`
But I guess we will see?

ararslan:
Registrator didn't complain, so it would just the people who merge PRs on the General registry. Should be fine.

PR #49, opened 2019-06-07
Add in-place updating for gemm and gemv rrules

ararslan:


test/rules/blas.jl
```diff
@@ -20,6 +20,41 @@ using LinearAlgebra.BLAS: gemm
                 x̄_ad = dx(ȳ)
                 x̄_fd = j′vp(central_fdm(5, 1), f, ȳ, x)
                 @test x̄_ad ≈ x̄_fd rtol=1e-9 atol=1e-9
+                if size(x) != ()  # A and B
+                    @test dx isa Rule{<:Function,<:Function}
+                    x̄ = zeros(size(x)...)
```
willtebbutt: Could we please implement helper functionality similar to that found in Nabla.jl's tests for these accumulation rules? We're going to be adding quite a few of these `accumulate!` rules, so would be worthwhile doing now, with this PR, rather than later on.

test/rules/blas.jl
```diff
@@ -20,6 +20,41 @@ using LinearAlgebra.BLAS: gemm
...
+            for (f, z, dz) in [(z->gemv(t, z, A, x), α, dα),
+                               (z->gemv(t, α, z, x), A, dA),
+                               (z->gemv(t, α, A, z), x, dx)]
+                ȳ = randn(rng, size(y)...)
+                z̄_ad = dz(ȳ)
+                z̄_fd = j′vp(central_fdm(5, 1), f, ȳ, z)
+                @test z̄_ad ≈ z̄_fd atol=1e-9 rtol=1e-9
+                if size(z) != ()  # A and x
+                    @test dz isa Rule{<:Function,<:Function}
+                    z̄ = zeros(size(z)...)
```
willtebbutt: Zero-ing out the cache in which we accumulate is dangerous since it leaves us vulnerable to bugs in which the accumulation rule _only_ works correctly if the cache is full of zeros.

ararslan: I'm not sure what you mean. This is only zeroing out in the test. Should we be using like `ones` or something instead?

willtebbutt: Yeah, preferably `randn` to avoid any particular deterministic structure in the values in the cache. If I'm not misremembering, we've had bugs in the past in Nabla whereby an in-place sensitivity accidentally zero-d out the cache, and we failed to pick this up in previous versions of the test suite because we were doing the above and only testing with caches whose values started at zero anyway.

test/rules/blas.jl
ararslan: We should be able to use `rrule_test` here, which includes accumulation testing, once #56 is merged. Currently `rrule_test` can't be used because it assumes all arguments are differentiable, which isn't the case for the BLAS functions.

ararslan:
I think I pushed something by accident. Will fix in a bit.

ararslan:
This now uses `rrule_test`, which tests accumulation appropriately without having to do things manually.

willtebbutt:
LGTM. Great work :)

PR #53, opened 2019-06-14
 implement a fallthrogh for Wirtinger based on input type 

### LINKED ISSUE #19: Wrong derivative for `conj`

simeonschaub:
Implements the correct fallthrough behavior for `WirtingerRule`s. Succeeds #34. Should fix #19 and #20. Still needs tests and documentation. It should also be decided how `DNE` behaves with respect to Wirtinger, I haven't thought about that too much yet. @jrevels Do you agree with this approach?


src/rules.jl
```diff
@@ -206,13 +206,26 @@ DNERule(args...) = DNE()
...
+derivative ``∂/∂x̅``. If the domain `𝒟` is specified, return a `Rule` evaluating
+to `P(Δ) + C(Δ)` if `𝒟 <: Real`, otherwise return `WirtingerRule(P, C)`.
 """
 struct WirtingerRule{P<:AbstractRule,C<:AbstractRule} <: AbstractRule
     primal::P
     conjugate::C
 end
 
+function WirtingerRule(𝒟::Type, P::AbstractRule, C::AbstractRule)
+    if 𝒟 <: Real || eltype(𝒟) <: Real
```
jrevels: We should make sure that the compiler elides this check when these are "statically" computable (it should)

jrevels:
Good stuff, thanks!

src/rules.jl
simeonschaub: You mean by adding `@pure` or making it a generated function? I thought about it, but this seems fairly trivial and `@code_warntype` says it's fine. Still, probably wouldn't hurt.

jrevels: > You mean by adding @pure or making it a generated function?

Nope, the compiler should do this without any tricks, but type inference can be fairly conservative w.r.t. compositions of predicate statements it will choose to compute statically. This one definitely should be fine, but was worth checking 🙂 

> @code_warntype says it's fine

Awesome! Can we add some `@inferred` tests?

simeonschaub: > Awesome! Can we add some `@inferred` tests?

Will do

simeonschaub:
Seems like test failure is due to the infamous JuliaLang/julia#15276, will try to fix

simeonschaub:
Seems like test failure is due to the infamous JuliaLang/julia#15276, will try to fix

simeonschaub:
@jrevels Are there any more test cases, you think should be added? Any other changes before this gets merged?

nickrobinson251:
The was merged into ChainrulesCore https://github.com/JuliaDiff/ChainRulesCore.jl/pull/29 🎉

PR #54, opened 2019-06-14
fix `BLAS.nrm2` and `BLAS.dot`

simeonschaub:
The rules for the strided methods were completely broken. `BLAS.nrm2` now correctly returns wirtinger derivatives, if arguments are complex. I added a constructor for `Wirtinger` for arrays, which just returns an array of `Wirtinger`s in order for that to work. Still needs tests.


YingboMa:
I will close this PR since ChainRules.jl has no Wirtinger now.

PR #55, opened 2019-06-15
 fix `hypot`, return `Wirtinger` appropriately

simeonschaub:
Turns out the derivative for `hypot` was wrong, even in the real case. Now supports an arbitrary number of real or complex arguments. Also in the process discovered that the `rrule` for `*` was wrong for complex arguments. ~~Still needs tests.~~


simeonschaub:
Bump. Can we get this merged?

src/rules/base.jl
```diff
@@ -37,7 +37,7 @@
 @scalar_rule(conj(x), Wirtinger(Zero(), One()))
 @scalar_rule(adjoint(x), Wirtinger(Zero(), One()))
 @scalar_rule(transpose(x), One())
-@scalar_rule(abs(x), sign(x))
+@scalar_rule(abs(x), x isa Real ? sign(x) : Wirtinger(x' / 2Ω, x / 2Ω))
```
ararslan:
```suggestion
@scalar_rule(abs(x::Real), sign(x))
@scalar_rule(abs(x::Complex), Wirtinger(x' / 2Ω, x / 2Ω))
```

src/rules/base.jl
```diff
@@ -62,7 +62,37 @@
 @scalar_rule(cotd(x), -(π / oftype(x, 180)) * (1 + Ω^2))
 @scalar_rule(sech(x), -tanh(x) * Ω)
 @scalar_rule(csch(x), -coth(x) * Ω)
-@scalar_rule(hypot(x, y), (y / Ω, x / Ω))
+
+function frule(::typeof(hypot), x::Real...)
+    Ω = hypot(x...)
+    return Ω, Rule((Δ...) -> sum(Δ .* x) * inv(Ω))
```
ararslan:
```suggestion
    return Ω, Rule((Δ...) -> sum(Δ .* x) / Ω)
```

src/rules/base.jl
```diff
@@ -62,7 +62,37 @@
...
+    Ω = hypot(x...)
+    return Ω, map(x) do x_i
+        Rule(ΔΩ -> ΔΩ * (x_i / Ω))
+    end
+end
+
+function frule(::typeof(hypot), x...)
+    Ω = hypot(x...)
+    return Ω, WirtingerRule(
+        Rule((Δ...) -> sum(Δ .* conj.(x) * inv(2Ω))),
```
ararslan: This should be moved flush with the opening parenthesis with the second line aligned below it.
```julia
    return Ω, WirtingerRule(Rule((Δ...) -> sum(Δ .* conj.(x) * inv(2Ω))),
                            Rule((Δ...) -> sum(Δ .* x) * inv(2Ω)))
```
Again in this case, why multiply by `inv` instead of just dividing?

test/rules/base.jl
```diff
@@ -1,10 +1,27 @@
+_isapprox(x, y; kwargs...) = isapprox(x, y; kwargs...)
```
ararslan: Same comments as in the other PR.

test/rules/base.jl
```diff
@@ -65,36 +82,46 @@ end
         # TODO: atan2 sincos
     end
     @testset "Misc. Tests" begin
-        @testset "*(x, y)" begin
-            x, y = rand(3, 2), rand(2, 5)
+        @testset "*(x, y)" for T in (Float32, Float64, Complex{Float64})
```
ararslan:
```suggestion
        @testset "*(x::$T, y::$T)" for T in (Float32, Float64, Complex{Float64})
```
otherwise the name of the testset doesn't actually tell you what it's testing.

src/rules/base.jl
simeonschaub: The problem is that `Zero() / x` and `One() / x` doesn't get overdubbed, so this throws errors if all `Δ`s are `Zero()`.

src/rules/base.jl
simeonschaub: See my comment above. It might make sense to overload `/` for `Zero()` and `One()` though.

test/rules/base.jl
simeonschaub: Thanks for the suggestion! Will change that.

src/rules/base.jl
ararslan: Is that because only multiplications and additions are overdubbed?

simeonschaub: Yes, exactly

src/rules/base.jl
ararslan: Yeah I think that would probably make sense.

src/rules/base.jl
simeonschaub: Now that we have that, this is definitely nicer!

src/rules/base.jl
simeonschaub: Would it be ok if I left it for now? I'm not too experienced with Cassette and there would be some edge cases to think about, so I think this deserves its own PR.

test/rules/base.jl
simeonschaub: I'll delete it once #64 is merged

src/rules/base.jl
simeonschaub: I had a go at it. Could maybe be done more elegantly, but works for now.

PR #56, opened 2019-06-16
Add rrule for map and expand the testing framework

ararslan:
This implements `rrule(map, f, xs...)` and expands `rrule_test` to allow non-differentiable arguments. For such cases, the user need only pass `nothing` as the argument's assumed sensitivity.

~~The `rrule` for `map` is admittedly a little bizarre looking. It was ported from Nabla and attempts to use the same logic but stuffed awkwardly into the ChainRules framework.~~ `rrule` for `map` requires that the mapped function have an `rrule` defined for it.


src/rules/mapreduce.jl
```diff
@@ -0,0 +1,30 @@
+#####
+##### `map`
+#####
+
+function rrule(::typeof(map), f, xs...)
+    T = Tuple{Core.Typeof(f), map(eltype, xs)...}
+    if hasmethod(rrule, T)
+        body = i->(ȳi, xis...)->begin
+            _, ∂x = rrule(f, xis...)
+            extern((∂x isa Tuple ? ∂x[i] : ∂x)(ȳi))
```
ararslan: #57 will allow us to dump the `isa Tuple` check and just use `∂x[i]` unconditionally

src/rules/mapreduce.jl
```diff
@@ -0,0 +1,30 @@
...
+        end
+    elseif hasmethod(frule, T)
```
ararslan: We could alternatively say that the `rrule` for `map(f, xs...)` requires that `f` have an `rrule`. That seems reasonable to me. I did it this way because that's what Nabla does: it does a `hasmethod` check to see whether there's a sensitivity defined, and if not, it uses a fallback FMAD implementation. Requiring an `rrule` for `f` would simplify this bizarre looking logic considerably.

test/test_util.jl
```diff
@@ -50,23 +50,58 @@ function rrule_test(f, ȳ, (x, x̄)::Tuple{Any, Any}; rtol=1e-9, atol=1e-9, fdm
     test_accumulation(Zero(), dx, ȳ, x̄_ad)
 end
 
+function _make_fdm_call(fdm, f, ȳ, xs, ignores)
```
ararslan: Eventually what this function does should be moved to FDM, but it took less time to write this than it would to figure out how to incorporate it into the FDM framework directly.

src/rules/mapreduce.jl
```diff
@@ -0,0 +1,30 @@
+#####
+##### `map`
+#####
+
+function rrule(::typeof(map), f, xs...)
+    T = Tuple{Core.Typeof(f), map(eltype, xs)...}
+    if hasmethod(rrule, T)
```
jrevels: You should actually call `rrule`/`frule` and check the result, instead of trying to use reflection to drive closure creation. In general, we should try to discourage unnecessary reflection; that was one of the old DiffRules problems ChainRules set out to fix.

ararslan: The reason I did it this way was to avoid having the check inside of the `map`ped function, so that we don't do the check at every iteration. But that's probably preferable to this mess anyway.

jrevels: > was to avoid having the check inside of the mapped function

The no-rule case (returning `nothing`) is very compiler-friendly; an `isa Nothing` check should get compiled away. But even if it wasn't, AFAICT, the reflection approach seems incorrect for arguments with non-concrete eltypes (rules work even if they are dynamically dispatched to)

ararslan: I dumped all of this crazy stuff in favor of a simple check within the mapped function inside the rule.

src/rules/mapreduce.jl
ararslan: I've implemented that.

src/rules/mapreduce.jl
ararslan: Rebased to include that change.

ararslan:
In the interest of time and getting more stuff moved over from Nabla, I'm going to go ahead with this since it passes tests and is (now) pretty straightforward, plus nothing yet depends on ChainRules. Please feel free to open an issue about anything added here.

PR #57, opened 2019-06-17
Define `getindex` on rules

ararslan:
We define `iterate` on them for convenience, and `getindex` provides a similar convenience for cases where you're not sure whether the rules resulting from a call to `rrule`/`frule` will be a `Tuple`. So instead of writing `partials isa Tuple ? partials[i] : partials`, you can now just write `partials[i]`.


ararslan:
That's the hope :D

PR #58, opened 2019-06-18
Add `_checked_rrule`, which errors if `rrule` returns `nothing`

ararslan:
There are cases where you want to ensure that a call to `rrule` produced something before you proceed. In such cases, one can now use an internal function called `_checked_rrule`, which produces an informative error if `rrule` returns `nothing`.

One such case is exemplified in the second commit, which uses `_checked_rrule` inside of the `rrule` for `map`.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -391,7 +391,7 @@ end
 
 function _checked_rrule(f, args...; kwargs...)
     r = rrule(f, args...; kwargs...)
-    r === nothing && _throw_checked_rrule_error(f, args...; kwargs...)
+    r isa Nothing && _throw_checked_rrule_error(f, args...; kwargs...)
     return r
 end
 
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -377,6 +377,24 @@ See also: [`frule`](@ref), [`AbstractRule`](@ref), [`@scalar_rule`](@ref)
 """
 rrule(::Any, ::Vararg{Any}) = nothing
 
+@noinline function _throw_checked_rrule_error(f, args...; kwargs...)
+    io = IOBuffer()
+    print(io, "can't differentiate `", f, '(')
+    join(io, map(arg->string("::", typeof(arg)), args), ", ")
+    if !isempty(kwargs)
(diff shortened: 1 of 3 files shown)


src/rules.jl
```diff
@@ -377,6 +377,18 @@ See also: [`frule`](@ref), [`AbstractRule`](@ref), [`@scalar_rule`](@ref)
 """
 rrule(::Any, ::Vararg{Any}) = nothing
 
+function _checked_rrule(f, args...; kwargs...)
+    r = rrule(f, args...; kwargs...)
+    if r === nothing
```
jrevels: Just to make it easier on the compiler, might want to change this to a `isa Nothing` and hide the body of the conditional behind a `@noinline` funciton, e.g.

```julia
r = rrule(f, args...; kwargs...)
r isa Nothing && _throw_checked_rrule_error(f, args)
return r
```

ararslan: `isa Nothing` is identical to `=== nothing`, as I understand it; both `isa` and `===` are special-cased.

jrevels: One requires constant propagation and one doesn't (unless there is additional special casing to render them actually equivalent, in which case it's even more tenuous to rely on). Either way, would be good to make things as easy as possible on the compiler side and require as little special casing as possible (learned this the hard way during the whole Cassette journey 😛)

ararslan: ¯\\\_(ツ)\_/¯ Keno says, "Up to compiler bugs, those are equivalent."

jrevels: "Up to compiler bugs" being the operative phrase. Yes, the `===` and `isa` tfuncs are similar, but there's a higher burden (i.e. more special casing required). I know this because I've implemented some of the code in the compiler that handles this w.r.t. correct caching for constant propagation (and further special casing for singletons). 

I'm not asking for a complicated change here, just a simple defensive one.

ararslan: Yeah, it's not a complicated change, I just wanted the full story on the motivation before implementing it. Thanks for explaining, changed in the most recent commit.

PR #59, opened 2019-06-19
Add a few more reduction rrules

ararslan:
These are all ported from Nabla.


src/rules/mapreduce.jl
```diff
@@ -15,10 +15,72 @@ function rrule(::typeof(map), f, xs...)
...
+        y = $call
+        ∂x = Rule() do ȳ
+            broadcast(x, ȳ) do xi, ȳi
+                _, ∂xi = _checked_rrule(f, xi)
+                extern(∂xi(ȳi))
+            end
+        end
+        return y, (DNERule(), DNERule(), ∂x)
+    end
+    eval(Expr(:function, sig, body))
```
oxinabox: Can't we just do:
```suggestion
    $(Expr(:function, sig, body))
```

ararslan: No, this isn't inside of a quoted expression.

PR #60, opened 2019-06-19
Allow keyword arguments in the fallback rrule/frule methods

ararslan:
This permits using `rrule(f, args...; kwargs...)` unconditionally and receiving `nothing` as expected rather than an error about `rrule` not taking keyword arguments.


ararslan:
Coverage change looks bogus. The web UI for Coveralls, as far as I can tell, shows no change.

PR #61, opened 2019-06-19
Bring in BLAS names that are used without qualification

ararslan:
Currently some of the rules are calling `blascopy!` and `scal!` without qualification, which fails, since we haven't done a full `using` for the BLAS submodule. (We can do that later, but currently there's an export
conflict with LinearAlgebra itself.) So for now we can just selectively use the names.

### DIFF
--- a/src/ChainRules.jl
+++ b/src/ChainRules.jl
@@ -2,9 +2,17 @@ module ChainRules
 
 using Cassette
 using LinearAlgebra
+using LinearAlgebra.BLAS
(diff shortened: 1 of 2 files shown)


oxinabox:
Can we just do a full import, and then qualify the clash?

ararslan:
If I remember correctly, I hadn't done that before because a warning is emitted, but I might not be remembering correctly. I'll give that a go.

ararslan:
Implemented that. Thanks Lyndon!

PR #62, opened 2019-06-19
Ensure scalar_rule rules are Number typed

ararslan:
Currently, `@scalar_rule` generates `rrule` methods with no type constraints. This poses a problem for rules with different definitions that have not yet been implemented. For example, the rule for matrix
exponential is quite different from that for scalar exponential, but in the absence of an `rrule(::typeof(exp), ::AbstractMatrix)` method, the incorrect fallback generated by `@scalar_rule` is used.

To solve this, `@scalar_rule` now allows type constraints, e.g. `@scalar_rule(f(x::Complex), g(x))`, and it adds explicit `::Number` constraints to the generated methods if no such constraints are provided.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -410,7 +410,7 @@ A convenience macro that generates simple scalar forward or reverse rules using
 the provided partial derivatives. Specifically, generates the corresponding
 methods for `frule` and `rrule`:
 
-    function ChainRules.frule(::typeof(f), x₁, x₂, ...)
+    function ChainRules.frule(::typeof(f), x₁::Number, x₂::Number, ...)
         Ω = f(x₁, x₂, ...)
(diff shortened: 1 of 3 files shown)


jrevels:
IIUC, will also need to update https://github.com/JuliaDiff/ChainRules.jl/blob/1df670ed9a9f516d26a88315efc2cdbad50f075c/src/rules/base.jl#L41 (the second argument is not `::Number`, but `::RoundingMode`)

ararslan:
Good notes, thanks! I've added a blurb about it to the docstring and added a `::RoundingMode` annotation to `rem2pi`.

ararslan:
Not sure what's up with the coverage drop. I think it doesn't like that I touched `rem2pi`.

PR #64, opened 2019-06-21
Add basic complex functions, fix 2-arg `atan`

simeonschaub:


test/rules/base.jl
```diff
@@ -1,10 +1,31 @@
-function test_scalar(f, f′, xs...)
+_isapprox(x, y; kwargs...) = isapprox(x, y; kwargs...)
```
ararslan: Why is this necessary?

test/rules/base.jl
```diff
@@ -1,10 +1,31 @@
...
+
+const AbstractDifferential = ChainRules.AbstractDifferential
```
ararslan: This is not how you import specific bindings in Julia. Use `using ChainRules: AbstractDifferential` at the top of `test/runtests.jl` if the binding you want is not already exported.

test/rules/base.jl
```diff
@@ -1,10 +1,31 @@
...
+_isapprox(x::Wirtinger, y; kwargs...) = _isapprox(x.primal, y; kwargs...) &&
+    _isapprox(x.conjugate, Zero(); kwargs...)
+_isapprox(x, y::Wirtinger; kwargs...) = _isapprox(x, y.primal; kwargs...) &&
+    _isapprox(Zero(), y.conjugate; kwargs...)
+_isapprox(x::Wirtinger, y::Wirtinger; kwargs...) = _isapprox(x.primal, y.primal; kwargs...) &&
+    _isapprox(x.conjugate, y.conjugate; kwargs...)
+
+const ≈₁ = _isapprox
+
+function test_scalar(f, f′, xs...; broken = false)
```
ararslan: The `broken` keyword argument doesn't appear to actually be used; it should just be removed.

test/rules/base.jl
simeonschaub: Mainly because `isapprox` isn't implemented for `Zero()` and `One()`. It might make sense to overload `Base.isapprox` instead, but I don't know if that belongs in this PR.

test/rules/base.jl
simeonschaub: Will do. The problem was that having `using ChainRules: AbstractDifferential` in this file, made the test crash on my machine, for some reason, so this seemed like the quickest fix.

test/rules/base.jl
simeonschaub: Yes, that's just a remainder. I'll just delete it.

test/rules/base.jl
ararslan: If it's relevant to this change then it's fine in this PR. It's better than adding a separate function at any rate.

test/rules/base.jl
ararslan: Sounds good, thanks.

test/rules/base.jl
ararslan: Hm, that's... bizarre.

simeonschaub:
@ararslan I've implemented your requested changes. Do you agree with the way `isapprox` behaves now?

ararslan:
I think the `isapprox` overloads should stay in the test utilities.

simeonschaub:
> I think the `isapprox` overloads should stay in the test utilities.

Yes, probably good not to overcomplicate the definition of `AbstractDifferential`s.

src/rules/base.jl
```diff
@@ -64,14 +64,26 @@
...
+@scalar_rule(atan(y, x), @setup(u = x^2 + y^2), (x / u, -y / u))
 @scalar_rule(max(x, y), @setup(gt = x > y), (gt, !gt))
 @scalar_rule(min(x, y), @setup(gt = x > y), (!gt, gt))
 @scalar_rule(mod(x, y), @setup((u, nan) = promote(x / y, NaN16)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -floor(u))))
 @scalar_rule(rem(x, y), @setup((u, nan) = promote(x / y, NaN16)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -trunc(u))))
 
+@scalar_rule(angle(x), @setup(u = abs2(x)), Wirtinger(-im//2 * x' / u, im//2 * x / u))
+frule(::typeof(angle), x::Real) = angle(x), Rule(_ -> Zero())
```
ararslan: Can't you just make this
```julia
@scalar_rule(angle(x::Complex), @setup(u = abs2(x)), Wirtinger(-im//2 * x' / u, im//2 * x / u))
@scalar_rule(angle(x::Real), Zero())
```
?

simeonschaub: I can now!

simeonschaub:
Anything still preventing this from getting merged?

oxinabox:
@simeonschaub sorry for not attending to these earlier, if we had have I think some of the problems @shashi has been running into would already be fixed.

I am away for next 2 weeks. But hopefully someone else can attend to these while I am gone.

nickrobinson251:
I think this just needs rebasing to resolve the conflicts?

simeonschaub:
Something went wrong...

oxinabox:
@simeonschaub  are you using `merge`?
You want to be using `rebase`.
So you don't get a bunch of extra commits in your PR from changes in master.

I am not great at explaining it, but googling 
"How to rebase my PR" hopefully has explaination.

simeonschaub:
@oxinabox Yes, that was probably it. I now just force-pushed it as one commit

test/rulesets/Base/base.jl
```diff
@@ -99,6 +99,18 @@
         end
     end
 
+    @testset "Unary complex functions" begin
+        for x in rand.((Int, Float32, Float64, complex.((Float32, Float64))...))
```
nickrobinson251:
```suggestion
        for x in rand.((Int32, Int64, Float32, Float64, Complex{Float32}, Complex{Float64}))
```

nickrobinson251: also, is it a problem that rand(Int) is going to give an "extreme" number
```julia
julia> rand(Int32)
507053106

julia> rand(Int64)
7317803238138782059
```

src/rulesets/Base/base.jl
```diff
@@ -83,14 +83,20 @@
 
 @scalar_rule(hypot(x, y), (x / Ω, y / Ω))
 @scalar_rule(sincos(x), @setup((sinx, cosx) = Ω), cosx, -sinx)
-@scalar_rule(atan(x, y), @setup(u = x^2 + y^2), (y / u, -x / u))
```
nickrobinson251: this is jsut renaming x -> y, y -> x, right?

simeonschaub: Yes, that corresponds better to the geometric meaning

test/rulesets/Base/base.jl
simeonschaub: Probably best just to remove the integer tests, or at least hard code them to more reasonable values

simeonschaub:
It seems like some of the complex tests need larger tolerances

simeonschaub:
@oxinabox Would it be possible to tag a new release for `ChainRulesCore`, because that seems to be what is breaking the Wirtinger tests right now?

oxinabox:
Tagged, waiting for it to merge.
https://github.com/JuliaRegistries/General/pull/3056

oxinabox:
I have restarted the CI now that the never version of ChainRulesCore has been tagged.

simeonschaub:
Should be ready to merge now

PR #67, opened 2019-08-01
Big re-organization

oxinabox:
Split core out into AbstractChainRules,
Requires-ize 3rd party packages.
Reorganize all files


oxinabox:
This needs https://github.com/JuliaDiff/AbstractChainRules.jl/pull/1

src/ChainRules.jl
```diff
@@ -13,22 +22,32 @@ if VERSION < v"1.3.0-DEV.142"
...
+include("rulesets/Base/broadcast.jl")
+include("rulesets/Base/mapreduce.jl")
+
+include("rulesets/LinearAlgebra/utils.jl")
+include("rulesets/LinearAlgebra/blas.jl")
+include("rulesets/LinearAlgebra/dense.jl")
+include("rulesets/LinearAlgebra/structured.jl")
+include("rulesets/LinearAlgebra/factorization.jl")
+
+# Note: The following is only required because package authors do not use
```
willtebbutt: Grammar. Not entirely clear on what you mean here.

src/rulesets/packages/SpecialFunctions.jl
```diff
@@ -1,3 +1,8 @@
+module SpecialFunctionsGlue
```
willtebbutt: Is `SpecialFunctionsGlue` the best name for this? Perhaps `SpecialFunctionsRules` not more appropriate?

oxinabox: One of the best things about making thes internal is not needing to stress too much over the name.
I'm inclide to leave it as is.
It is an awkward thing so gets an awkward name

test/helper_functions.jl
```diff
@@ -0,0 +1,33 @@
+@testset "helper functions" begin
+    # Hits fallback, since we can't update `Diagonal`s in place
```
mattBrzezinski: These blocks of code (e.g. `#2-6, 8:11, etc.`) should we re-written as nested `@test_sets`, this would increase readability, begin moving to tests being independent of each other, allow better documentation, and simplify running independent blocks.

test/runtests.jl
```diff
@@ -1,30 +1,35 @@
-# TODO: more tests!
-
 using ChainRules, Test, FDM, LinearAlgebra, LinearAlgebra.BLAS, Random, Statistics
-using ChainRules: extern, accumulate, accumulate!, store!, @scalar_rule,
+using Base.Broadcast: broadcastable
+import LinearAlgebra: dot
```
mattBrzezinski: Do we not also need to include an `if` check here for which version of Julia we are running?

> https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/ChainRules.jl#L9

oxinabox: I don't think so. As we did not `using Blas`

src/ChainRules.jl
nickrobinson251: updated

test/helper_functions.jl
oxinabox: 8-16 probably should be a single testsetset.

But in general yes.

nickrobinson251: will do :)

nickrobinson251:
Tests pass for me locally. Waiting on https://github.com/JuliaRegistries/General/pull/2473

test/helper_functions.jl
mattBrzezinski: Looks good, ship it!

nickrobinson251:
fyi ChainRulesCore is now registered (with v0.1.0)

PR #73, opened 2019-08-03
Bump travis tests

willtebbutt:
Make travis run on more versions of Julia


willtebbutt:
I've added 1.3 to allowed failures (hopefully correctly) as `Cassette` doesn't appear to be working on 1.3 at the minute.

PR #74, opened 2019-08-08
Fix hypot frule

YingboMa:
As `d/dx hypot(x,y) = x/hypot(x, y)` and `d/dy hypot(x,y) = y/hypot(x, y)`, I think that the rule for `hypot` was incorrect.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -62,7 +62,7 @@
 @scalar_rule(cotd(x), -(π / oftype(x, 180)) * (1 + Ω^2))
 @scalar_rule(sech(x), -tanh(x) * Ω)
 @scalar_rule(csch(x), -coth(x) * Ω)
-@scalar_rule(hypot(x, y), (y / Ω, x / Ω))
+@scalar_rule(hypot(x, y), (x / Ω, y / Ω))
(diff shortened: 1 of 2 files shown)


willtebbutt:
I'm slightly confused as to why [our tests](https://github.com/JuliaDiff/ChainRules.jl/blob/e24c190e2a82c4651266ba994df7c124cf44cf33/test/rulesets/Base/base.jl#L82) didn't pick up on this. Any thoughts @YingboMa ?

YingboMa:
The tests were also wrong https://github.com/JuliaDiff/ChainRules.jl/pull/74/files#diff-0097e516f82afcd0ed09884b0e458112R86

willtebbutt:
Eurgh, that's frustrating. And this is why we should be testing these with finite differences... many thanks for fixing this

oxinabox:
Can we add FDM based tests?
using the functions for `test/test_utils.jl`

willtebbutt:
> Can we add FDM based tests?

I'd rather just merge this and open a good-first-issue issue for this.

oxinabox:
@willtebbutt ok.  Will do

oxinabox:
This now aligns with DiffRules
https://github.com/JuliaDiff/DiffRules.jl/blob/a9b00c81ac0d103888b124b2b9f58a7c815cebf9/src/rules.jl#L94

PR #77, opened 2019-08-09
Restrict the scalar multiplication rules to `Number`s

shashi:
Before this, e.g. `frule(*, [1,2], Dual(1))` would return a non-nothing. This change helps avoid special casing this in a generic implementation of scalar diff.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -74,9 +74,9 @@
 
 # product rule requires special care for arguments where `mul` is non-commutative
 
-frule(::typeof(*), x, y) = x * y, Rule((Δx, Δy) -> Δx * y + x * Δy)
+frule(::typeof(*), x::Number, y::Number) = x * y, Rule((Δx, Δy) -> Δx * y + x * Δy)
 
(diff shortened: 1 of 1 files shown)


oxinabox:
Will merge once CI passes.

nickrobinson251:
Shall we add a test, eg that the example in the OP throws a MethodError, just so this doesn't get reverted?

oxinabox:
> Shall we add a test, eg that the example in the OP throws a MethodError, just so this doesn't get reverted?

It doesn't seem needful, and it is hard to qualify exactly what that case is, except via very much special casing.
E,g, there are (2?)  other Dual numbers that *are* subtypes of `Number` in julia.
So that test would boil down to 

```
struct MyNonNumber
x
end

@testset "Don't just do scalar multiplication without any type constraints (#77)" begin
    @test nothing === frule(*, [1,2], MyNonNumber(1))
end
```

@shashi  do you think we should add that?
I guess now that I have written it...

willtebbutt:
Would be good to have the various permutations of @oxinabox 's suggestion for non-number arguments, as well as for the `rrule`

oxinabox:
could add it to `test/test_utils.jl`

oxinabox:
Ok I have openned an issue, re:tests
So can merge this

PR #78, opened 2019-08-09
add rules for constant returning zero and one function

oxinabox:
Is this right?
Should I have some type constraints somewhere 
or does `@scalar_rule` give me that for free?

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -112,7 +112,7 @@ end
 
             y, rule = rrule(f, x)
             @test y == expected
-            @test extern(rule(1))==0.0
+            @test extern(rule(1)) == 0.0
         end
         test_constant(one, 5, 1)
         test_constant(one, -4.1, 1)
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -103,5 +103,22 @@ end
         rrule_test(identity, randn(rng), (randn(rng), randn(rng)))
         rrule_test(identity, randn(rng, 4), (randn(rng, 4), randn(rng, 4)))
     end
+
+    @testset "Constants" begin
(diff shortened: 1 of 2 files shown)


willtebbutt:
LGTM, not sure about the type constraints though.

oxinabox:
@willtebbutt  can you approve if you approve?

test/rulesets/Base/base.jl
```diff
@@ -103,5 +103,22 @@ end
         rrule_test(identity, randn(rng), (randn(rng), randn(rng)))
         rrule_test(identity, randn(rng, 4), (randn(rng, 4), randn(rng, 4)))
     end
+
+    @testset "Constants" begin
+        function test_constant(f, x, expected)
```
nickrobinson251: is there any other use-case for this? that would warrent it moving to `test/test_utils.jl`?

test/rulesets/Base/base.jl
```diff
@@ -103,5 +103,22 @@ end
...
+            y, rule = frule(f, x)
+            @test y == expected
+            @test extern(rule(1))==0.0
```
nickrobinson251: nitpick: 
```suggestion
            @test extern(rule(1)) == 0.0
```

test/rulesets/Base/base.jl
```diff
@@ -103,5 +103,22 @@ end
...
+
+            y, rule = rrule(f, x)
+            @test y == expected
+            @test extern(rule(1))==0.0
```
nickrobinson251:
```suggestion
            @test extern(rule(1)) == 0.0
```

test/rulesets/Base/base.jl
oxinabox: For now I am happy to leave it here.
We can move it later.

PR #79, opened 2019-08-09
Fix atan2 rules

YingboMa:
The rules for `atan2` were wrong.

https://github.com/JuliaDiff/DiffRules.jl/blob/a9b00c81ac0d103888b124b2b9f58a7c815cebf9/src/rules.jl#L92


oxinabox:
Can you add tests based on Finite Differencing?
Helper functions are in `test/test_utils.jl`

YingboMa:
Can I do that in a separate PR?

oxinabox:
Sure.
I thought `test_scalar` did this. But it seems not.

We should change that.
And then use the improved  `@test_scalar`  here

oxinabox:
I have checked this against Wikipedia. So I am satisfied.

PR #80, opened 2019-08-09
WIP: adding more rules and fixing old ones

shashi:
Work in progress for comments!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.2.3"
+version = "0.2.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 ChainRulesCore = "0.4"
 FiniteDifferences = "^0.7"
 Reexport = "0.2"
-Requires = "0.5.2"
+Requires = "0.5.2, 1"
 julia = "^1.0"
 
 [extras]
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -36,7 +36,7 @@
 
 @scalar_rule(sinh(x), cosh(x))
 @scalar_rule(cosh(x), sinh(x))
-@scalar_rule(tanh(x), sech(x)^2)
+@scalar_rule(tanh(x), 1-Ω^2)
 @scalar_rule(coth(x), -(csch(x)^2))
 
 @scalar_rule(asinh(x), inv(sqrt(x^2 + 1)))
@@ -66,7 +66,7 @@
 @scalar_rule(-(x, y), (One(), -1))
 @scalar_rule(/(x, y), (inv(y), -(x / y / y)))
 @scalar_rule(\(x, y), (-(y / x / x), inv(x)))
-@scalar_rule(^(x, y), (y * x^(y - 1), Ω * log(x)))
+@scalar_rule(^(x, y), (ifelse(iszero(y), Zero(), y * x^(y - 1)), Ω * log(x)))
 
 @scalar_rule(inv(x), -Ω^2)
 @scalar_rule(sqrt(x), inv(2 * Ω))
@@ -92,10 +92,12 @@
 
 @scalar_rule(max(x, y), @setup(gt = x > y), (gt, !gt))
 @scalar_rule(min(x, y), @setup(gt = x > y), (!gt, gt))
-@scalar_rule(mod(x, y), @setup((u, nan) = promote(x / y, NaN16)),
+@scalar_rule(mod(x, y), @setup((u, nan) = promote(x / y, NaN16), isint = isinteger(x / y)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -floor(u))))
-@scalar_rule(rem(x, y), @setup((u, nan) = promote(x / y, NaN16)),
+@scalar_rule(rem(x, y), @setup((u, nan) = promote(x / y, NaN16), isint = isinteger(x / y)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -trunc(u))))
+@scalar_rule(fma(x, y, z), (y, x, One()))
+@scalar_rule(muladd(x, y, z), (y, x, One()))
 @scalar_rule(angle(x::Complex), @setup(u = abs2(x)), Wirtinger(-im//2 * x' / u, im//2 * x / u))
 @scalar_rule(angle(x::Real), Zero())
 @scalar_rule(real(x::Complex), Wirtinger(1//2, 1//2))
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -66,7 +66,7 @@
 @scalar_rule(-(x, y), (One(), -1))
 @scalar_rule(/(x, y), (inv(y), -(x / y / y)))
 @scalar_rule(\(x, y), (-(y / x / x), inv(x)))
-@scalar_rule(^(x, y), (ifelse(iszero(y), Zero(), y * x^(y - 1)), Ω * log(x)))
+@scalar_rule(^(x, y), (ifelse(iszero(y), zero(Ω), y * x^(y - 1)), Ω * log(x)))
 
 @scalar_rule(inv(x), -Ω^2)
 @scalar_rule(sqrt(x), inv(2 * Ω))
--- a/src/rulesets/packages/NaNMath.jl
+++ b/src/rulesets/packages/NaNMath.jl
@@ -1,6 +1,7 @@
 module NaNMathGlue
 using ChainRulesCore
 using ..NaNMath
+using ..SpecialFunctions
 
 @scalar_rule(NaNMath.sin(x), NaNMath.cos(x))
 @scalar_rule(NaNMath.cos(x), -NaNMath.sin(x))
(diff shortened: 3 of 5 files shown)


src/rulesets/Base/base.jl
```diff
@@ -67,16 +67,17 @@
 @scalar_rule(atan(x, y), @setup(u = x^2 + y^2), (y / u, -x / u))
 @scalar_rule(max(x, y), @setup(gt = x > y), (gt, !gt))
 @scalar_rule(min(x, y), @setup(gt = x > y), (!gt, gt))
-@scalar_rule(mod(x, y), @setup((u, nan) = promote(x / y, NaN16)),
+@scalar_rule(mod(x, y), @setup((u, nan) = promote(x / y, NaN16), isint = isinteger(x / y)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -floor(u))))
-@scalar_rule(rem(x, y), @setup((u, nan) = promote(x / y, NaN16)),
+@scalar_rule(rem(x, y), @setup((u, nan) = promote(x / y, NaN16), isint = isinteger(x / y)),
              (ifelse(isint, nan, one(u)), ifelse(isint, nan, -trunc(u))))
+@scalar_rule(fma(x, y, z), (y, x, One()))
```
oxinabox: Add `muladd` also with same definition?

src/rulesets/packages/NaNMath.jl
```diff
@@ -1,9 +1,11 @@
 module NaNMathGlue
 using ChainRulesCore
 using NaNMath
+using SpecialFunctions
```
oxinabox: Maaan we needed tests for this file

src/rulesets/Base/base.jl
```diff
@@ -67,16 +67,17 @@
...
 
 # product rule requires special care for arguments where `mul` is non-commutative
 
-frule(::typeof(*), x, y) = x * y, Rule((Δx, Δy) -> Δx * y + x * Δy)
+frule(::typeof(*), x::Number, y::Number) = x * y, Rule((Δx, Δy) -> Δx * y + x * Δy)
 
-rrule(::typeof(*), x, y) = x * y, (Rule(ΔΩ -> ΔΩ * y'), Rule(ΔΩ -> x' * ΔΩ))
+rrule(::typeof(*), x::Number, y::Number) = x * y, (Rule(ΔΩ -> ΔΩ * y'), Rule(ΔΩ -> x' * ΔΩ))
```
simeonschaub: It's important to think about the complex case here too. All other `rrule`s don't return the complex conjugate of the derivative, so either the conjugate transposes should be changed to unconjugated transposes, or this convention needs to be changed everywhere else

src/rulesets/Base/base.jl
```diff
@@ -45,7 +45,7 @@
 @scalar_rule(-(x, y), (One(), -1))
 @scalar_rule(/(x, y), (inv(y), -(x / y / y)))
 @scalar_rule(\(x, y), (-(y / x / x), inv(x)))
-@scalar_rule(^(x, y), (y * x^(y - 1), Ω * log(x)))
+@scalar_rule(^(x, y), (ifelse(iszero(y), Zero(), y * x^(y - 1)), Ω * log(x)))
```
simeonschaub: This just seems like it unnecessarily introduces type instabilities. Is there a particular reason you need this?

simeonschaub: A better way to do this would be to overload `Base.literal_pow`. Have a look at how it is implemented in Julia base, it seems to be exactly for purposes like this.

shashi: I thought this type instability is not a big deal because of union splitting? Also it seems to be the convention everywhere in ChainRules to return `Zero()` instead of trying to figure out the right zero in the rules.

I'll defer to @YingboMa about `Base.literal_pow`

src/rulesets/Base/base.jl
```diff
@@ -24,7 +24,7 @@
 @scalar_rule(acotd(x), -oftype(x, 180) / π / (1 + x^2))
 @scalar_rule(sinh(x), cosh(x))
 @scalar_rule(cosh(x), sinh(x))
-@scalar_rule(tanh(x), sech(x)^2)
+@scalar_rule(tanh(x), 1-Ω^2)
```
oxinabox: This is correct.

src/rulesets/Base/base.jl
oxinabox: Small union optimization is better, but still has some slowdown.

Important thing about `Zero()`  is it's ability to remove `Thunks` via multiplication before it is evaluated.
Idk how often that occurs.

I think this this is a bit of a premature optimisation unless you have benchmarks showing it helps.

src/rulesets/packages/NaNMath.jl
```diff
@@ -1,9 +1,11 @@
...
 
 @scalar_rule(NaNMath.sin(x), NaNMath.cos(x))
 @scalar_rule(NaNMath.cos(x), -NaNMath.sin(x))
+@scalar_rule(NaNMath.tan(x), 1 + NaNMath.pow(NaNMath.tan(x), 2))
```
oxinabox: This is wrong I think?
Can you cite for this?

src/rulesets/Base/base.jl
simeonschaub: Also note that there is no promotion mechanism for `Zero` and `One` yet, so user code might break if an unexpected `Zero` or `One` is returned

nickrobinson251:
Needs tests?

src/rulesets/Base/base.jl
oxinabox: That is not a problem. One should always be calling `extern` if you are not prepared for a Differential to come back.

src/rulesets/packages/NaNMath.jl
simeonschaub: Wikipedia has the same identity here: [https://en.m.wikipedia.org/wiki/Trigonometric_functions](https://en.m.wikipedia.org/wiki/Trigonometric_functions). But why not just use `1 + Ω^2`? `NaN` should propagate through that just fine.

src/rulesets/Base/base.jl
YingboMa: We need this branch, otherwise, exponentiation of zero is arguably wrong. See https://github.com/JuliaDiff/ForwardDiff.jl/commit/123cc19f081ac83c38b27d2c7a5e141217717fa3

simeonschaub: Ah, ok, I see. So the problem is that integers to a negative integer power throw a domain error. But for the purpose of type stability, `Zero()` should probably be replaced by `zero(x*y)`.

src/rulesets/packages/NaNMath.jl
shashi: I copied this from DiffRules

src/rulesets/Base/base.jl
shashi: `Zero()` is the convention in the rest of the package. But we can surely make an exception and forget about it.

oxinabox: I am happy to go with `Zero`

src/rulesets/packages/NaNMath.jl
oxinabox: I think we should change it to `1 + Ω^2`
unless there is a reason not to?

oxinabox:
@shashi  what is the status with this branch?

shashi:
I'll not be able to work on this till Sep 1. Will address the comments and add tests.

src/rulesets/packages/NaNMath.jl
oxinabox: Fixed on master

willtebbutt:
@shashi reckon you'll be able to finish this off at some point? Would be really great to have this merged.

src/rulesets/Base/base.jl
simeonschaub: I've had another look into this and am worrying that this introduced type instability will cause significant overhead and bite us in the long run. I would imagine that an exponent of zero is really an edge case and won't occur too often in real code, so I'm wondering if not materializing that thunk in these cases is worth the trade-off of a significant performance hit in all other cases. Some quick benchmarks imply an up to a 25x(!) penalty with this instability (if I have overlooked something in my benchmarking methodology, please do correct me):
```julia
julia> f1(x, y) = ifelse(iszero(y), Zero(), y * x^(y - 1));

julia> f2(x, y) = ifelse(iszero(y), zero(x^y), y * x^(y - 1));

julia> a = rand(1000);

julia> @btime f1.($a, 2);
  24.201 μs (5 allocations: 8.05 KiB)

julia> @btime f2.($a, 2);
  787.900 ns (1 allocation: 7.94 KiB)
```

nickrobinson251: Seems like we should change to `zero` and add a comment?

Project.toml
```diff
@@ -10,10 +10,10 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-Reexport = "0.2"
-Requires = "0.5.2"
 ChainRulesCore = "0.4"
 FiniteDifferences = "^0.7"
+Reexport = "0.2"
+Requires = "0.5.2"
```
oxinabox: Prob allow 0.5.2 or 1

src/rulesets/Base/base.jl
oxinabox: @shashi are you happy to do that?

oxinabox:
I'ld actually really like the tests
Tests for scalar rules are really easy.
Can you add them?

src/rulesets/Base/base.jl
YingboMa: What about
```
@scalar_rule(^(x, y), (ifelse(iszero(y), zero(Ω), y * x^(y - 1)), Ω * log(x)))
```
?

Floating-point exponentiation is very expensive.

Project.toml
mattBrzezinski: Is there a reason why we're locking to `0.5.2` specifically?

src/rulesets/packages/NaNMath.jl
```diff
@@ -15,5 +16,11 @@ using ..NaNMath
 @scalar_rule(NaNMath.lgamma(x), SpecialFunctions.digamma(x))
 @scalar_rule(NaNMath.sqrt(x), inv(2 * Ω))
 @scalar_rule(NaNMath.pow(x, y), (y * NaNMath.pow(x, y - 1), Ω * NaNMath.log(x)))
+@scalar_rule(NaNMath.max(x, y),
```
mattBrzezinski: Can we add tests for each new rule being added?

src/rulesets/Base/base.jl
oxinabox: Yeah, I agree that that is legit.

(For my own futre reference:
```
julia> NaN^0.0
1.0
```
so this is correct, even then.
)

YingboMa:
I will add tests in a separate PR.

src/rulesets/Base/base.jl
shashi: I'm ok with `zero` or anything that is fast !

oxinabox:
> I will add tests in a separate PR.

Can we add the tests for the functions in `src/Base/base.jl` in this PR?  
I am less fussed about the ones for NaNMath, as we don't have any tests there for that.
But for `base.jl` we have basically complete coverage.

YingboMa:
Sure thing.

YingboMa:
Don't merge this yet. Git history is a bit messed up.

YingboMa:
Nevermind, it seems like a GitHub glitch.

oxinabox:
Merge when you are ready and do remember to tag a release when you do

PR #82, opened 2019-08-11
Fix documentation to use ChainRulesCore

nickrobinson251:
1. Right now the API docs don't render because `ChainRules.foo` has been moved to `ChainRulesCore.foo`
2. This PR essentially means we have a single set of docs for ChainRules and ChainRules core (https://github.com/JuliaDiff/ChainRulesCore.jl/issues/12)

Before:
image

After:
image

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,24 +1,5 @@
-# ChainRules API Documentation
+# API Documentation
 
-```@docs
-ChainRules.frule
-ChainRules.rrule
-ChainRules.AbstractRule
-ChainRules.Rule
-ChainRules.DNERule
-ChainRules.WirtingerRule
-ChainRules.accumulate
-ChainRules.accumulate!
-ChainRules.store!
-```
-
-```@docs
-ChainRules.AbstractDifferential
-ChainRules.extern
-ChainRules.Casted
-ChainRules.Wirtinger
-ChainRules.Thunk
-ChainRules.Zero
-ChainRules.DNE
-ChainRules.One
+```@autodocs
+Modules = [ChainRulesCore, ChainRules]
 ```
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,5 +1,5 @@
 # API Documentation
 
 ```@autodocs
-Modules = [ChainRulesCore, ChainRules]
+Modules = [ChainRulesCore]
 ```
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -4,7 +4,7 @@
 It only depends on [Cassette.jl](https://github.com/jrevels/Cassette.jl) which is also light-weight and has no dependencies.
 
 [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl) provides the full functionality, including sensitivities for Base Julia and standard libraries.
-Sensitivities for some other packages, currently SpecialFunctions.jl and NanMath.jl, will also be loaded if those packages are in your environment.
+Sensitivities for some other packages, currently SpecialFunctions.jl and NaNMath.jl, will also be loaded if those packages are in your environment.
 In general, we recommend adding custom sensitivities to your own packages with ChainRulesCore.
 
 ## Defining Custom Sensitivities
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -1,8 +1,11 @@
 # Getting Started
 
 [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl) is a light-weight dependency for defining sensitivities for functions in your packages, without you needing to depend on ChainRules itself.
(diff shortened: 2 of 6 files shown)


nickrobinson251:
Since most of the docstring are actually in ChainRulesCore.jl, and many contain code that could be `doctest`ed... should we instead have the documentation live in the ChainRulesCore repo, so doctests run on PRs there rather than here?

docs/src/getting_started.md
```diff
@@ -1,5 +1,13 @@
 # Getting Started
 
+[ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl) is a light-weight dependency for defining sensitivities for functions in your packages, without you needing to depend on ChainRules itself.
```
willtebbutt: Maybe emphasise that this package literally only has one dependency.

docs/src/getting_started.md
```diff
@@ -1,5 +1,13 @@
...
+
+[ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl) provides the full functionality, including sensitivities for Base Julia and standard libraries.
```
willtebbutt: Maybe also point out here that we load other sensitivities for other non-stdlib packages e.g. SpecialFunctions and nanmath on demand.

docs/src/api.md
```diff
@@ -1,24 +1,5 @@
...
-ChainRules.AbstractDifferential
-ChainRules.extern
-ChainRules.Casted
-ChainRules.Wirtinger
-ChainRules.Thunk
-ChainRules.Zero
-ChainRules.DNE
-ChainRules.One
+```@autodocs
+Modules = [ChainRulesCore, ChainRules]
```
willtebbutt: I think this might wind up being a bit general. Just built the docs and the API section now includes some very cholesky-specific stuff that it really probably shouldn't. Is there a way to exclude unwanted docstrings when doing this?

nickrobinson251: we can set `private = false` to ignore non-exported stuff... but for some reason when i tried this locally the `chol_*` and `level*` stuff still seems to get included

For now ChainRules doesn't define any user-facing behaviour AFAIU, so I might just exclude it altogether and only just do `Modules = [ChainRulesCore]`

docs/src/getting_started.md
nickrobinson251: sure, will add

nickrobinson251:
> Since most of the docstring are actually in ChainRulesCore.jl, and many contain code that could be `doctest`ed... should we instead have the documentation live in the ChainRulesCore repo, so doctests run on PRs there rather than here?

Any opinions either way on this, @willtebbutt ?

Thanks for reviewing!

docs/src/getting_started.md
```diff
@@ -1,5 +1,16 @@
 # Getting Started
 
+[ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl) is a light-weight dependency for defining sensitivities for functions in your packages, without you needing to depend on ChainRules itself.
+It only depends on [Cassette.jl](https://github.com/jrevels/Cassette.jl) which is also light-weight and has no dependencies.
+
+[ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl) provides the full functionality, including sensitivities for Base Julia and standard libraries.
+Sensitivities for some other packages, currently SpecialFunctions.jl and NanMath.jl, will also be loaded if those packages are in your environment.
```
nickrobinson251:
```suggestion
Sensitivities for some other packages, currently SpecialFunctions.jl and NaNMath.jl, will also be loaded if those packages are in your environment.
```

willtebbutt:
> Any opinions either way on this, @willtebbutt ?

Probably makes sense.

nickrobinson251:
> > Any opinions either way on this, @willtebbutt ?
> 
> Probably makes sense.

TL;DR: Scrap that. Won't work. I think this is good to go as is. :)

It turns out that the docstrings in ChainRulesCore  (e.g. for `frule`/`rrule`) rely on ChainRules (or at least, there being a rule defined). This isn't really a problem, and the docstring examples are good, but we should figure out a way to run the doctests, because one of the docstring examples was wrong (it uses hypot but was not updated with #55). 

I thought I could run do `doctest(ChainRulesCore)` in `ChainRules.jl/test/runtests`, but that gave me `ArgumentError: Package ChainRulesCore does not have a documentation source directory at standard location.` now that ChainRulesCore doesn't have docs (or won't, after it shares these docs).

docs/src/api.md
```diff
@@ -1,24 +1,5 @@
...
-ChainRules.AbstractDifferential
-ChainRules.extern
-ChainRules.Casted
-ChainRules.Wirtinger
-ChainRules.Thunk
-ChainRules.Zero
-ChainRules.DNE
-ChainRules.One
+```@autodocs
+Modules = [ChainRulesCore]
```
ararslan: The downside to this is that you lose the ability to order things explicitly, but :man_shrugging:

Issue #86, opened 2019-08-23
Rules for dervative of asec and acsc with complex input incorrect? 


oxinabox:
Found when doing #76 

### `asec`
```
x = 1.6 - 0.8im
(central_fdm(5, 1))(asec, x) = 0.16078185380606091 + 0.29837305540037473im
(last(frule(asec, x)))(1) = 0.27724414876919296 + 0.19496914288010564im
```
[WolframAlpha says](https://www.wolframalpha.com/input/?i=derivative+of+asec%28x%29+at+x%3D%281.6+-+0.8i%29) `0.160782 + 0.298373 i`
Argreeing with the `central_fdm`

### `acsc`
```
x = 1.6 - 0.8im
(central_fdm(5, 1))(acsc, x) = -0.16078185380578297 - 0.29837305540024617im
(last(frule(acsc, x)))(1) = -0.27724414876919296 - 0.19496914288010564im
```
[WolframAlpha says](https://www.wolframalpha.com/input/?i=derivative+of+acsc%28x%29+at+x%3D%281.6+-+0.8i%29) `-0.160782 - 0.298373 i`
Argreeing with the `central_fdm`

@simeonschaub  do you know what is going on here?
It has been years since I did complex calculus

oxinabox:
Investigating further.
This is basically because we use a simplification of the derviative that only works for real numbers.
So it is just wrong in the general case.

oxinabox:
I have a fix, and will include it in my PR fro #76

PR #87, opened 2019-08-23
Add FiniteDifferences based scalar rule tests

### LINKED ISSUE #76: Add FiniteDifferences based tests to some scalar functions
As found for `hypot` 
https://github.com/JuliaDiff/ChainRules.jl/pull/74
we are not testing some of our simpler scalar rules with FiniteDifferences.
We have easy helpers for how to do that in `test/test_utils.jl`
but we just are not using them

oxinabox:
Closes #76
and the mistakes found in #86 for incorrect acshc asec acsc, asecd acscd rules

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -1,66 +1,56 @@
-function test_scalar(f, f′, xs...)
-    for r = (rrule, frule)
-        rr = r(f, xs...)
-        @test rr !== nothing
-        fx, ∂x = rr
-        @test fx == f(xs...)
-        @test ∂x(1) ≈ f′(xs...) atol=1e-5
-    end
-end
-
 @testset "base" begin
     @testset "Trig" begin
-        @testset "Basics" for x = (Float64(π), Complex(π, π/2))
-            test_scalar(sin, cos, x)
-            test_scalar(cos, x -> -sin(x), x)
-            test_scalar(tan, x -> 1 + tan(x)^2, x)
-            test_scalar(sec, x -> sec(x) * tan(x), x)
-            test_scalar(csc, x -> -csc(x) * cot(x), x)
-            test_scalar(cot, x -> -1 - cot(x)^2, x)
-            test_scalar(sinpi, x -> π * cospi(x), x)
-            test_scalar(cospi, x -> -π * sinpi(x), x)
+        @testset "Basics" for x = (Float64(π)-0.01, Complex(π, π/2))
+            test_scalar(sin, x)
+            test_scalar(cos, x)
+            test_scalar(tan, x)
+            test_scalar(sec, x)
+            test_scalar(csc, x)
+            test_scalar(cot, x)
+            test_scalar(sinpi, x)
+            test_scalar(cospi, x)
         end
-        @testset "Hyperbolic" for x = (Float64(π), Complex(π, π/2))
-            test_scalar(sinh, cosh, x)
-            test_scalar(cosh, sinh, x)
-            test_scalar(tanh, x -> sech(x)^2, x)
-            test_scalar(sech, x -> -tanh(x) * sech(x), x)
-            test_scalar(csch, x -> -coth(x) * csch(x), x)
-            test_scalar(coth, x -> -csch(x)^2, x)
+        @testset "Hyperbolic" for x = (Float64(π)-0.01, Complex(π-0.01, π/2))
+            test_scalar(sinh, x)
+            test_scalar(cosh, x)
+            test_scalar(tanh, x)
+            test_scalar(sech, x)
+            test_scalar(csch, x)
+            test_scalar(coth, x)
         end
         @testset "Degrees" begin
             x = 45.0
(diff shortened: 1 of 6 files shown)


test/test_util.jl
```diff
@@ -1,13 +1,43 @@
...
+    @testset "$f at $x, $(nameof(rule))" for rule in (rrule, frule)
+        res = rule(f, x)
+        @test res !== nothing  # Check the rule was defined
+        fx, ∂x = res
+        @test fx == f(x)  # Check we still get the normal value, right
+
+        # Check that we get the derivative right:
+        @test isapprox(
+            ∂x(1), fdm(f, x);
+            rtol=rtol, atol=atol, kwargs...
```
willtebbutt: Do we need to have `atol` and `rtol` outside of `kwargs`? If we don't set the defaults to `1e-9`, do we start to get errors?

oxinabox:
> Assuming that we now have 100% coverage of the scalars rules through the use of test_scalar, I'm happy for this to be merged.

We don't yet.
I''m going to add a few more commits to get close to that

test/test_util.jl
oxinabox: idk, I was copying the `test_frule` and `test_rrule` setup

nickrobinson251:
This is great

oxinabox:
we got the derviative of `inv(x)` wrong...

willtebbutt:
> we got the derviative of inv(x) wrong...

lol

oxinabox:
OK.
Done and ready for review

test/rulesets/Base/base.jl
```diff
@@ -1,66 +1,55 @@
...
+            test_scalar(cospi, x)
         end
-        @testset "Hyperbolic" for x = (Float64(π), Complex(π, π/2))
-            test_scalar(sinh, cosh, x)
-            test_scalar(cosh, sinh, x)
-            test_scalar(tanh, x -> sech(x)^2, x)
-            test_scalar(sech, x -> -tanh(x) * sech(x), x)
-            test_scalar(csch, x -> -coth(x) * csch(x), x)
-            test_scalar(coth, x -> -csch(x)^2, x)
+        @testset "Hyperbolic" for x = (Float64(π)-0.01, Complex(π-0.01, π/2))
```
nickrobinson251: out of interest, why this change?

test/rulesets/Base/base.jl
```diff
@@ -83,40 +72,63 @@ end
...
         end
-    end
-    @testset "Misc. Tests" begin
-        @testset "*(x, y)" begin
-            x, y = rand(3, 2), rand(2, 5)
-            z, (dx, dy) = rrule(*, x, y)
+    end  # Trig
 
-            @test z == x * y
+    @testset "math" begin
```
nickrobinson251: i like that the rest isn't "math" 😆

test/rulesets/Base/base.jl
```diff
@@ -83,40 +72,63 @@ end
...
 
-            @test dx(z̄) == extern(accumulate(zeros(3, 2), dx, z̄))
-            @test dy(z̄) == extern(accumulate(zeros(2, 5), dy, z̄))
+            test_scalar(exp, x)
+            test_scalar(exp2, x)
+            test_scalar(exp10, x)
 
-            test_accumulation(rand(3, 2), dx, z̄, z̄ * y')
-            test_accumulation(rand(2, 5), dy, z̄, x' * z̄)
+            x isa Real && test_scalar(cbrt, x)
```
nickrobinson251: TIL there is a `cbrt` function

nickrobinson251:
we should re-run CI for Julia 1.0?

test/rulesets/Base/base.jl
oxinabox: Movew it away from the boundries where it isn't defined and we thus can;t use `central_fdm`

nickrobinson251: seems fair enough to me 👍

simeonschaub:
@nickrobinson251 It seems that Travis needs to be reconfigured to not abort if tests take longer than 10 minutes

nickrobinson251:
> @nickrobinson251 It seems that Travis needs to be reconfigured to not abort if tests take longer than 10 minutes

Yes, we should extend that time limit - I don't think I have permissions to do that, but hopefully @oxinabox does? It's possible that allowing more time won't solve this case, if e.g. SpecialFunctions is just hanging for some reason on 1.1 (don't know if that's what's happening or not)

oxinabox:
It is not possible to increase the time without log output in Travis.
Plus in the versions that pass the total time is not much more than 10 minutes.
(and sometimes is much less)
So it can't be that SpecialFunctions is suposed to need 11minutes to build or something

simeonschaub:
Is the solution just to insert random print statements?

oxinabox:
> Is the solution just to insert random print statements?

Apparently.

I wonder if we should just get rid of the top-level TestSet so it prints each testset as it completes.

Issue #12, opened 2019-04-12
Broadcasting with differentiable functions (Remove Cast?)


willtebbutt:
The current implementation of [broadcast](https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/rules/broadcast.jl) assumes that the function being `broadcast`ed doesn't contain any differentiable bits, and that we can therefore safely assume that there is no gradient information to be associated with it. It also assumes that the forwards-inside-reverse-mode trick is the correct choice for implementing the adjoint, which isn't necessarily the case.

Presumably this implementation is a placeholder, however, it will definitely be necessary to relax the above assumptions before e.g. Zygote is able to adopt ChainRules, so I believe it should be addressed as a priority.

jrevels:
> The current implementation of broadcast assumes that the function being broadcasted doesn't contain any differentiable bits

Relevant dev note (will turn into an advanced usage doc section or something at some point): https://github.com/JuliaDiff/ChainRules.jl/blob/6cfd4b94926575e7c94368744b279052179f88e4/src/rules.jl#L142

> Presumably this implementation is a placeholder, however, it will definitely be necessary to relax the above assumptions before e.g. Zygote is able to adopt ChainRules, so I believe it should be addressed as a priority.

It's a half-placeholder 😉kind of expresses the idea, but the implementation is mainly a toy one compared to e.g. https://github.com/jrevels/MixedModeBroadcastAD.jl.

Regardless - unless I'm misunderstanding something, which could always be the case - none of the default rule definitions should be a barrier to adoption by downstream AD, since downstream ADs can overload rules in whatever manner best fits their specific implementations. Zygote can define the broadcast rule in whatever way makes the most sense for Zygote.

> It also assumes that the forwards-inside-reverse-mode trick is the correct choice for implementing the adjoint, which isn't necessarily the case.

It assumes that forward mode is the correct choice of "fallback" mode for unary scalar functions, which is almost certainly the case. It doesn't make any assumptions about non-unary functions, and/or functions where there's a specialized `broadcast` rule that could be defined ahead of time.

[...]

PR #90, opened 2019-08-26
test wirtinger derivatives with fdm too

simeonschaub:
This tests if the derivatives are well-defined for complex arguments, too, or if they're wrongly assumed to be holomorphic.

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -87,11 +87,13 @@
 
             x isa Real && test_scalar(cbrt, x)
             if (x isa Real && x >= 0) || x isa Complex
-                test_scalar(sqrt, x)
-                test_scalar(log, x)
-                test_scalar(log2, x)
-                test_scalar(log10, x)
-                test_scalar(log1p, x)
+                # this check is needed because the imaginary part jumps from pi to -pi
+                nr = imag(x) != 0 && real(x) < 0
+                test_scalar(sqrt, x, test_wirtinger=nr)
+                test_scalar(log, x, test_wirtinger=nr)
+                test_scalar(log2, x, test_wirtinger=nr)
+                test_scalar(log10, x, test_wirtinger=nr)
+                test_scalar(log1p, x, test_wirtinger=nr)
             end
         end
     end
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -87,13 +87,14 @@
 
             x isa Real && test_scalar(cbrt, x)
             if (x isa Real && x >= 0) || x isa Complex
-                # this check is needed because the imaginary part jumps from pi to -pi
-                nr = imag(x) != 0 && real(x) < 0
-                test_scalar(sqrt, x, test_wirtinger=nr)
-                test_scalar(log, x, test_wirtinger=nr)
-                test_scalar(log2, x, test_wirtinger=nr)
-                test_scalar(log10, x, test_wirtinger=nr)
-                test_scalar(log1p, x, test_wirtinger=nr)
+                # this check is needed because these have discontinuities between
+                # `-10 + im*eps()` and `-10 - im*eps()`
+                should_test_wirtinger = imag(x) != 0 && real(x) < 0
+                test_scalar(sqrt, x; test_wirtinger=should_test_wirtinger)
+                test_scalar(log, x; test_wirtinger=should_test_wirtinger)
+                test_scalar(log2, x; test_wirtinger=should_test_wirtinger)
+                test_scalar(log10, x; test_wirtinger=should_test_wirtinger)
+                test_scalar(log1p, x; test_wirtinger=should_test_wirtinger)
             end
         end
     end
--- a/test/test_util.jl
+++ b/test/test_util.jl
@@ -1,6 +1,7 @@
 using FiniteDifferences, Test
 using FiniteDifferences: jvp, j′vp
 using ChainRules
+using ChainRules: AbstractDifferential
 
 const _fdm = central_fdm(5, 1)
 
--- a/test/test_util.jl
+++ b/test/test_util.jl
@@ -162,7 +162,7 @@ end
 function Base.isapprox(d_ad::DNE, d_fd; kwargs...)
     error("Tried to differentiate w.r.t. a DNE")
 end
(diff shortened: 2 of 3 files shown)


simeonschaub:
It seems that Travis fails because it doesn't see any output for too long. Is there a way to change this behavior?

simeonschaub:
Not the most elegant solution, but it works...

oxinabox:
@simeonschaub  arg, sorry, I did not realize this was against my `ox/testmore` branch.
I have retargetted it against master.

test/rulesets/Base/base.jl
```diff
@@ -87,11 +87,13 @@
...
-                test_scalar(log2, x)
-                test_scalar(log10, x)
-                test_scalar(log1p, x)
+                # this check is needed because the imaginary part jumps from pi to -pi
+                nr = imag(x) != 0 && real(x) < 0
+                test_scalar(sqrt, x, test_wirtinger=nr)
+                test_scalar(log, x, test_wirtinger=nr)
+                test_scalar(log2, x, test_wirtinger=nr)
+                test_scalar(log10, x, test_wirtinger=nr)
+                test_scalar(log1p, x, test_wirtinger=nr)
```
oxinabox: kwargs should be seperated from non-kwargs by a `;`

Also can we rename `nr` to something clearer.
like `should_test_wirtinger`?
Or maybe we can move it into the definition of `test_scalar` ?

simeonschaub: I can definitely rename it. The problem with moving this into the definition of `test_scalar` is that it only applies to these particular functions since these have discontinuities between `-10 + im*eps()` and `-10 - im*eps()`.

test/test_util.jl
```diff
@@ -14,20 +14,37 @@ at input point `x` to confirm that there are correct ChainRules provided.
...
+            # For complex arguments, also check if the wirtinger derivative is correct
+            ∂Re, ∂Im = fdm.((ϵ -> f(x + ϵ), ϵ -> f(x + im*ϵ)), 0)
+            ∂ = .5(∂Re - im*∂Im)
+            ∂̅ = .5(∂Re + im*∂Im)
+            @test isapprox(
+                wirtinger_primal(∂x(1)), ∂;
+                rtol=rtol, atol=atol, kwargs...
+            )
+            @test isapprox(
+                extern(wirtinger_conjugate(∂x(1))), ∂̅;
```
nickrobinson251: Why do we need `extern` here but not for `wirtinger_primal`? That doesn't seem nice...

test/test_util.jl
```diff
@@ -14,20 +14,37 @@ at input point `x` to confirm that there are correct ChainRules provided.
...
-            rtol=rtol, atol=atol, kwargs...
-        )
+        if !test_wirtinger
+            @test isapprox(
+                ∂x(1), fdm(f, x);
+                rtol=rtol, atol=atol, kwargs...
+            )
+        else
+            # For complex arguments, also check if the wirtinger derivative is correct
+            ∂Re, ∂Im = fdm.((ϵ -> f(x + ϵ), ϵ -> f(x + im*ϵ)), 0)
```
nickrobinson251: style suggestion:

```suggestion
            ∂Re = fdm(ϵ -> f(x + ϵ), 0)
            ∂Im = fdm(ϵ -> f(x + im*ϵ), 0)
```

test/test_util.jl
```diff
@@ -14,20 +14,37 @@ at input point `x` to confirm that there are correct ChainRules provided.
...
+            ∂ = .5(∂Re - im*∂Im)
+            ∂̅ = .5(∂Re + im*∂Im)
```
nickrobinson251: style:

```suggestion
            ∂̅ = 0.5(∂Re + im*∂Im)
```

test/test_util.jl
```diff
@@ -14,20 +14,37 @@ at input point `x` to confirm that there are correct ChainRules provided.
...
-        )
+        if !test_wirtinger
+            @test isapprox(
+                ∂x(1), fdm(f, x);
+                rtol=rtol, atol=atol, kwargs...
+            )
+        else
+            # For complex arguments, also check if the wirtinger derivative is correct
+            ∂Re, ∂Im = fdm.((ϵ -> f(x + ϵ), ϵ -> f(x + im*ϵ)), 0)
+            ∂ = .5(∂Re - im*∂Im)
```
nickrobinson251: style:

```suggestion
            ∂ = 0.5(∂Re - im*∂Im)
```

test/test_util.jl
nickrobinson251: also, do we want `0` -> `zero(x)`?

simeonschaub: It should definitely always be real, so maybe `zero(real(x))` to support different float precision? Then again, do we even want to do fdm with anything less precise than `Float64`?

test/test_util.jl
simeonschaub: Because most of the time `wirtinger_conjugate` is of type `Zero`. We probably should call `extern` on `wirtinger_primal` too, since this can also in general be any `AbstractDifferential`. But maybe overload `isapprox`?

test/test_util.jl
nickrobinson251: cool - perhaps just `0.0` is fine then :)

oxinabox: We should use `0` and let FiniteDifferences sort out promotion.

test/test_util.jl
nickrobinson251: yeah, seems reasonable to me to overload `isapprox` for `AbstractDifferential`s, to `extern` them`

test/test_util.jl
```diff
@@ -144,7 +162,7 @@ end
 function Base.isapprox(d_ad::DNE, d_fd; kwargs...)
     error("Tried to differentiate w.r.t. a DNE")
 end
-function Base.isapprox(d_ad::Thunk, d_fd; kwargs...)
+function Base.isapprox(d_ad::Union{Thunk,One,Zero}, d_fd; kwargs...)
```
oxinabox:
```suggestion
function Base.isapprox(d_ad::AbstractDifferential, d_fd; kwargs...)
```

In general `extern` is what one does to an `AbstractDifferential` to get a numeric value.
And we have covered the special cases.
And `Casted` is its own ball of fun #12  that can be worried about another day

simeonschaub:
Ready to merge?

PR #93, opened 2019-08-29
Remove uses of add and mul from tests because they have been removed ChainRulesCore

oxinabox:
We access some of the internals of ChainRulesCore for testing purposes.

This changes that over to match the changes in
https://github.com/JuliaDiff/ChainRulesCore.jl/pull/35

That PR has to be merged first.

I have tested these together and they do work together.
Passing all tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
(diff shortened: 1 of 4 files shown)


nickrobinson251:
since docs for both live here

Can you also remove reference to Cassette. 

I think there is just one 
https://github.com/JuliaDiff/ChainRules.jl/blame/master/docs/src/getting_started.md#L4

oxinabox:
Just waiting for 
https://github.com/JuliaRegistries/General/pull/3076
to be tagged so can retest and merge this

PR #95, opened 2019-08-31
Fix rules for `*`

simeonschaub:
Our derivatives for multiplication are wrong for complex numbers. This also adds `rrule` and `frule` for multiplication of matrices with scalars or vectors and for scalar-vector. Still needs more tests.


simeonschaub:
Tests need JuliaDiff/FiniteDifferences.jl#43

oxinabox:
we should rebase and bring this up to code and merge this

PR #96, opened 2019-09-02
Switch constants over to use test_scalar

oxinabox:
Simplifies out testing infastructure slightly

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -152,7 +152,7 @@
         rrule_test(identity, randn(rng, 4), (randn(rng, 4), randn(rng, 4)))
     end
 
-    @testset "Constants" for x in (-0.1, 6.4, 1.0+0.5im, -10.0+0im)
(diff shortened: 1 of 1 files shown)


simeonschaub:
Can we have a test with complex input too? Shouldn't really be a problem,  but just to be safe.

test/rulesets/Base/base.jl
```diff
@@ -152,11 +152,8 @@
         rrule_test(identity, randn(rng, 4), (randn(rng, 4), randn(rng, 4)))
     end
 
-    @testset "Constants" begin
-        test_scalar(one, 5)
-        test_scalar(one, -4.1)
-
-        test_scalar(zero, 5)
-        test_scalar(zero, -4.1)
+    @testset "Constants" for x in (-0.1, 6.4, 1.0+0.5im, -10.0+0im)
```
willtebbutt: For symmetry's sake, perhaps add a purely imaginary test?

oxinabox:
```suggestion
    @testset "Constants" for x in (-0.1, 6.4, 1.0+0.5im, -10.0+0im, 0+200im)
```

PR #98, opened 2019-09-02
Fix abs(::Complex)

### LINKED ISSUE #88: stop defining derviative of `abs(::Complex)`?
Currently it `sgn` which is apparently not correct.

[Wikipedia](https://en.wikipedia.org/wiki/Absolute_value#Derivative)

> The complex absolute value function is continuous everywhere but complex differentiable nowhere because it violates the Cauchy–Riemann equations.[13]

simeonschaub:
Closes #88, also `hypot`. More minimal version of #55, we can address `hypot` for `VarArg` and complex input later.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -52,7 +52,10 @@
 @scalar_rule(adjoint(x), Wirtinger(Zero(), One()))
 @scalar_rule(transpose(x), One())
 
-@scalar_rule(abs(x), sign(x))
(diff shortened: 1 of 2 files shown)


oxinabox:
Thanks, will merge when tests pass

simeonschaub:
Why would this only fail on 1.0? Could this be an issue with `FiniteDifferences`?

oxinabox:
We need to set the random number seed at global scope.

PR #103, opened 2019-09-12
introdocs + demo + instructions on writing good rules

oxinabox:
Preview: http://www.juliadiff.org/ChainRules.jl/preview-PR103/

---
Old OP:

This will describe the post https://github.com/JuliaDiff/ChainRulesCore.jl/pull/30
world, so we won't merge this til after that then.

But reading this will help reviewers of that PR and 
https://github.com/JuliaDiff/ChainRules.jl/pull/91

PR #105, opened 2019-09-16
PR preview docs

oxinabox:
This trick is taken from @fredrikekre's https://github.com/fredrikekre/Literate.jl/pull/75

I will rebase #103 ontop of this if we can get this in fairly quick

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -15,7 +15,9 @@ makedocs(
 )
 
 const repo="github.com/JuliaDiff/ChainRules.jl.git"
-if get(ENV, "TRAVIS_PULL_REQUEST", "false") == "false"
+const PR = get(ENV, "TRAVIS_PULL_REQUEST", "false")
+@show PR
+if PR == "false"
     # Normal case, only deply docs if merging to master or release tagged
     deploydocs(repo=repo)
 else
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -14,4 +14,20 @@ makedocs(
     ],
 )
 
-deploydocs(repo="github.com/JuliaDiff/ChainRules.jl.git")
+const repo=="github.com/JuliaDiff/ChainRules.jl.git"
+if get(ENV, "TRAVIS_PULL_REQUEST", "false") == "false"
+    # Normal Case
+    deploydocs(repo=repo)
+else
+    @info "Deploying review docs for PR #$PR"
+    # Overwrite Documenter's function for generating the versions.js file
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -14,4 +14,20 @@ makedocs(
...
 
-deploydocs(repo="github.com/JuliaDiff/ChainRules.jl.git")
+const repo=="github.com/JuliaDiff/ChainRules.jl.git"
+if get(ENV, "TRAVIS_PULL_REQUEST", "false") == "false"
+    # Normal Case
+    deploydocs(repo=repo)
+else
+    @info "Deploying review docs for PR #$PR"
+    # Overwrite Documenter's function for generating the versions.js file
+    foreach(Base.delete_method, methods(Documenter.Writers.HTMLWriter.generate_version_file))
```
nickrobinson251: 🤢

nickrobinson251: Can we link to an issue and make clear that this is evil

nickrobinson251: also can we use `invoke` to get the same behaviour (slightly less evil)

oxinabox: we can not use invoke because we are changing a mether that is called internally by Documentor.deploy docs.

I think real solution is for this to be added as a feature to Documentor,
and then we will just call that.

oxinabox: Link to issue added

fredrikekre: This is basically just a "safer" version of overwriting/defining a more specific method.

docs/make.jl
```diff
@@ -14,4 +14,23 @@ makedocs(
     ],
 )
 
-deploydocs(repo="github.com/JuliaDiff/ChainRules.jl.git")
+const repo=="github.com/JuliaDiff/ChainRules.jl.git"
+if get(ENV, "TRAVIS_PULL_REQUEST", "false") == "false"
+    # Normal Case
```
oxinabox:
```suggestion
    # Normal case, only reply docs if merging to master or release tagged
```

oxinabox:
So this doesn't show up on the list of versions if you are on a different version because it doesn't populate `versions.js` to say it exits
http://www.juliadiff.org/ChainRules.jl/versions.js

I'm not actually sure if that is a problem, since it means  the version selector for docs doesn't get cluttered with old PR doc builds.

Also had to change link in readme from going to latest (i.e. tyhe lasty deployed, which could be a PR) to go the `dev` which i titile master, because @nickrobinson251  found dev confusing

Issue #52, opened 2019-06-13
Make sensitivities for structured matrix arguments structured


ararslan:
Quoting Will in #29:

> FWIW, the other thing to think about is what is actually happening computationally under the hood. Ultimately the `Diagonal` matrix type doesn't use any off-diagonal elements when used in e.g. a matrix-matrix multiply - the `Diagonal` type simply doesn't allow you to have non-zero off-diagonal elements, so it's a slightly odd question to ask what happens if you perturb the off-diagonals by an infinitesimal amount (i.e. compute the gradient w.r.t. them).
> 
> It's this slightly weird situation in which thinking about a `Diagonal` matrix as a regular dense matrix that happens to contain zeros on its off-diagonals isn't really faithful to the semantics of the type (not sure if I've really phrased that correctly, but hopefully the gist is clear)

tkf:
Since (say) `*(x::Diagonal, y::AbstractVector)` in LinearAlgebra is defined in terms of combination of operations whose rules are already defined, would just "nulling" automatically handle the situation?

```julia
frule(::typeof(*), ::Diagonal, ::AbstractVector) = nothing
rrule(::typeof(*), ::Diagonal, ::AbstractVector) = nothing
```

Ref a similar question I posted in Zygote: https://github.com/FluxML/Zygote.jl/issues/316

oxinabox:
I think "disabling" the rules by returnuing nothing would do it,
assuming inside an AD framework.
Better would be to not though and just define the correct answer directly.

tkf:
I think there are too many specializations in LinearAlgebra (and likely in many other in the wild).  But maybe not so bad at least as a "midterm" solution?  Looking at how Zygote handles broadcasting, I have a feeling that it'll take some time to support broadcasting properly...

If ChainRules.jl's stance is "rules specializations are welcome" I can make a PR to implement https://github.com/FluxML/Zygote.jl/issues/316 in ChainRules.jl (after the overhaul #91).

oxinabox:
> I think there are too many specializations in LinearAlgebra (and likely in many other in the wild). But maybe not so bad at least as a "midterm" solution?

yeah, I think bailing out sometimes is the right answer for ChainRules, 
prob with a link to a TODO issue for these cases.

> If ChainRules.jl's stance is "rules specializations are welcome" I can make a PR to implement FluxML/Zygote.jl#316 in ChainRules.jl (after the overhaul #91).

Yeah, that would be great.
[...]

PR #108, opened 2019-09-20
Add rrule for Diagonal * AbstractVector

tkf:
ref https://github.com/JuliaDiff/ChainRules.jl/issues/52#issuecomment-531146406

Reading #103 and other source code, I suppose we are only defining `rrule`s for `Real` element types at the moment so didn't put `adjoint` in the definition. (By the way, if that's the case, I think it makes sense to mention it in #103.)

### DIFF
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -21,9 +21,9 @@ end
 
 function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
     function times_pullback(Ȳ)
-        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D.diag .* Ȳ))
+        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D * Ȳ))
     end
-    return D.diag .* V, times_pullback
+    return D * V, times_pullback
 end
 
 #####
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -19,6 +19,13 @@ function rrule(::typeof(diag), A::AbstractMatrix)
     return diag(A), diag_pullback
 end
 
+function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D.diag .* Ȳ))
+    end
+    return D.diag .* V, times_pullback
+end
+
 #####
 ##### `Symmetric`
 #####
--- a/test/rulesets/LinearAlgebra/structured.jl
+++ b/test/rulesets/LinearAlgebra/structured.jl
@@ -7,6 +7,12 @@
         # Concrete type instead of UnionAll
         rrule_test(typeof(D), D, (randn(rng, N), randn(rng, N)))
     end
+    @testset "::Diagonal * ::AbstractVector" begin
+        rng, N = MersenneTwister(123456), 3
+        rrule_test(*, randn(rng, N),
+                   (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
+                   (randn(rng, N), randn(rng, N)))
+    end
     @testset "diag" begin
         rng, N = MersenneTwister(123456), 7
         rrule_test(diag, randn(rng, N), (randn(rng, N, N), randn(rng, N, N)))
--- a/test/rulesets/LinearAlgebra/structured.jl
+++ b/test/rulesets/LinearAlgebra/structured.jl
@@ -9,9 +9,12 @@
     end
     @testset "::Diagonal * ::AbstractVector" begin
         rng, N = MersenneTwister(123456), 3
-        rrule_test(*, randn(rng, N),
-                   (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
-                   (randn(rng, N), randn(rng, N)))
+        rrule_test(
+            *,
+            randn(rng, N),
+            (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
+            (randn(rng, N), randn(rng, N)),
+        )
     end
     @testset "diag" begin
         rng, N = MersenneTwister(123456), 7


oxinabox:
> I suppose we are only defining rrules for Real element types at the moment so didn't put adjoint in the definition. (By the way, if that's the case, I think it makes sense to mention it in #103.)

it is https://github.com/JuliaDiff/ChainRules.jl/pull/103/files#diff-493d89206c11fea99259d0771034d26aR283

We explictly don't conjugate inside rules.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -19,6 +19,13 @@ function rrule(::typeof(diag), A::AbstractMatrix)
     return diag(A), diag_pullback
 end
 
+function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D.diag .* Ȳ))
```
oxinabox: Do you think it is better to use `parent(D)` rather than accessing a the field directly?
(My benchmark shows they perform identically).
I could go either way.

I would suggest `diag(D)` but that does a copy.
 
```suggestion
        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(parent(D) .* Ȳ))
```

oxinabox:
```suggestion
        return (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D * Ȳ))
```
My benchmarks show that  `D * Ȳ` is just as performant as  `D.diag .* Ȳ`
and I think it is clearer
```
julia> dd*k == dd.diag .* k
true

julia> @show typeof(dd); @show size(dd)
typeof(dd) = Diagonal{Float64,Array{Float64,1}}
size(dd) = (10000, 10000)

julia> @show typeof(k); @show size(k);
typeof(k) = Array{Float64,1}
size(k) = (10000,)


julia> @btime $dd.diag .* $k;
  5.250 μs (2 allocations: 78.20 KiB)

julia> @btime $dd * $k;
  5.257 μs (2 allocations: 78.20 KiB)
```

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -19,6 +19,13 @@ function rrule(::typeof(diag), A::AbstractMatrix)
...
+    end
+    return D.diag .* V, times_pullback
```
oxinabox: Similarly
```suggestion
    return D * V, times_pullback
```
`D.diag .* V` results in a vector

test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -7,6 +7,12 @@
         # Concrete type instead of UnionAll
         rrule_test(typeof(D), D, (randn(rng, N), randn(rng, N)))
     end
+    @testset "::Diagonal * ::AbstractVector" begin
+        rng, N = MersenneTwister(123456), 3
+        rrule_test(*, randn(rng, N),
+                   (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
+                   (randn(rng, N), randn(rng, N)))
```
oxinabox: For consistancy with the rest of the code, the intentation style is:
```
        rrule_test(
            *, randn(rng, N),
            (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
            (randn(rng, N), randn(rng, N)),
        )
```

i.e. always start and end a multiline call with a line that has no arguments on it,
and always ident by 4, rather than indenting to line-up with the brace

nickrobinson251: actually it would be with one argument per line
```julia
rrule_test(
    *,
    randn(rng, N),
    (Diagonal(randn(rng, N)), Diagonal(randn(rng, N))),
    (randn(rng, N), randn(rng, N)),
)
```

oxinabox: Yes, but I feel like this is one of the cases where the clear relationship between 
`*, randn(rng, N),` as inputs to the function makes it worth putting them on one-line.

I'ld be happy with either

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -19,6 +19,13 @@ function rrule(::typeof(diag), A::AbstractMatrix)
     return diag(A), diag_pullback
 end
 
+function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
+    function times_pullback(Ȳ)
```
mattBrzezinski: I'm not very familiar with this package, is there a reason why this sub-function is called `times_pullback()` while others are called `diag_pullback()`?

Also do we need to be nesting functions here? Could this not just be a variable?

```julia
function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
    times_pullback = (NO_FIELDS, @thunk(Diagonal(Ȳ .* V)), @thunk(D.diag .* Ȳ))
    return D.diag * times_pullback
end
```

oxinabox: So it would ideally be called `*_pullback`
but that is not a valid function name.
So we subsitute `times` for the `*`

Note: yo u seem to maybe have misread it, we return the function.
we don't multiply by it.
a pullback is always a function, not a value.


The reason it is defared as a named local comment is for if it ever needs debugging.

Constrast the stack-trace
```
julia> function mkanon()
       x = ()->error()
       return x
       end
mkanon (generic function with 1 method)

julia> mkanon()()
ERROR:
Stacktrace:
 [1] error() at ./error.jl:42
 [2] (::var"##7#8")() at ./REPL[1]:2
 [3] top-level scope at REPL[2]:1
```
vs
```
julia> function mkfoo()
       function foo()
           error()
       end
       return foo
       end
mkfoo (generic function with 1 method)

julia> mkfoo()()
ERROR:
Stacktrace:
 [1] error() at ./error.jl:42
 [2] (::var"#foo#9")() at ./REPL[4]:3
 [3] top-level scope at REPL[5]:1
```

The above is on Julia#master,  but evern the LTS does still render the local function names
just uglier.

tkf: > So it would ideally be called `*_pullback`

We can in principle use the `var` macro/syntax introduced in Julia 1.3 which can be backported easily https://github.com/JuliaLang/julia/pull/32408#issuecomment-506398755:

```julia
julia> VERSION
v"1.0.3"

julia> macro var_str(str)
           Symbol(str)
       end
@var_str (macro with 1 method)

julia> var"*_pullback"() = "hello"
*_pullback (generic function with 1 method)

julia> var"*_pullback"()
"hello"
```

oxinabox: Not worth it.

test/rulesets/LinearAlgebra/structured.jl
tkf: Good to know that ChainRules.jl is using a sane code style!

tkf:
OK. I think I reflected the comments in the new version. Travis is happy with it.

oxinabox:
Thanks

Issue #115, opened 2019-10-04
 Warning: Error requiring NaNMath/SpecialFunctions from ChainRules


awf:
Julia 1.1.0 - is this expected?

```
(v1.1) pkg> add ChainRules.jl
  Updating registry at `C:\Users\awf\.julia\registries\General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
 Installed ChainRulesCore ─ v0.2.0
 Installed FFTW ─────────── v1.0.1
 Installed FillArrays ───── v0.7.2
 Installed ChainRules ───── v0.1.1
 Installed DataStructures ─ v0.17.1
  Updating `C:\Users\awf\.julia\environments\v1.1\Project.toml`
  [082447d4] + ChainRules v0.1.1
  Updating `C:\Users\awf\.julia\environments\v1.1\Manifest.toml`
  [082447d4] + ChainRules v0.1.1
  [d360d2e6] + ChainRulesCore v0.2.0
  [864edb3b] ↑ DataStructures v0.17.0 ⇒ v0.17.1
  [7a1cc6ca] ↑ FFTW v0.3.0 ⇒ v1.0.1
  [1a297f60] ↑ FillArrays v0.7.0 ⇒ v0.7.2
  Building FFTW → `C:\Users\awf\.julia\packages\FFTW\MJ7kl\deps\build.log`

julia> using ChainRules
[ Info: Precompiling ChainRules [082447d4-558c-5d27-93f4-14fc19e9eca2]
┌ Warning: Error requiring NaNMath from ChainRules:
│ LoadError: ArgumentError: Package ChainRules does not have NaNMath in its dependencies:
│ - If you have ChainRules checked out for development and have
│   added NaNMath as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with ChainRules
│ Stacktrace:
│  [1] require(::Module, ::Symbol) at .\loading.jl:836
│  [2] include at .\boot.jl:326 [inlined]
```

oxinabox:
It is harmless, but we really should fix it. 
I think #116 is the fix.

oxinabox:
Fixed in v0.2.1 
Til then do `]add ChainRules#master` 
which will also have other more important changes since 0.2.0 apparently was not correctly released.
https://github.com/JuliaRegistries/General/pull/4115

PR #116, opened 2019-10-04
do using of at-required packages relative

oxinabox:
I believe this should solve #115 

From the [Requires.jl](https://github.com/MikeInnes/Requires.jl/blob/master/README.md) readme:

> In the @require block, or any included files, you can use or import the package, but note that you must use the syntax using .Gadfly or import .Gadfly, rather than the usual syntax. Otherwise you will get a warning about Gadfly not being in dependencies.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.2.0"
+version = "0.2.1"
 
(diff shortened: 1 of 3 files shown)


nickrobinson251:
Have you verified this works (locally)?

oxinabox:
It does not. 
I think I need an extra .

oxinabox:
verified property: it doesn't give that warning anymore.
And also it doesn't just error.

PR #120, opened 2019-10-17
Removes rules depending on casted

oxinabox:
Matches to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/55

There is also a surprisingly large change to `hypot`'s tests
as they were AFAWCT nonsense.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.2.1"
+version = "0.2.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -10,7 +10,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "^0.3"
+ChainRulesCore = "0.3, 0.4"
 FiniteDifferences = "^0.7"
 julia = "^1.0"
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.2.2"
+version = "0.2.2-DEV"
 
 [deps]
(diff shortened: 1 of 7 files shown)


src/rulesets/Base/mapreduce.jl
```diff
@@ -59,7 +59,7 @@ end
 
 function rrule(::typeof(sum), x)
     function sum_pullback(ȳ)
-        return (NO_FIELDS, cast(ȳ))
+        return (NO_FIELDS, @thunk(fill(ȳ, size(x))))
```
nickrobinson251: should we consider using `Fill` from `FillArrays` instead of `fill`?

test/rulesets/Base/base.jl
```diff
@@ -147,22 +147,13 @@
...
-        dx, dy = extern(dxy(cx, cy))
-        @test dx === x / h * cx.value[1]
-        @test dy === y / h * cy.value[2]
+        rng = MersenneTwister(123456)
+        x, Δx, x̄ = randn(rng, 3)
+        y, Δy, ȳ = randn(rng, 3)
+        Δz = randn(rng)
+
+        frule_test(hypot, (x, Δx), (y, Δy))
+        rrule_test(hypot, Δz, (x, x̄), (y, ȳ))
```
nickrobinson251: if there a reason these use `frule_test` but the tests for e.g. `atan2` call `frule` then test the output.

Is it because the `atan2` tests pre-date `frule_test`?

src/rulesets/Base/mapreduce.jl
willtebbutt: That's definitely a thing we're going to do, but at a later date, just to avoid the added complexity for now (we don't want to have to deal with any `FillArray`-specific issues that aren't really `ChainRules` things for the time being)

test/rulesets/Base/base.jl
willtebbutt: > Is it because the atan2 tests pre-date frule_test?

I suspect that this is the case. Will open an issue and resolve at a later date.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.2.1"
+version = "0.2.2"
```
willtebbutt:
```suggestion
version = "0.2.2-DEV"
```

PR #119, opened 2019-10-14
WIP: support SpecialFunctions 0.8 and 0.7

oxinabox:
When done will close #118 
it basically requires solving https://github.com/MikeInnes/Requires.jl/issues/3

Right now this just contains the tooling for how to findout package versions and a single example of using it.
Cross-ref: 

TODO:
 - [ ]Go through and split up deprecations https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/src/deprecated.jl
 - [ ] split-up tests across the versions
 - [ ] decide what to do for as yet unreleased breaking version of SpecialFunctions.jl, (maybe a Warning and then not load any rules?)

Maybe there is a better way.
Ideally from ChainRules perspective would just put the rules into SpecialFunctions.jl
and then they would always be the right version.
However @ararslan as a maintainer of SpecialFunctions.jl makes the reasonable argument that the SpecialFunctions.jl maintainers should not be obliged to take on responsibility for maintaining the derivatives of the SpecialFunctions.
Particularly while ChainRules is still in alpha.

Issue #118, opened 2019-10-14
Fixup SpecialFunctions `lgamma` deprecation


nickrobinson251:
probably means just renaming `lgamma` to just be `loggamma`, but not 100% sure if that will get things write for both Complex and Real values

https://github.com/JuliaDiff/ChainRules.jl/blob/296723ccaeeb4b02c708c9edc039614b2e23e690/src/rulesets/packages/SpecialFunctions.jl#L6

See https://github.com/JuliaMath/SpecialFunctions.jl/pull/156

---

Here's the noisy warning i see currently (when running tests)
```julia
 ┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
│   caller = rrule(::typeof(lgamma), ::Int64) at rule_definition_tools.jl:104
└ @ ChainRules.SpecialFunctionsGlue ~/.julia/packages/ChainRulesCore/ZpveH/src/rule_definition_tools.jl:104
┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
│   caller = #test_scalar#3(::Float64, ::Float64, ::FiniteDifferences.Central{UnitRange{Int64},Array{Float64,1}}, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(test_scalar), ::typeof(lgamma), ::Int64) at test_util.jl:27
└ @ Main ~/projects/autodiff/dev/ChainRules/test/test_util.jl:27
┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
│   caller = fdm(::FiniteDifferences.Central{UnitRange{Int64},Array{Float64,1}}, ::typeof(lgamma), ::Int64, ::Val{true}) at methods.jl:222
└ @ FiniteDifferences ~/.julia/packages/FiniteDifferences/kgtFk/src/methods.jl:222
┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
│   caller = #20 at methods.jl:263 [inlined]
└ @ Core ~/.julia/packages/FiniteDifferences/kgtFk/src/methods.jl:263
┌ Warning: `lgamma(x::Real)` is deprecated, use `(logabsgamma(x))[1]` instead.
[...]

PR #121, opened 2019-10-17
Support SpecialFunction 0.8 + other improvements

oxinabox:
solves #118,
replaces #119 

Along the way we (@willtebbutt and I)  improved the tests ing tools for multiple output functions
so other tests got improved.

Also added the gradient for `sign` because `logabsgamma` does this

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,13 +20,14 @@ println("Testing ChainRules.jl")
 @testset "ChainRules" begin
     include("helper_functions.jl")
     @testset "rulesets" begin
+
         @testset "Base" begin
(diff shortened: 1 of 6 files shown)


test/runtests.jl
```diff
@@ -14,12 +14,15 @@ using ChainRulesCore: extern, accumulate, accumulate!, store!, @scalar_rule,
     Wirtinger, wirtinger_primal, wirtinger_conjugate,
     Zero, One, DNE, Thunk, AbstractDifferential
 
+Random.seed(1) # Set seed that all testsets should reset to.
```
simeonschaub:
```suggestion
Random.seed!(1) # Set seed that all testsets should reset to.
```

PR #133, opened 2019-11-25
[WIP] ChainRules Mathematical Manifesto

oxinabox:
This PR includes my attempt to explain the math behind chainrules.
In particular behind how we define a Differential

[Preview Docs Build](http://www.juliadiff.org/ChainRules.jl/preview-PR133/)

TODO:
- [ ] Explain why simpler case without type-unions does not work (https://github.com/JuliaDiff/ChainRules.jl/pull/133#discussion_r351518481)
- [ ] Explain Zero
- [ ] Scalar Multiplication
- [ ] Structured Multiplication (not really required)
- [ ] Natural Differential Type
- [ ] Extern

### DIFF
--- a/docs/src/math.md
+++ b/docs/src/math.md
@@ -26,6 +26,18 @@ we can consider them seperately.
     - ``d::\mathcal D``: a value ``d`` of type ``\mathcal D``, or the assertion that the value ``d`` has type ``\mathcal D`` </br>
     - ``\mathcal D <: \mathbb D``: the type ``\mathcal D`` is a subtype of the type ``\mathbb D``. In particular, we are only concerned with the case of ``\mathbb D`` being a type-union. So in this case it can be seen as also saying that ``\mathcal D`` is a member of the type-union ``\mathbb D``.
 
+### On Computational Practicalities and Errors.
+In this writeup we will ignore many of the practicalities of computation.
+For example we will treat `Float64` as if they were the real number field.
+We will also ingore basically all errors (except `MethodError`s).
+In particular, we will ignore overflow and underflow, and even `ArgumentError`s.
+If you attempt to add a ``3 \times 3`` matrix to a ``4 \times 4`` matrix,
+then indeed that does not work.
+However, we will not let that prevent that from letting us say that `Matrix` is a differential for `Matrix`.
+Roughly speaking on consider all requirements on functions to he an option to do this or throw an error.
+E.g. that `a + b = c`, to be `a + b = c` or `a + b` throws an error. 
+
+
 ## Part 1: What is a Differential ?
 
 We begin by considering some function we would like to differentiate.
@@ -53,8 +65,7 @@ Consider some type ``\mathcal D``.
 
  - If there exists a type-union ``\mathbb U``, with ``\mathcal D <: \mathbb U``,
  - if for all ``u :: \mathbb U`` and for all ``p :: \mathcal P``, there exists a ``q :: \mathcal P`` such that `u + p = p + u = q`
- - and for all ``d :: \mathcal D``, and for all ``x :: \mathbb U``,  exists ``s :: \mathbb U`` such that ``d + x = x + d = s``
-(!!# TODO should this be on just ``d :: \mathcal D`` or on all of U)
+ - and for all ``d :: \mathcal U``, and for all ``x :: \mathbb U``,  exists ``s :: \mathbb U`` such that ``d + x = x + d = s``
 
 
 then we say that ``\mathcal D`` is a (valid) differential type for ``\mathcal P``.
--- a/docs/src/math.md
+++ b/docs/src/math.md
@@ -11,7 +11,7 @@ The key feature of the the Julia type system is that it enables multiple dispatc
 
 As Julia is a dynamic programming language functions can return different types of values depending on the values of their inputs (i.e. not be type stable).
 
-Bacause of multiple dispatch,
+Because of multiple dispatch,
 all functions are able to be define different _methods_
 depending on the types of the inputs.
 We thus can have defined different ``+``
--- a/docs/src/math.md
+++ b/docs/src/math.md
@@ -84,3 +84,29 @@ Note: in this case it is also true that every other type in the type-union ``\ma
 The informal version of this if a direct-differential type are closed under addition to always give a instance of a direct differential type (potentiall the same one), then it is a valid differential type.
 The important take away is you can add instances of all valid differential types for ``\mathcal P``,
 and know you will always get an instance of valid differential type back.
+
+
+-----
+
+### REWRITE: Differential Type
+
+Consider a Primal Type ``\mathcal P``.
+Consider some type ``\mathcal D``.
+
+
+ - If there exists a type-union ``\mathbb U``, with ``\mathcal D <: \mathbb U``,
+ - if for all ``u :: \mathbb U`` and for all ``p :: \mathcal P``, there exists a ``q :: \mathcal P`` such that `u + p = p + u = q`
+ - and for all ``d :: \mathcal D``, and for all ``x :: \mathbb U``,  exists ``s :: \mathbb U`` such that ``d + x = x + d = s``
+(!!# TODO should this be on just ``d :: \mathcal D`` or on all of U)
+
+
+then we say that ``\mathcal D`` is a (valid) differential type for ``\mathcal P`.
+And we write this as ``\mathcal D \triangleleft \mathcal P``.
+
+Note: in this case it is also true that every other type in the type-union ``\mathbb D``, and and indeed ``\mathbb D`` itself are also valid differential types for ``\mathcal P``.
+
+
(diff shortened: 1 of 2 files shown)


oxinabox:
The definition of valid differential would be much clearer if written in terms of type unions instead of sets of types.

docs/src/math.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+The informal version of this if a direct-differential type are closed under addition to always give a instance of a direct differential type (potentiall the same one), then it is a valid differential type.
+The important take away is you can add instances of all valid differential types for ``\mathcal P``,
+and know you will always get an instance of valid differential type back.
+
+
+-----
+
+### REWRITE: Differential Type
```
willtebbutt: Could we simplify even further by saying something like the following:

A type ``\mathcal D`` is a valid differential type for primal type ``\mathcal P`` if
- addition between a primal in ``\mathcal P`` and differential in ``\mathcal D`` yields another primal in ``\mathcal P``
- addition of two differentials in ``\mathcal D`` produces another differential in ``\mathcal D``. We say that ``\mathcal D`` is closed under addition.

oxinabox: We need the ability to add 2 differentials of different types and come back with another differential, potentially of a 3rd type.
Common example would be `[1.0,2.0,3.0] + Zero() = [1.0,2.0,3.0]`

another isi
`Composite{Foo, NamedTuple{(:x,),Tuple{Float64}})(x=2.0) + Composite{Foo, NamedTuple{(:y,),Tuple{Float64}})(y=3.0) = Composite{Foo, NamedTuple{(:x,:y),Tuple{Float64,Float64}})(x=2.0, y=3.0)`

docs/src/math.md
```diff
@@ -0,0 +1,69 @@
...
+all functions are able to be define different _methods_
+dependencing on the types of the inputs.
+We thus can have defined different ``+``
+for different input types, and so will not distinguish between them.
+The functions thus stand alone from the types,
+except that a set of input type may special case them.
+Where as most similar definitions might decribe a object as a type and some operations on the type,
+we can consider them seperately.
+
+!!! terminology Notation:
```
simeonschaub: This block doesn't seem to render correctly in the PR-preview

docs/src/math.md
```diff
@@ -0,0 +1,69 @@
...
+Consider some type ``\mathcal D``.
+
+
+ - If there exists a type-union ``\mathbb U``, with ``\mathcal D <: \mathbb U``,
+ - if for all ``u :: \mathbb U`` and for all ``p :: \mathcal P``, there exists a ``q :: \mathcal P`` such that `u + p = p + u = q`
+ - and for all ``d :: \mathcal D``, and for all ``x :: \mathbb U``,  exists ``s :: \mathbb U`` such that ``d + x = x + d = s``
+(!!# TODO should this be on just ``d :: \mathcal D`` or on all of U)
+
+
+then we say that ``\mathcal D`` is a (valid) differential type for ``\mathcal P`.
```
simeonschaub:
```suggestion
then we say that ``\mathcal D`` is a (valid) differential type for ``\mathcal P``.
```

docs/src/math.md
nickrobinson251: may need to be 
```
!!! terminology "Notation"
```

oxinabox:
```suggestion
!!! terminology "Notation"
```

docs/src/math.md
```diff
@@ -0,0 +1,69 @@
...
+
+Bacause of multiple dispatch,
+all functions are able to be define different _methods_
+depending on the types of the inputs.
+We thus can have defined different ``+``
+for different input types, and so will not distinguish between them.
+The functions thus stand alone from the types,
+except that a set of input types may special case them.
+Where as most similar definitions might decribe a object as a type and some operations on the type,
+we can consider them seperately.
```
mattBrzezinski: I feel like this reads a bit strange. I understand what you mean, but with multiple dispatch we would be distinguishing between different types. Just the higher level abstraction of `Base.+` would remain the same to the end user.

oxinabox: There is a fun thing in a lot of math write ups that say:

> "A vector space over a field `F` is a set `V` together with two operations `*` and `+` that satisfy the eight axioms"

and treat the `*` and `+` as part of the vector space.
And then for a different vector space, it will be defined with `*'` and `+'` etc,
completely distrinct.

Where as we are saying that we will treat `+` as always a thing that exists.


But yeah this bit does read weirdly

simeonschaub:
Looks very promising! Just a couple of additional thoughts:

1. Should we introduce `Zero` here formally as well? It's mentioned above, but it isn't really explained, how this fits into the framework.
2. Do we want to formalize multiplication of a differential with a scalar as well? Seems that we need this in a lot of cases, too. (Oh, I see this is already on your TODO)
3. I'm wondering, whether we should make use of more conventional mathematical terminology alongside the ChainRules-specific one. What you're describing as a _Differential Type_ here seems to be quite closely related to the concept of vector spaces. (Although floating point arithmetic does not actually describe a field, so things get a bit muddier) I guess, your concept of a _Primal Type_ is also analogous to the concept of a manifold in differential geometry, where your _Differential Types_ would correspond to differential one-forms on a tangent space. I think that would help people coming from a more mathematical background to better understand why we need this extra terminology and where exactly it differs from the more abstract mathematical concepts.

docs/src/math.md
```diff
@@ -0,0 +1,69 @@
...
+    For example booleans, and integers except when they are being uses as special cases of real numbers; as a computational optimization.
+
+### Differential Type
+
+Consider a Primal Type ``\mathcal P``.
+Consider some type ``\mathcal D``.
+
+
+ - If there exists a type-union ``\mathbb U``, with ``\mathcal D <: \mathbb U``,
+ - if for all ``u :: \mathbb U`` and for all ``p :: \mathcal P``, there exists a ``q :: \mathcal P`` such that `u + p = p + u = q`
```
simeonschaub: Technically, `Rational`s wouldn't form a _Primal Type_ according to this definition, since `typemax(Int)//1 + typemax(Int)//1` doesn't correspond to any value `q`, since it throws an error. We might want to loosen this and allow for some edge cases throwing exceptions.

docs/src/math.md
```diff
@@ -0,0 +1,69 @@
...
+We thus can have defined different ``+``
+for different input types, and so will not distinguish between them.
+The functions thus stand alone from the types,
+except that a set of input types may special case them.
+Where as most similar definitions might decribe a object as a type and some operations on the type,
+we can consider them seperately.
+
+!!! terminology "Notation"
+    We use some notation here based closely off of how JuliaLang indicates type relationships. </br>
+    - ``d::\mathcal D``: a value ``d`` of type ``\mathcal D``, or the assertion that the value ``d`` has type ``\mathcal D`` </br>
```
simeonschaub: I don't think the `</br>` are necessary?

oxinabox: Yeah, it just fixes how Atom renders it

docs/src/math.md
oxinabox: Yep, I think a disclaimer at the top about  for purposes of these definitions one can ignore overflow underflow etc

oxinabox:
> Looks very promising! Just a couple of additional thoughts:
> 
>     1. Should we introduce `Zero` here formally as well? It's mentioned above, but it isn't really explained, how this fits into the framework.

Indeed we should explain it here. added to the TODO

>     3. I'm wondering, whether we should make use of more conventional mathematical terminology alongside the ChainRules-specific one.

Yes, we should I just don't know them.


>  (Although floating point arithmetic does not actually describe a field, so things get a bit muddier)

Yeah, but we will ignore that in this write up. 

> I guess, your concept of a _Primal Type_ is also analogous to the concept of a manifold in differential geometry, where your _Differential Types_ would correspond to differential one-forms on a tangent space. 

Yeah, basically.  
The concept of a _Primal Type_ is a _vector_, I guess one might say 1-vector (people don't say that, do they?).
The concept of a _Differential Type_ is indeed a 1-form.  
I thought 1-fotm and differential were synonomous?  
(Maybe one has to ignore orintation/coordinate-ish stuff, which we do do)

oxinabox:
Here is all the math that I want in the PR.
I wrote it on a bus.
I will rewrite it into this PR

https://photos.google.com/share/AF1QipO1fqEhTGO7BwFKGczb8ebCHf0McWpbPrWgO4UVLjjuhYCt4mdwE722MrldJzIyMA?key=MHBFZVBsTWgwRl9hVUlJSUpQQlEzVmhQMzZaQ01B

oxinabox:
Pages 1 and 2, of 9 pages are written

nickrobinson251:
xref https://github.com/JuliaDiff/ChainRulesCore.jl/pull/138

willtebbutt:
@oxinabox what's the status of this PR. Has it been superceded?

oxinabox:
Largely superceded by https://www.juliadiff.org/ChainRulesCore.jl/dev/design/many_differentials.html

oxinabox:
Closing this because it is on the wrong package now

PR #137, opened 2019-12-20
Setup nightly notifications

omus:
Informs the Invenia team of nightly CI failures. @oxinabox setup the daily Travis CI cron job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,12 @@ matrix:
   - julia: 1.3
   - julia: nightly
 notifications:
-  email: false
+  email:
+    recipients:
+      - nightly-rse@invenia.ca
+    on_success: never
+    on_failure: always
+    if: type = cron
 # uncomment the following lines to override the default test script
 #script:
 #  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi


omus:
Related:
- https://github.com/JuliaDiff/ChainRulesCore.jl/pull/72
- https://github.com/JuliaDiff/FiniteDifferences.jl/pull/55

mattBrzezinski:
A nice to have in the future would be to also have a public mailing list, I'm sure there are some non-Invenians that might be interested in keeping up with packages.

oxinabox:
> 

A nice to have in the future would be to also have a public  mailing list, I'm sure there are some non-Invenians that might be  interested in keeping up with packages.

Most people I think do not want to get hit with an email every time the nightly build breaks.
(Though hopefully that is rare.)
And if they do want that, we can just add them to the CI config.

mattBrzezinski:
> Most people I think do not want to get hit with an email every time the nightly build breaks.
> (Though hopefully that is rare.)
> And if they do want that, we can just add them to the CI config recipients list.

That's true, it'd be nice to grow the OSS community and have more people contributing and involved.

oxinabox:
Nightly emails are a thing that generally I would expect people only if it is part of their job.

PR #145, opened 2020-01-10
WIP: Update for ChainRulesCore v0.5

oxinabox:
This PR needs to have ChainRulesCore checked out to master.
I am making sure everything works before tagging the ChainRulesCore v0.5 release.
Since that is a bunch of chages there.

Main thins to check:

- [ ] Wirtinger
- [ ] Casted (I think these are all gone)
- [ ] Composite (I don't think we are using these yet)


oxinabox:
closed by #146

PR #148, opened 2020-01-13
Add broadcast in the `frule` for `*`

YingboMa:
Ref https://github.com/JuliaDiff/ChainRulesCore.jl/pull/93

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -103,7 +103,7 @@
 # product rule requires special care for arguments where `mul` is non-commutative
 
 function frule(::typeof(*), x::Number, y::Number, _, Δx, Δy)
-    return x * y, Δx * y + x * Δy
+    return x * y, @. muladd(Δx, y, x * Δy)
 end
 
 function rrule(::typeof(*), x::Number, y::Number)
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -103,7 +103,11 @@
 # product rule requires special care for arguments where `mul` is non-commutative
 
 function frule(::typeof(*), x::Number, y::Number, _, Δx, Δy)
-    return x * y, @. muladd(Δx, y, x * Δy)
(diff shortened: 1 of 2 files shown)


src/rulesets/Base/base.jl
```diff
@@ -103,7 +103,7 @@
 # product rule requires special care for arguments where `mul` is non-commutative
 
 function frule(::typeof(*), x::Number, y::Number, _, Δx, Δy)
-    return x * y, Δx * y + x * Δy
+    return x * y, @. muladd(Δx, y, x * Δy)
```
oxinabox: I find `@.` surprisingly hard to read.
Took me 10 seconds to realize the `muladd` was getting broadcast also 😂 

So this is the same as:

```suggestion
    return x * y, muladd.(Δx, y, x .* Δy)  # optimized version of `Δx .* y .+ x .* Δy
```

For interest i did some benchmarking:
Writing out the broadcast seems consistently slightly faster.
Is pretty small though.
But I think for the `Number` case which this always is, its just going to hit the 
`muladd(x,y,z) = x*y+z` definition.

```
julia> @btime $dx .* $y .+ $x .* $dy;
  49.495 μs (2 allocations: 781.33 KiB)

julia> @btime muladd.($dx, $y, $x .* $dy);
  49.468 μs (2 allocations: 781.33 KiB)
```

YingboMa: It is potentially more accurate on machines with FMA instructions since there are only two roundings, one in `muladd/fma` the other in `*`.

oxinabox: 🤷‍♂ 

They come out identical on my machine (which has fma) over 100_000 values in `Δy` and `Δx`


I think the accuracy only comes into play if `y` was a matrix.
Which it is not permitted to be here.

YingboMa: Having `muladd` certainly won't hurt. Also, people could define their own number type, and overload `muladd`.

PR #150, opened 2020-01-15
Remove TODO sections in docs

oxinabox:
The information is in other sections,
and even if TODO, belongs in issues, rather than in docs.

It also just looks bad


oxinabox:
yeah maybe

mattBrzezinski:
I would be in favour of moving the two lines into [index.md](https://github.com/JuliaDiff/ChainRules.jl/blob/master/docs/src/index.md)

oxinabox:
I have moved the info to the FAQ

PR #152, opened 2020-01-17
Doc updates for ChainRulesCore 0.5

oxinabox:
New docs,
major change is about  the pushforward now fused into the `frule`.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -42,7 +42,7 @@ computing `foo(x)` is doing the _primal_ computation.
     Their exact functioning is fairly ChainRules specific, though other tools have similar functions.
     The core notion is sometimes called _custom AD primitives_, _custom adjoints_, _custom_gradients_, _custom sensitivities_.
 
-The rules are encoded as `rrule`s and `frule`s, for use in forward-mode and reverse-mode differentiation respectively.
+The rules are encoded as `frule`s and `rrule`s, for use in forward-mode and reverse-mode differentiation respectively.
 
 The `rrule` for some function `foo`, which takes the positional arguments `args` and keyword arguments `kwargs`, is written:
 
@@ -52,7 +52,7 @@ function rrule(::typeof(foo), args...; kwargs...)
     return y, pullback
 end
 ```
-Where `y` (the primal result) must be equal to `foo(args...; kwargs...)`.
+where `y` (the primal result) must be equal to `foo(args...; kwargs...)`.
 `pullback` is a function to propagate the derivative information backwards at that point.
 That pullback function is used like:
 `∂self, ∂args... = pullback(Δy)`
@@ -238,7 +238,7 @@ Similarly every `pullback` returns an extra `∂self`, which for things without 
 
 The most trivial use of the `pushforward` from within `frule` is to calculate the directional derivative:
 
-If we would like to know the the directional derivative of `f` for an input chage of `(1.5, 0.4, -1)`
+If we would like to know the the directional derivative of `f` for an input change of `(1.5, 0.4, -1)`
 
 ```julia
 direction = (1.5, 0.4, -1) # (ȧ, ḃ, ċ)
@@ -279,7 +279,7 @@ The most important `AbstractDifferential`s when getting started are the ones abo
 
 #### Other `AbstractDifferential`s:
  - `Composite{P}`: this is the differential for tuples and  structs. Use it like a `Tuple` or `NamedTuple`. The type parameter `P` is for the primal type.
- - `DoesNotExist`: Zero-like, represents that the operation on this input is not differentiable. Its primal type is normally integer or boolean.
+ - `DoesNotExist`: Zero-like, represents that the operation on this input is not differentiable. Its primal type is normally `Integer` or `Bool`.
  - `InplaceableThunk`: it is like a Thunk but it can do `store!` and `accumulate!` in-place.
 
  -------------------------------
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -56,7 +56,6 @@ where `y` (the primal result) must be equal to `foo(args...; kwargs...)`.
 `pullback` is a function to propagate the derivative information backwards at that point.
 That pullback function is used like:
 `∂self, ∂args... = pullback(Δy)`
-function takes arguments;:  (more later).
 
 
 Almost always the _pullback_ will be declared locally within the `rrule`, and will be a _closure_ over some of the other arguments, and potentially over the primal result too.
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -438,7 +438,7 @@ It is very easy to check gradients or derivatives with a computer algebra system
 
 #### `Δx`, `∂x`, `dx`
 ChainRules uses these perhaps atyptically.
-As a notation that is the same across propagators, regardless of direction. (Incontrast see `ẋ` and `x̄` below)
+As a notation that is the same across propagators, regardless of direction (incontrast see `ẋ` and `x̄` below).
 
  - `Δx` is the input to a propagator, (i.e a _seed_ for a _pullback_; or a _perturbation_ for a _pushforward_)
  - `∂x` is the output of a propagator
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -161,7 +161,7 @@ If the function is `y = f(x)` often the pullback will be written `s̄elf, x̄ = 
 
     The pullback returns one `∂arg` per `arg` to the original function, plus one `∂self` for the fields of the function itself (explained below).
 
-!!! terminology
+!!! terminology "perturbation, seed, sensitivity"
     Sometimes _perturbation_, _seed_, and even _sensitivity_ will be used interchangeably.
     They are not generally synonymous, and ChainRules shouldn't mix them up.
     One must be careful when reading literature.
@@ -200,11 +200,10 @@ If the function is `y = f(x)` often the pushforward will be written `ẏ = last(
 
 ### Self derivative `Δself`, `∂self`, `s̄elf`, `ṡelf` etc.
 
-!!! terminology  "`Δself`, `∂self`, `s̄elf`, `ṡelf`"
+!!! terminology "Δself, ∂self, s̄elf, ṡelf"
     It is the derivatives with respect to the internal fields of the function.
     To the best of our knowledge there is no standard terminology for this.
     Other good names might be `Δinternal`/`∂internal`.
-
 
 From the mathematical perspective, one may have been wondering what all this `Δself`, `∂self` is.
 Given that a function with two inputs, say `f(a, b)`, only has two partial derivatives:
@@ -218,6 +217,7 @@ For example a closure has the fields it closes over; a callable object (i.e. a f
 So every `pushforward` takes in an extra argument, which is ignored unless the original function has fields.
 It is common to write `function foo_pushforward(_, Δargs...)` in the case when `foo` does not have fields.
 Similarly every `pullback` returns an extra `∂self`, which for things without fields is the constant `NO_FIELDS`, indicating there are no fields within the function itself.
+
 
 #### Pushforward / Pullback summary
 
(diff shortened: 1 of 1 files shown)


nickrobinson251:
view at http://www.juliadiff.org/ChainRules.jl/preview-PR152/

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
+    - 1 argument per original function argument + 1 for the function itself
+    - 1 return per original function return
+
+
+#### Pullback/Pushforward and Directional Derivative/Gradient
+
+The most trivial use of the `pushforward` from within `frule` is to calculate the directional derivative:
 
-The most trivial use of `frule` and returned `pushforward` is to calculate the [Total Derivative](https://en.wikipedia.org/wiki/Total_derivative):
+If we would like to know the the directional derivative of `f` for an input chage of `(1.5, 0.4, -1)`
```
nickrobinson251:
```suggestion
If we would like to know the the directional derivative of `f` for an input change of `(1.5, 0.4, -1)`
```

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
+direction = (1.5, 0.4, -1) # (ȧ, ḃ, ċ)
+y, ẏ = frule(f, a, b, c, Zero(), 1, 0, 0)
 ```
 
-Then we have that `ẏ` is the _total derivative_ of `f` at `(a, b, c)`, written mathematically as ``df_{(a,b,c)}``
+On the basis directions one gets the partial derivatives of `y`:
+```julia
+y, dy_da = frule(f, a, b, c, Zero(), 1, 0, 0)
+y, dy_db = frule(f, a, b, c, Zero(), 0, 1, 0)
+y, dy_dc = frule(f, a, b, c, Zero(), 0, 0, 1)
```
nickrobinson251: `dy_da` could also be written `ā`, and `dy_db` could be written `b̄` and so on?

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
 or it might be one of the `AbstractDifferential` subtypes.
 
 Differentials support a number of operations.
 Most importantly: `+` and `*`, which let them act as mathematical objects.
-And `extern` which converts `AbstractDifferential` types into a conventional non-ChainRules type.
 
 The most important `AbstractDifferential`s when getting started are the ones about avoiding work:
 
- - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `extern` is called on the `Thunk`. More on thunks later.
+ - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `unthunk` is called on the `Thunk`. (`unthunk` is a no-op on nonthunked inputs.)
```
nickrobinson251:
```suggestion
 - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `unthunk` is called on the `Thunk`. `unthunk` is a no-op on inputs that are not `Thunk`s.
```

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
  - `One`, `Zero`: There are special representations of `1` and `0`. They do great things around avoiding expanding `Thunks` in multiplication and (for `Zero`) addition.
 
-#### Other `AbstractDifferential`s: don't worry about them right now
- - `Casted`: it implements broadcasting mechanics. See [#10](https://github.com/JuliaDiff/ChainRulesCore.jl/issues/10)
+#### Other `AbstractDifferential`s:
+ - `Composite{P}`: this is the differential for tuples and  structs. Use it like a tuple or named tuple. The type parameter `P` is for the primal type.
```
nickrobinson251:
```suggestion
 - `Composite{P}`: this is the differential for tuples and  structs. Use it like a `Tuple` or `NamedTuple`. The type parameter `P` is for the primal type.
```

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
+ - `DoesNotExist`: Zero-like, represents that the operation on this input is not differentiable. Its primal type is normally integer or boolean.
```
nickrobinson251:
```suggestion
 - `DoesNotExist`: Zero-like, represents that the operation on this input is not differentiable. Its primal type is normally `Integer` or `Bool`.
```

docs/src/index.md
```diff
@@ -397,7 +438,7 @@ It is very easy to check gradients or derivatives with a computer algebra system
 
 #### `Δx`, `∂x`, `dx`
 ChainRules uses these perhaps atyptically.
-As a notation that is the same across propagators, regardless of direction. (Incontrast see `ẋ` and `x̄` below)
+As a notation that is the same across propagators, regardless of direction. (Incontrast see `ẋ` and `x̄` below)
```
nickrobinson251:
```suggestion
As a notation that is the same across propagators, regardless of direction (incontrast see `ẋ` and `x̄` below).
```

docs/src/index.md
```diff
@@ -18,49 +18,77 @@ Knowing rules for more complicated functions speeds up the autodiff process as i
...
+    return y, pullback
 end
 ```
-where `y = foo(args; kwargs...)`, and `pushforward` is a function to propagate the derivative information forwards at that point (more later).
-
+Where `y` (the primal result) must be equal to `foo(args; kwargs...)`
+`pullback` is a function to propagate the derivative information backwards at that point.
+That pullback function is used:
+`∂self, ∂args... = pullback(Δy)`
+function takes arguments;:  (more later).
```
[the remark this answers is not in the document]
oxinabox: I think I got distracted when writing it, it probably can be deleted.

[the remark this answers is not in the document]
nickrobinson251:
```suggestion
```

docs/src/index.md
```diff
@@ -114,55 +142,65 @@ pushforward to find ``\dfrac{∂f}{∂x}``:
...
+    In particular if that function returned a tuple then `∂y` will be a tuple of the same size.
 
-### Self derivative `Δself`, `∂self`, `s̄elf`, `ṡelf` etc.
+### Self derivative `Δself`, `∂self`, `s̄elf`, `ṡelf` etc.
 
-!!! terminology  `Δself`, `∂self`, `s̄elf`, `ṡelf`
+!!! terminology  "`Δself`, `∂self`, `s̄elf`, `ṡelf`"
     It is the derivatives with respect to the internal fields of the function.
     To the best of our knowledge there is no standard terminology for this.
     Other good names might be `Δinternal`/`∂internal`.
```
nickrobinson251: hmmm, this still doesn't seem to render right... may something like
```julia
!!! terminology "self"
    `Δself`, `∂self`, `s̄elf`, `ṡelf` denote the derivatives with respect to the internal fields of the function.
    To the best of our knowledge there is no standard terminology for this.
    Other good names might be `Δinternal`/`∂internal`.
```

docs/src/index.md
oxinabox: Not quite since both `Thunk` and `InplaceThunk` are thunks

docs/src/index.md
oxinabox: I think just dropping the backticks is best

docs/src/index.md
nickrobinson251: Ah I forgot Thunk isn't the abstract type. 

```suggestion
 - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `unthunk` is called on the `Thunk`. `unthunk` is a no-op on non-thunked inputs.
```

docs/src/index.md
```diff
@@ -18,49 +18,77 @@ Knowing rules for more complicated functions speeds up the autodiff process as i
...
+
 ### `frule` and `rrule`
 
 !!! terminology "`frule` and `rrule`"
     `frule` and `rrule` are ChainRules specific terms.
     Their exact functioning is fairly ChainRules specific, though other tools have similar functions.
     The core notion is sometimes called _custom AD primitives_, _custom adjoints_, _custom_gradients_, _custom sensitivities_.
 
-The rules are encoded as `frule`s and `rrule`s, for use in forward-mode and reverse-mode differentiation respectively.
+The rules are encoded as `rrule`s and `frule`s, for use in forward-mode and reverse-mode differentiation respectively.
```
willtebbutt:
```suggestion
The rules are encoded as `frule`s and `rrule`s, for use in forward-mode and reverse-mode differentiation respectively.
```

docs/src/index.md
```diff
@@ -18,49 +18,77 @@ Knowing rules for more complicated functions speeds up the autodiff process as i
...
 ```julia
-function frule(::typeof(foo), args; kwargs...)
+function rrule(::typeof(foo), args...; kwargs...)
     ...
-    return y, pushforward
+    return y, pullback
 end
 ```
-where `y = foo(args; kwargs...)`, and `pushforward` is a function to propagate the derivative information forwards at that point (more later).
+Where `y` (the primal result) must be equal to `foo(args...; kwargs...)`.
```
willtebbutt:
```suggestion
where `y` (the primal result) must be equal to `foo(args...; kwargs...)`.
```

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
+The most trivial use of the `pushforward` from within `frule` is to calculate the directional derivative:
 
-The most trivial use of `frule` and returned `pushforward` is to calculate the [Total Derivative](https://en.wikipedia.org/wiki/Total_derivative):
+If we would like to know the the directional derivative of `f` for an input chage of `(1.5, 0.4, -1)`
 
 ```julia
-y, f_pushforward = frule(f, a, b, c)
-ẏ = f_pushforward(1, 1, 1, 1)  # for appropriate `1`-like perturbation.
+direction = (1.5, 0.4, -1) # (ȧ, ḃ, ċ)
+y, ẏ = frule(f, a, b, c, Zero(), 1, 0, 0)
```
willtebbutt: Should the derivative arguments here not be `direction`?

docs/src/index.md
willtebbutt: The overbar notation is usually reserved for reverse-mode-related quantities, so probably best not to do that here. Might make sense to write these as partials though (to be pedantic):
```julia
∂y_∂a
```
etc

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
 ```julia
 y, f_pullback = rrule(f, a, b, c)
 ∇f = f_pullback(1)  # for appropriate `1`-like seed.
-s̄elf, ā, b̄, c̄ = ∇f
+s̄elf, ā, b̄, c̄ = ∇f
 ```
 Then we have that `∇f` is the _gradient_ of `f` at `(a, b, c)`.
 And we thus have the partial derivatives ``\overline{\mathrm{self}}, = \dfrac{∂f}{∂\mathrm{self}}``, ``\overline{a} = \dfrac{∂f}{∂a}``, ``\overline{b} = \dfrac{∂f}{∂b}``, ``\overline{c} = \dfrac{∂f}{∂c}``, including the and the self-partial derivative, ``\overline{\mathrm{self}}``.
 
 ### Differentials
```
willtebbutt: I think this whole section needs to be expanded on to make various points:

- `+` is the only operation guaranteed to be defined on differentials and, moreover, it's only guaranteed to be defined between differentials that are valid for the same primal type.
- `*` is only guaranteed to be defined between scalars and differentials. You can't generally multiple a differential by another differential
- `+` between primals and differentials isn't always defined, but it is for a lot of interesting cases e.g. `Real`, `Matrix{<:Real}` etc
- stuff that I've missed that we've figured out since this section was last written.

docs/src/index.md
```diff
@@ -183,57 +221,65 @@ Similarly every `pullback` returns an extra `∂self`, which for things without
...
 
 Differentials support a number of operations.
 Most importantly: `+` and `*`, which let them act as mathematical objects.
-And `extern` which converts `AbstractDifferential` types into a conventional non-ChainRules type.
 
 The most important `AbstractDifferential`s when getting started are the ones about avoiding work:
 
- - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `extern` is called on the `Thunk`. More on thunks later.
+ - `Thunk`: this is a deferred computation. A thunk is a [word for a zero argument closure](https://en.wikipedia.org/wiki/Thunk). A computation wrapped in a `@thunk` doesn't get evaluated until `unthunk` is called on the `Thunk`. (`unthunk` is a no-op on nonthunked inputs.)
  - `One`, `Zero`: There are special representations of `1` and `0`. They do great things around avoiding expanding `Thunks` in multiplication and (for `Zero`) addition.
```
willtebbutt:
```suggestion
 - `Zero`: There is a special representation of `0`. It's defined as the differential that leaves and differential that it's added to unchanged. It's great for avoiding computation.
```
I just don't really want to talk about `One` in the docs since it's looking increasingly likely that we'll drop it in the near future.

oxinabox:
> Also, how would you feel about changing all references to autodiff / automatic differentiation to AD / algorithmic differentiation?
In the "On writing good rrule and frule " bit, we should consider modifying the recommendations for Zero() or One(). In particular, removing any mention of One because it's not generally appropriate or helpful. Additionally, the "Write Tests" section should probably be entitled "Write Tests using FiniteDifferences". Similarly, the "CAS Systems are your friends" should have some reference to them not being helpful when writing tests for derivatives.

These are valid points, but beyond he scope of this PR. Open an issue?

docs/src/index.md
oxinabox: Yep. But out of scope for this PR

docs/src/index.md
nickrobinson251: i don't feel strongly, but i'd slightly leave change to `One` in docs to be included in (or after) any changes to `One` itself

willtebbutt: Yeah, you're probably right. I'm being overzealous.

docs/src/index.md
oxinabox: They should. Good catch

PR #156, opened 2020-01-20
Add difference between `Zero` and `DoesNotExist` to FAQ

oxinabox:


docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
 or to be functionality changing to perform a different operation, e.g. `dims=3`, and thus not differentiable.
 To the best of our knowledge no Julia AD system, with support for the definition of custom primitives, supports differentiating with respect to keyword arguments.
 At some point in the future ChainRules may support these. Maybe.
+
+
+### What is the difference between `Zero` and `DoesNotExist` ?
+`Zero` and `DoesNotExist` act in practice -- they result in not change whenever added to anything.
```
nickrobinson251:
```suggestion
`Zero` and `DoesNotExist` act almost exactly the same in practice: they result in no change whenever added to anything.
```

docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
...
+Odds are if you write a rule that returns the wrong one everything will just work fine. 
+We provide both to allow for more clear writing of rules, and easier debuggability.
```
nickrobinson251:
```suggestion
We provide both to allow for clearer writing of rules, and easier debugging.
```

docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
...
+
+`Zero()` represents the fact that if one perturbes (adds a small change to) the matching primal there will be no change in the behavour of the primal function.
```
nickrobinson251:
```suggestion
`Zero()` represents the fact that if one perturbs (adds a small change to) the matching primal there will be no change in the behavour of the primal function.
```

docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
...
+For example in `fst(x,y) = x`, then the derivative of `fst` with respect to `y` is `Zero()`.
+`fst(10, 5) == 10` and if we add `0,1` to `5` we still get `fst(10, 5.1)=10`.
+
+`DoesNotExist()` represents the favt that if one perturbes the matching primal, the primal function will now error.
```
nickrobinson251:
```suggestion
`DoesNotExist()` represents the fact that if one perturbs the matching primal, the primal function will now error.
```

docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
...
+For example in `nth(xs, n) = xs[n]` then the derivative of `nth` with respect to `n` is `DoesNotExist()`.
```
nickrobinson251: let's not use `nth` because "the derivative of `nth`" and "the `nth` derivative" and "the nth derivative" are too easy to confuse; and it's easy to make a different example

e.g.
```suggestion
For example in `access(xs, n) = xs[n]` then the derivative of `access` with respect to `n` is `DoesNotExist()`.
```

docs/src/index.md
```diff
@@ -430,3 +430,17 @@ As a rule keyword arguments tend to control side-effects, like logging verbosity
...
+`nth([10, 20, 30], 2) = 20`, but if we add `0.1` to `2` we get `nth([10, 20, 30], 2.1)` which errors as indexing can't be applied at fractional indexes.
```
nickrobinson251:
```suggestion
`access([10, 20, 30], 2) = 20`, but if we add `0.1` to `2` we get `access([10, 20, 30], 2.1)` which errors as indexing can't be applied at fractional indexes.
```

oxinabox: what is wrong with `nth` ? F# calls the function this IIRC.
Its like `first`

oxinabox: Oh you replied above. legit

PR #158, opened 2020-01-21
Break up giant single page of docs

oxinabox:
I had been meaning to break the FAQ and Writing good Rules sections out into their own pages for ages.

Could potentially break out a bit more, but I think this is an improvement and it  gets harder if we want to think about moving out e.g. the direct usage example or part of the descriptions of the propagators.

Only change other than moving things is to the levels of headings.
To match the next structure.

### DIFF
--- a/docs/src/FAQ.md
+++ b/docs/src/FAQ.md
@@ -1,0 +1,69 @@
+# FAQ
+
+## What is up with the different symbols?
+
+### `Δx`, `∂x`, `dx`
+ChainRules uses these perhaps atyptically.
+As a notation that is the same across propagators, regardless of direction (incontrast see `ẋ` and `x̄` below).
+
+ - `Δx` is the input to a propagator, (i.e a _seed_ for a _pullback_; or a _perturbation_ for a _pushforward_)
+ - `∂x` is the output of a propagator
+ - `dx` could be either
+
+
+### dots and bars: ``\dot{y} = \dfrac{∂y}{∂x} = \overline{x}``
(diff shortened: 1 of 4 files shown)


docs/src/FAQ.md
```diff
@@ -0,0 +1,69 @@
...
+For example in `fst(x,y) = x`, then the derivative of `fst` with respect to `y` is `Zero()`.
+`fst(10, 5) == 10` and if we add `0,1` to `5` we still get `fst(10, 5.1)=10`.
+
+`DoesNotExist()` represents the fact that if one perturbs the matching primal, the primal function will now error.
+For example in `access(xs, n) = xs[n]` then the derivative of `access` with respect to `n` is `DoesNotExist()`.
+`access([10, 20, 30], 2) = 20`, but if we add `0.1` to `2` we get `access([10, 20, 30], 2.1)` which errors as indexing can't be applied at fractional indexes.
+
+
+## When to use ChainRules vs ChainRulesCore?
+
```
nickrobinson251:
```suggestion
```

docs/src/writing_good_rules.md
```diff
@@ -0,0 +1,79 @@
+# On writing good `rrule` / `frule` methods
+
+## Use `Zero()` or `One()` as return value
+
+The `Zero()` and `One()` differential objects exist as an alternative to directly returning
+`0` or `zeros(n)`, and `1` or `I`.
+They allow more optimal computation when chaining pullbacks/pushforwards, to avoid work.
+They should be used where possible.
+
+## Use `Thunk`s appropriately:
```
nickrobinson251:
```suggestion
## Use `Thunk`s appropriately
```

PR #165, opened 2020-01-26
Switch to Blue Style

### LINKED ISSUE #141: Change to BlueStyle
Talked to @jrevels about this a few months back. And he wasfine with it.
Right now the code is roughly in the intersection of what YASGuide and BlueStyle allow.
Which is pretty large as they overlap significantly.

But given most people maintaining it now are used to BlueStyle, we should use that.

#80 doesn't follow BlueStyle indenting.
And that's fine because the readme says YASGuide.
But we should change that

nickrobinson251:
- closes #141 
- We mostly follow this style anyway :)
- Alphabetises some `scalar_rules` (although not required by the guide, seemed a good opportunity)
- Rewrites some long lines (with some pleasing improvements in clarity in my opinion)


Project.toml
```diff
@@ -10,7 +10,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.6"
```
nickrobinson251: This allows us to remove the line `using ChainRulesCore: AbstractZero`

src/rulesets/Base/base.jl
```diff
@@ -2,62 +2,64 @@
 @scalar_rule(zero(x), Zero())
 @scalar_rule(sign(x), Zero())
 
+@scalar_rule(abs(x::Real), sign(x))
 @scalar_rule(abs2(x), 2x)
+@scalar_rule(exp(x), Ω)
```
nickrobinson251: Not required by Blue Style, but I alphabetised these blocks of code, to make the rules easier to find

src/rulesets/LinearAlgebra/blas.jl
```diff
@@ -129,8 +124,9 @@ function rrule(::typeof(gemv), tA::Char, α::T, A::AbstractMatrix{T},
     return y, gemv_pullback
 end
 
-function rrule(::typeof(gemv), tA::Char, A::AbstractMatrix{T},
-               x::AbstractVector{T}) where T<:BlasFloat
+function rrule(
+    ::typeof(gemv), tA::Char, A::AbstractMatrix{T}, x::AbstractVector{T}
+) where T<:BlasFloat
```
nickrobinson251: This is one of the few actual style changes -- it now matches the style we use everywhere else

Issue #149, opened 2020-01-13
Tests broken on Julia 1.3


YingboMa:
https://travis-ci.org/JuliaDiff/ChainRules.jl/jobs/636506595?utm_medium=notification&utm_source=github_status

I cannot reproduce it locally. @oxinabox do you mind to take a look?

oxinabox:
They also pass locally for me.

oxinabox:
I have restarted the job, it could be some weird randomness crept in around the `seed` 
and managed to get it to break on finite differencing.
If so that is an issue, we should have it seeded everywhere

nickrobinson251:
seem to be passing (and we run CI every night)

oxinabox:
Have you checked?
If so rena

PR #169, opened 2020-02-19
Update to new API

willtebbutt:
See [ChainRulesCore](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/129) for related PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ julia:
   - nightly
 jobs:
   allow_failures:
-    - 1.3
+    - julia: 1.3
     - julia: nightly
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ julia:
   - nightly
 jobs:
   allow_failures:
-    - julia: 1.3
     - julia: nightly
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ julia:
   - nightly
 jobs:
   allow_failures:
+    - 1.3
     - julia: nightly
 
 after_success:
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.7.0-DEV"
+ChainRulesCore = "0.7.0"
 ChainRulesTestUtils = "0.2"
 FiniteDifferences = "0.7"
 Reexport = "0.2"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.3.4"
+version = "0.4.0"
 
 [deps]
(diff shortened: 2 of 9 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.3.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
```
nickrobinson251: Should be test-only dep?

Project.toml
```diff
@@ -12,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "0.6.1"
```
nickrobinson251: Need to change this to `0.7.0`

Project.toml
```diff
@@ -12,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
...
 ChainRulesTestUtils = "0.1.3"
-FiniteDifferences = "^0.7"
+FiniteDifferences = "0.7"
 Reexport = "0.2"
 Requires = "0.5.2, 1"
 julia = "^1.0"
```
nickrobinson251: since we're doing this for FiniteDifferences

```suggestion
julia = "1"
```

Project.toml
nickrobinson251:
```suggestion
```

Project.toml
```diff
@@ -12,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "0.6.1"
 ChainRulesTestUtils = "0.1.3"
```
nickrobinson251: Needs bumping to match https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/17

```suggestion
ChainRulesTestUtils = "0.2"
```

Project.toml
willtebbutt:
```suggestion
ChainRulesCore = "0.7.0-DEV"
```

willtebbutt:
@nickrobinson251 let me know if you're happy for this to be merged now.

.travis.yml
```diff
@@ -10,7 +10,6 @@ julia:
   - nightly
```
oxinabox:
```suggestion
  - 1.4
```

willtebbutt:
> It is worrying how few tests needed to be updated

Yeah. This is what caused this slip up with the API in the first place -- we simply didn't have the range of things implemented to catch the problem. I think that the right place to resolve this issue is the core though -- I've implemented extra tests to prevent API regression, but presumably there are other things we're not handling properly (kwargs, for example).

willtebbutt:
@oxinabox @nickrobinson251 anyone able to reproduce the errors locally? Tests are passing for me on 1.3.1. Have checked versions, and I appear to be using the same as CI is using...

oxinabox:
Oh is this: https://github.com/JuliaDiff/ChainRules.jl/issues/149 ?

I think we had travis disabled on 1.3 for reasons because it was just broken even though it worked fine in 1.3 on everyones local machines

willtebbutt:
Ahh I see. Will re-disable and merge once tests pass.

PR #173, opened 2020-04-17
Added support for ldexp and clamp

freemin7:
I really tried to find a solution for testing ldexp, however i fear finding a solution will not be possible without changing ChainRulesTestUtils.jl.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -23,6 +23,7 @@
 @scalar_rule deg2rad(x) π / oftype(x, 180)
 @scalar_rule rad2deg(x) oftype(x, 180) / π
 
+@scalar_rule(ldexp(x, y), (2^y, DoesNotExist()))
 
 # Can't multiply though sqrt in acosh because of negative complex case for x
 @scalar_rule acosh(x) inv(sqrt(x - 1) * sqrt(x + 1))
@@ -65,6 +66,14 @@
 @scalar_rule sinpi(x) π * cospi(x)
 @scalar_rule tand(x) (π / oftype(x, 180)) * (1 + Ω ^ 2)
 
+@scalar_rule(
+    clamp(x, low, high),
+    @setup(
+        islow = x < low,
+        ishigh = high < x,
+    ),
+    (!(islow | ishigh), islow, ishigh),
+)
 @scalar_rule x \ y (-((y / x) / x), inv(x))
 
 function frule((_, ẏ), ::typeof(identity), x)
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -92,6 +92,18 @@
         @test extern(dy) == extern(zeros(2, 5) .+ dy)
     end
 
+     @testset "ldexp" begin
+            x, Δx, x̄ = 10rand(3)
+            Δz = rand()
+
+            for n in (0,1,20)
+                # TODO: Forward test does not work when parameter is Integer
+                # See: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/22
+                #frule_test(ldexp, (x, Δx), (n, nothing))
+                rrule_test(ldexp, Δz, (x, x̄), (n, nothing))
+            end
+     end
+
     @testset "binary function ($f)" for f in (mod, \)
         x, Δx, x̄ = 10rand(3)
         y, Δy, ȳ = rand(3)
@@ -134,5 +146,22 @@
 
         frule_test(f, (x, Δx), (y, Δy), (z, Δz))
         rrule_test(f, Δk, (x, x̄), (y, ȳ), (z, z̄))
+    end
+    @testset "clamp"  begin
+        x̄, ȳ, z̄    = randn(3)
+        Δx, Δy, Δz = randn(3)
+        Δk = randn()
+
+        x, y, z = 1., 2., 3.  # to left
+        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
+        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))
+
+        x, y, z = 2.5, 2., 3.  # in the middle
+        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
+        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))
+
+        x, y, z = 4., 2., 3.  # to right
+        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
+        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))
     end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.6.0"
+version = "0.6.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


freemin7:
I decided to go with this hack:
```
     @testset "ldexp" begin
         @warn "ldexp not tested, since functions with Integer arguments have no ∂f/∂n"
         ## I tried to find a way test it,
         ## however finite diff produces Floats which ldexp can't handle.

     end
```

src/rulesets/Base/base.jl
```diff
@@ -91,6 +92,12 @@
 @scalar_rule(fma(x, y, z), (y, x, One()))
 @scalar_rule(max(x, y), @setup(gt = x > y), (gt, !gt))
 @scalar_rule(min(x, y), @setup(gt = x > y), (!gt, gt))
+@scalar_rule(clamp(x, low, high),
+@setup(
+islow = x < low,
+ishigh = high < x),
+(!( islow | ishigh), islow, ishigh)
+)
```
simeonschaub, line 100: We generally try to follow https://github.com/invenia/BlueStyle here, so I would format this more like this:
```suggestion
@scalar_rule(
    clamp(x, low, high),
    @setup(
        islow = x < low
        ishigh = high < x
    ),
    (!(islow | ishigh), islow, ishigh)
)
```
I think, we should also use `Zero()` and `One()` here instead of `Bool`s, which you can write like `islow ? One() : Zero()`. This has the benefit, that it allows us to avoid allocations when multiplying large arrays, for example.

test/rulesets/Base/base.jl
```diff
@@ -115,6 +115,13 @@
         @test extern(dy) == extern(zeros(2, 5) .+ dy)
     end
 
+     @testset "ldexp" begin
+         @warn "ldexp not tested, since functions with Integer arguments have no ∂f/∂n"
+         ## I tried to find a way test it,
```
simeonschaub, line 120: You could test only the first argument by calling `ChainRulesTestUtils._make_fdm_call` directly, but it's not really well documented right now, so if you want, I can also add that later. Generally, it's not really useful to throw warnings while the tests run, because this easily gets lost, so I would probably put something `@test_broken ldexp(ldexp, 1, 2)` here instead.

simeonschaub:
Thanks for the PR! I left some suggestions, but this helps a lot already!

test/rulesets/Base/base.jl
freemin7: How would i make that change now that i made a pull request?
I never used Git in multiplayer before

simeonschaub: You should probably pull the changes made here to your local clone first with `git pull origin clamp+ldexp`. You can then just make the changes there, commit them and push to this branch of your fork again.

freemin7:
The forward pass does not seem to work:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/22#issuecomment-615961039

If a fix for this materializes it should be safe to uncoment the forward pass. If this makes no sense i will remove the line.

src/rulesets/Base/base.jl
```diff
@@ -91,6 +92,12 @@
 @scalar_rule(fma(x, y, z), (y, x, One()))
 @scalar_rule(max(x, y), @setup(gt = x > y), (gt, !gt))
 @scalar_rule(min(x, y), @setup(gt = x > y), (!gt, gt))
+@scalar_rule(clamp(x, low, high),
+    @setup(
+        islow = x < low,
+        ishigh = high < x),
+        (!( islow | ishigh), islow, ishigh)
+)
```
simeonschaub, line 100: Sorry, I was missing a comma before:
```suggestion
@scalar_rule(
    clamp(x, low, high),
    @setup(
        islow = x < low,
        ishigh = high < x,
    ),
    (!(islow | ishigh), islow, ishigh),
)
```

freemin7: Great i was puzzled why it failed.

test/rulesets/Base/base.jl
```diff
@@ -115,6 +115,17 @@
         @test extern(dy) == extern(zeros(2, 5) .+ dy)
     end
 
+     @testset "ldexp" begin
+            rng = MersenneTwister(123456)
+            x, Δx, x̄ = 10rand(rng, 3)
+            Δz = rand(rng)
```
oxinabox, line 121: The use of local random number generators in the tests is a bad habit that needs to be removed from this code base. but for now lets just not add any more of it.
 The global random number generator is fine.
```suggestion
            x, Δx, x̄ = 10rand(3)
            Δz = rand()
```

test/rulesets/Base/base.jl
```diff
@@ -115,6 +115,17 @@
...
+
+            for n in (0,1,20)
+                #frule_test(ldexp, (x, Δx), (n, nothing))
```
oxinabox, line 124: Can this have an issue openned on ChainRulesTestUtils.jl
and then a commented added here:
```
# TODO: the following doesn't work because of bug in ChainRulesTestUtils
# see [URL]
```

test/rulesets/Base/base.jl
```diff
@@ -154,7 +165,7 @@
         end
     end
 
-    @testset "trinary ($f)" for f in (muladd, fma)
+    @testset "trinary ($f)" for f in (muladd, fma, clamp)
```
oxinabox, line 168: This testset is not surficient for testing `clamp` as it has modal behavour.
Need to explictly test the 3  cases

freemin7:
@oxinabox I addressed your criticism.

src/rulesets/Base/base.jl
mcabbott: This may have similar issues to #187, although (1) it's less obvious to me how often this is broadcasted vs. used in scalar code, and (2) perhaps getting AbstractDifferential here is better than getting Any:
```julia
julia> grad_clamp1(x, lo, hi) = lo<x<hi 

julia> grad_clamp2(x, lo, hi) = lo<x<hi ? One() : Zero()

julia> grad_clamp1.(randn(1,10),0,1)
1×10 BitArray{2}:
 0  0  0  1  1  0  0  0  0  1

julia> grad_clamp2.(randn(1,10),0,1)
1×10 Array{ChainRulesCore.AbstractDifferential,2}:
 One()  One()  One()  One()  One()  Zero()  Zero()  One()  Zero()  Zero()
```

test/rulesets/Base/base.jl
```diff
@@ -164,4 +176,22 @@
...
+
+        x, y, z = 2.5, 2., 3.  # in the middle
+        frule_test(f, (x, Δx), (y, Δy), (z, Δz))
+        rrule_test(f, Δk, (x, x̄), (y, ȳ), (z, z̄))
+
+
+        x, y, z = 4., 2., 3.  # to right
+        frule_test(f, (x, Δx), (y, Δy), (z, Δz))
+        rrule_test(f, Δk, (x, x̄), (y, ȳ), (z, z̄))
+    end
```
oxinabox, line 196: Since this test will never work for things other than clamp since it is handled the case about its boundries
would it be more clar to not use a `for` loop?

```suggestion
    @testset "clamp"  begin
        x̄, ȳ, z̄    = randn(3)
        Δx, Δy, Δz = randn(3)
        Δk = randn()

        x, y, z = 1., 2., 3.  # to left
        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))

        x, y, z = 2.5, 2., 3.  # in the middle
        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))

        x, y, z = 4., 2., 3.  # to right
        frule_test(clamp, (x, Δx), (y, Δy), (z, Δz))
        rrule_test(clamp, Δk, (x, x̄), (y, ȳ), (z, z̄))
    end
```

freemin7:
I am finished with this piece of code as far as i am concerned.
You can rebase it, go ahead.

oxinabox:
That was a very annoying rebase. Fair enough not wanting to do it

PR #175, opened 2020-04-20
Fix errors revealed by Zygote's tests

oxinabox:
Found these because of Zygote's tests
while working on https://github.com/FluxML/Zygote.jl/pull/366

It is hard to add tests for them because they are about testing that
we didn't define a rule we should not have,
and that we don't thunk to many times

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -117,7 +117,7 @@ end
 
 function rrule(::typeof(*), x::Number, y::Number)
     function times_pullback(ΔΩ)
-        return (NO_FIELDS,  @thunk(ΔΩ * y'), @thunk(x' * ΔΩ))
+        return (NO_FIELDS,  @thunk(ΔΩ * y), @thunk(x * ΔΩ))
     end
     return x * y, times_pullback
 end
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -137,6 +137,7 @@ function rrule(::typeof(identity), x::Tuple)
     # `identity(::Tuple)` returns multiple outputs;because that is how we think of
     # returning a tuple, so its pullback needs to accept multiple inputs.
     # `identity(::Tuple)` has one input, so its pullback should return 1 matching output
+    # see https://github.com/JuliaDiff/ChainRulesCore.jl/issues/152
     function identity_pullback(ȳs...)
         return (NO_FIELDS, Composite{typeof(x)}(ȳs...))
     end
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -45,7 +45,8 @@
 @scalar_rule(sinh(x), cosh(x))
 @scalar_rule(tanh(x), 1-Ω^2)
 
-@scalar_rule(acosh(x), inv(sqrt(x^2 - 1)))
+# Can't multiply though sqrt in acosh because of negative complex case for x
+@scalar_rule(acosh(x), inv(sqrt(x - 1) * sqrt(x + 1)))
 @scalar_rule(acoth(x), inv(1 - x^2))
 @scalar_rule(acsch(x), -inv(x^2 * sqrt(1 + x^-2)))
 @scalar_rule(acsch(x::Real), -inv(abs(x) * sqrt(1 + x^2)))
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -67,7 +67,9 @@
 @scalar_rule(-(x, y), (One(), -1))
 @scalar_rule(/(x, y), (inv(y), -(x / y / y)))
 @scalar_rule(\(x, y), (-(y / x / x), inv(x)))
-@scalar_rule(^(x, y), (ifelse(iszero(y), zero(Ω), y * x^(y - 1)), Ω * log(x)))
+
+#log(complex(x)) is require so it give correct complex answer for x<0 
+@scalar_rule(^(x, y), (ifelse(iszero(y), zero(Ω), y * x^(y - 1)), Ω * log(complex(x))))
 
 @scalar_rule(cbrt(x), inv(3 * Ω^2))
 @scalar_rule(inv(x), -Ω^2)
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -4,7 +4,7 @@
 
 @scalar_rule(abs(x::Real), sign(x))
 @scalar_rule(abs2(x), 2x)
-@scalar_rule(exp(x), Ω)
+@scalar_rule(exp(x::Real), Ω)
 @scalar_rule(exp10(x), Ω * log(oftype(x, 10)))
 @scalar_rule(exp2(x), Ω * log(oftype(x, 2)))
 @scalar_rule(expm1(x), exp(x))
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -132,3 +132,13 @@ function rrule(::typeof(identity), x)
     end
     return x, identity_pullback
 end
+
+function rrule(::typeof(identity), x::Tuple)
+    # `identity(::Tuple)` returns multiple outputs;because that is how we think of
+    # returning a tuple, so its pullback needs to accept multiple inputs.
+    # `identity(::Tuple)` has one input, so its pullback should return 1 matching output
+    function identity_pullback(ȳs...)
+        return (NO_FIELDS, Composite{typeof(x)}(ȳs...))
+    end
+    return x, identity_pullback
+end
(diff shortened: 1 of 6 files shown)


src/rulesets/Base/base.jl
```diff
@@ -4,7 +4,7 @@
 
 @scalar_rule(abs(x::Real), sign(x))
 @scalar_rule(abs2(x), 2x)
-@scalar_rule(exp(x), Ω)
+@scalar_rule(exp(x::Real), Ω)
```
simeonschaub, line 7: Is `exp` really the only complex function that causes problems with Zygote? Or are the others just not tested?

willtebbutt:
Seems unlikely. Note [this issue](https://github.com/JuliaDiff/FiniteDifferences.jl/issues/75) on FiniteDifferences. [This](https://github.com/JuliaDiff/FiniteDifferences.jl/pull/76) PR aims to fix it.

src/rulesets/Base/base.jl
oxinabox: Testing is ongoing

simeonschaub: I'm pretty sure the tests will fail for all our complex functions. Zygote does something quite different there.

oxinabox:
Ok this is ready for review though it needs the fix in TestUtils before tests will run

src/rulesets/Base/base.jl
oxinabox: turns out it only need conjugating on input and output, and othereise all is well.

oxinabox:
Once we have https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/25
this is good to go.
(I tested and it passes on that branch)

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -121,14 +121,11 @@ function rrule(::typeof(/), A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:R
...
         # Optimization note: dAᵀ, dBᵀ, dC are calculated no matter which partial you want
-        # this is not a problem if you want the 2nd or 3rd, but if you want the first, it
-        # is fairly wasteful
-        _, dC = dC_pb(Ȳ)
-        _, dBᵀ, dAᵀ = dS_pb(extern(dC))
-
-        # need to extern as  dAᵀ, dBᵀ  are generally `Thunk`s, which don't support adjoint
-        ∂A = @thunk last(dA_pb(extern(dAᵀ)))
-        ∂B = @thunk last(dA_pb(extern(dBᵀ)))
+                _, dC = dC_pb(Ȳ)
```
nickrobinson251, line 124: Typo/tab here?

src/rulesets/Base/base.jl
```diff
@@ -132,3 +132,13 @@ function rrule(::typeof(identity), x)
     end
     return x, identity_pullback
 end
+
+function rrule(::typeof(identity), x::Tuple)
+    # `identity(::Tuple)` returns multiple outputs;because that is how we think of
+    # returning a tuple, so its pullback needs to accept multiple inputs.
+    # `identity(::Tuple)` has one input, so its pullback should return 1 matching output
```
nickrobinson251, line 139: Link to issue CRC on this?

oxinabox:
```suggestion
    # `identity(::Tuple)` has one input, so its pullback should return 1 matching output
    # see https://github.com/JuliaDiff/ChainRulesCore.jl/issues/152
```

src/rulesets/LinearAlgebra/dense.jl
oxinabox:
```suggestion
        _, dC = dC_pb(Ȳ)
```

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -121,14 +121,11 @@ function rrule(::typeof(/), A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:R
     C, dC_pb = rrule(adjoint, Cᵀ)
     function slash_pullback(Ȳ)
         # Optimization note: dAᵀ, dBᵀ, dC are calculated no matter which partial you want
-        # this is not a problem if you want the 2nd or 3rd, but if you want the first, it
-        # is fairly wasteful
         _, dC = dC_pb(Ȳ)
-        _, dBᵀ, dAᵀ = dS_pb(extern(dC))
+        _, dBᵀ, dAᵀ = dS_pb(unthunk(dC))
 
-        # need to extern as  dAᵀ, dBᵀ  are generally `Thunk`s, which don't support adjoint
```
nickrobinson251, line 129: why did this comment go?

oxinabox: It was wrong, we don't exten them

oxinabox: we `unthunk` them, but this is now a clear and standard pattern so pointing it out is redundant
things should in general be unthunked before being fed to pullbacks

nickrobinson251: good point! https://github.com/JuliaDiff/ChainRules.jl/issues/177

src/rulesets/Base/base.jl
```diff
@@ -117,7 +118,7 @@ end
 
 function rrule(::typeof(*), x::Number, y::Number)
     function times_pullback(ΔΩ)
-        return (NO_FIELDS,  @thunk(ΔΩ * y'), @thunk(x' * ΔΩ))
+        return (NO_FIELDS,  @thunk(ΔΩ * y), @thunk(x * ΔΩ))
```
nickrobinson251, line 121: this is functionally the same right? The `'` before was just a no-op

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -40,17 +40,17 @@ end
 ##### `det`
 #####
 
-function frule((_, ẋ), ::typeof(det), x)
+function frule((_, ẋ), ::typeof(det), x::Union{Number, AbstractMatrix})
```
nickrobinson251, line 43: why the type annotations? (these do i believe cover all `det`/`logdet` methods in Base, but now not those extended by other packages, i suppose?) 

dose it warrant a comment in the code (or docs?) ?

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -6,8 +6,15 @@
...
-    function Diagonal_pullback(ȳ)
-        return (NO_FIELDS, @thunk(diag(ȳ)))
+    function Diagonal_pullback(ȳ::AbstractMatrix)
+        return (NO_FIELDS, diag(ȳ))
+    end
+    function Diagonal_pullback(ȳ::Composite)
+        # TODO: Assert about the primal type in the Composite, It should be Diagonal
+        # infact it should be exactly the type of `Diagonal(d)`
+        # but right now Zygote loses primal type information so we can't use it.
+        # See https://github.com/FluxML/Zygote.jl/issues/603
```
nickrobinson251, line 16: i am confused by this: how do we end up with a `Composite` but not a `Composite{P}` for the correct `P`?

test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -6,7 +6,15 @@
         rrule_test(Diagonal, D, (randn(rng, N), randn(rng, N)))
         # Concrete type instead of UnionAll
         rrule_test(typeof(D), D, (randn(rng, N), randn(rng, N)))
+
+        # TODO: replace this with a `rrule_test` once we have that working
+        # see https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/24
+        res, pb = rrule(Diagonal, [1, 4])
+        @test pb(10*res) == (NO_FIELDS, [10, 40])
+        comp = Composite{typeof(res)}(;diag=10*res.diag)  # this is the structure of Diagonal
```
nickrobinson251, line 14:
```suggestion
        comp = Composite{typeof(res)}(; diag=10*res.diag)  # this is the structure of Diagonal
```

src/rulesets/Base/base.jl
oxinabox: On real numbers it was a noop,
on complex numbers it was AFAICT just wrong?

src/rulesets/LinearAlgebra/dense.jl
oxinabox: Correct, I do not thing it deserves an annoation though,
it should be our practice to always include these when the primal method has them

src/rulesets/LinearAlgebra/structured.jl
oxinabox: Because in the Zygote PR we turn all `NamedTuples` into `Composite{Any}`

willtebbutt:
Cross referencing [this](https://github.com/FluxML/Zygote.jl/pull/366#issuecomment-620743107) -- I'm not sure why `abs2` doesn't work for complex numbers, but I can see why our current implementation of `abs` doesn't work.

src/rulesets/LinearAlgebra/dense.jl
nickrobinson251: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/155

src/rulesets/Base/base.jl
nickrobinson251: oh that makes much more sense (i never remember to think about Complex...)

oxinabox: I wish i could forget about complex

oxinabox: I have added test which fail without this change

Issue #144, opened 2020-01-07
Backpropagation-Friendly Eigendecomposition


ChrisRackauckas:
Not really just a new rule, just a better one?

https://papers.nips.cc/paper/8579-backpropagation-friendly-eigendecomposition.pdf

oxinabox:
I was about to open this, then I got distracted.
CF: https://github.com/invenia/Nabla.jl/issues/183

PR #179, opened 2020-04-28
[WIP] rrule for eigen decomposition 

ajozefiak:
This PR implements the rrule for eigen decomposition following the paper referenced in #144.

I have tested the implementation locally, but have yet to write tests for this PR as the implementation (as presented in the paper) is a bit unconventional to me and I think there will need to be modifications to my PR.

Several issues that I have encountered:
- The adjoint of the eigenvalues is not used in the derivation of the adjoint of the eigen decomposition (by the authors). This is reflected in my implementation.
- The paper assumes that the input matrix is symmetric positive semidefinite (i.e a covariance matrix).
- In the paper, the forward pass relies on svd as the underlying implementation of the eigen decomposition (this is possible due to the assumption that the input is symmetric PSD). I assume calling eigen(X) for the forward pass dispatches on a sensible algorithm. 
- The proposed algorithm is to back-propagate through the power iteration, which requires choosing the number of iterations. This leads me to adding an additional parameter for rrule since eigen(X) is not iterative and does not allow a choice of iterations.

My local tests show that the proposed adjoint for eigen decomposition agrees with finite differencing a power iteration algorithm, but in many situations differs from finite differencing the eigen decomposition. This could be that I am incorrectly interpreting the algorithm outside the vase where the adjoint of the eigenvectors dL/dV has at most one non-negative column.

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -269,7 +269,7 @@ end
 # this determines how many iterations are differentiated.
 # I am also iffy about the implementation of rrule for eigen in general, the paper
 # ignores the adjoints of the eigenvalues and also relies on the primal input being
-# symmetric positive semidefinte
+# symmetric positive semidefinite
 function rrule(::typeof(eigen), X::AbstractMatrix{<:Real};k=1)
     # The paper only computes the adjoint using the adjoint of the eigenvectors,
     # not the eigenvalues, and uses svd in the forward pass
@@ -295,14 +295,15 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where T <: Eigen
 end
 
 function eigen_rev(ΛV::Eigen,Λ̄,V̄,k)
-    
+
     Λ = ΛV.values
     V = ΛV.vectors
     A = V*diagm(Λ)*V'
 
     Ā = zeros(size(A))
     tempĀ = zeros(size(A))
-    # eigen(A).values are in descending order
+    # eigen(A).values are in descending order, this current implementation
+    # assumes that the input matrix positive definite
     for j = length(Λ):-1:1
         tempĀ = (I-V[:,j]*V[:,j]') ./ norm(A*V[:,j])
         for i = 1:k-1
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -265,10 +265,17 @@ end
 ##### `eigen`
 #####
 
-function rrule(::typeof(eigen), X::AbstractMatrix{<:Real})
-    F = eigen(X) # Should this change to SVD for symmetric psd case? Check if Eigen defaults to svd for symmetric PSD
+# k is the number of iterations of the power iteration that is 'used' in the forward pass,
+# this determines how many iterations are differentiated.
+# I am also iffy about the implementation of rrule for eigen in general, the paper
+# ignores the adjoints of the eigenvalues and also relies on the primal input being
+# symmetric positive semidefinte
+function rrule(::typeof(eigen), X::AbstractMatrix{<:Real};k=1)
+    # The paper only computes the adjoint using the adjoint of the eigenvectors,
+    # not the eigenvalues, and uses svd in the forward pass
+    F = eigen(X)
     function eigen_pullback(Ȳ::Composite{<:Eigen})
-        ∂X = @thunk(eigen_rev(F, Ȳ.values, Ȳ.vectors))
+        ∂X = @thunk(eigen_rev(F, Ȳ.values, Ȳ.vectors,k))
         return (NO_FIELDS, ∂X)
     end
     return F, eigen_pullback
@@ -288,7 +295,7 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where T <: Eigen
 end
 
 function eigen_rev(ΛV::Eigen,Λ̄,V̄,k)
-
+    
     Λ = ΛV.values
     V = ΛV.vectors
     A = V*diagm(Λ)*V'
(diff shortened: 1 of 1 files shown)


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -260,3 +260,58 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+
+#####
+##### `eigen`
+#####
+
+# k is the number of iterations of the power iteration that is 'used' in the forward pass,
+# this determines how many iterations are differentiated.
+# I am also iffy about the implementation of rrule for eigen in general, the paper
+# ignores the adjoints of the eigenvalues and also relies on the primal input being
+# symmetric positive semidefinite
```
oxinabox, line 272: Can we include a link to the paper, both URL, and something in plain text (for future prooofing)
in the comment.

oxinabox:
Can you clarify what you mean by?

> The adjoint of the eigenvalues is not used in the derivation of the adjoint of the eigen decomposition 

Are the eigenvalues not scalars?
Why is surprising not to use the adjont of the eigenvalues ?

---

I am sorting out someone to check the math on this.
This is cool.

src/rulesets/LinearAlgebra/factorization.jl
ajozefiak: I can do that!

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -260,3 +260,58 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+
+    Ā = zeros(size(A))
+    tempĀ = zeros(size(A))
+    # eigen(A).values are in descending order, this current implementation
+    # assumes that the input matrix positive definite
+    for j = length(Λ):-1:1
+        tempĀ = (I-V[:,j]*V[:,j]') ./ norm(A*V[:,j])
+        for i = 1:k-1
+            tempĀ += A^i * (I-V[:,j]*V[:,j]') ./ (norm(A*V[:,j])^(i+1))
+        end
```
sethaxen, line 311: Have you considered using Horner's method (i.e. `evalpoly`) to compute these terms? It will avoid lots of matrix powers. Specifically, factor out `P = (I-V[:,j]*V[:,j]') ./ norm(A*V[:,j])`, and take `x = A ./ norm(A*V[:,j])` and `p = fill(I, (k - 1))`. Then `evalpoly(x, p)` will evaluate the polynomial efficiently. (for backwards compatibility, you may need `@evalpoly(x, p...)`.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -260,3 +260,58 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+        tempĀ *= V̄[:,j]*V[:,j]'
```
sethaxen, line 312: If you do the above, then you have `tempĀ *= (I-V[:,j]*V[:,j]') V̄[:,j]*V[:,j]'  ./ norm(A*V[:,j])`. Setting `v = V[:,j]`, I think this can evaluate this more efficiently as
```julia
w = (v̄ .-v .* dot(v, v̄)) ./ norm(A*v)
tempĀ = (tempĀ * w) * v'
```
This basically just tries to minimize the number of matrix products.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -260,3 +260,58 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+#####
+##### `eigen`
+#####
+
+# k is the number of iterations of the power iteration that is 'used' in the forward pass,
+# this determines how many iterations are differentiated.
+# I am also iffy about the implementation of rrule for eigen in general, the paper
+# ignores the adjoints of the eigenvalues and also relies on the primal input being
+# symmetric positive semidefinite
+function rrule(::typeof(eigen), X::AbstractMatrix{<:Real};k=1)
```
sethaxen, line 273: Since this only applies to real symmetric matrices, should the type of `X` be `Symmetric{<:Real}`?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -260,3 +260,58 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+    # The paper only computes the adjoint using the adjoint of the eigenvectors,
+    # not the eigenvalues, and uses svd in the forward pass
+    F = eigen(X)
```
sethaxen, line 276: Likewise, since the method only applies to positive semi-definite matrices, perhaps an error or warning should be thrown if any of the eigenvalues are negative?

sethaxen:
> Does it make sense that it doesn't pull back the cotangent of the eigenvectors at all? Take a function that calls `eigen` to get the eigenvalues only and discards the eigenvectors. This method should pull back an sensitivity of 0 on A, but the usual version (e.g. see Zygote's at https://github.com/FluxML/Zygote.jl/blob/master/src/lib/array.jl#L570-L588) would pull back a nonzero sensitivity. I haven't studied the paper yet, but that looked funny.

Answering my own question, since by comparison of equations 8 and 10, they argue that as `k → ∞`, the PI gradients converge to the gradients of the eigendecomposition minus the term that comes from the cotangent of the eigenvalues (which they treat as zero), I think you could just add that back in to get the full cotangent on `A`. That term is not unstable in the degenerate case anyways.

> The proposed algorithm is to back-propagate through the power iteration, which requires choosing the number of iterations. This leads me to adding an additional parameter for rrule since eigen(X) is not iterative and does not allow a choice of iterations.

Would it be better to provide a `kmax` and use a max error term to choose a `k` using their Eq 14?

Issue #180, opened 2020-04-29
Clean out all of the local random number generators from the test


oxinabox:
We should just use the global random number generator.
It is reseeded at the start of each testset.

This will simplify the testcode some what, which is always nice.

Should be an easy, though kind long, change to make.
just ralcing `x = rand(rng, ...)` with `x=rand(...)`
and deleting `rng = MersenneTwister(...)`

mattBrzezinski:
One small change needs to be made to `ChainRulesTestUtils.jl` to [genereate_well_conditioned_matrix()](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/2d20cf88bfa2d1f3c9fbb295f0a12a526236e44a/src/data_generation.jl#L2) to remove the `rng` as a param.

oxinabox:
It should keep it but should default to using the global one if not provided

PR #181, opened 2020-04-29
Use global random number generator

mattBrzezinski:
Resolves: #180 

From the Test [documentation](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@testset) for `@testset`:

> Before the execution of the body of a @testset, there is an implicit call to Random.seed!(seed) where seed is the current seed of the global RNG.

So we do not need to manually seed our `@testsets`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.5.1"
+version = "0.5.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/test/rulesets/LinearAlgebra/dense.jl
+++ b/test/rulesets/LinearAlgebra/dense.jl
@@ -1,118 +1,115 @@
 @testset "linalg" begin
     @testset "dot" begin
         @testset "Vector" begin
-            rng, M = MersenneTwister(123456), 3
-            x, y = randn(rng, M), randn(rng, M)
-            ẋ, ẏ = randn(rng, M), randn(rng, M)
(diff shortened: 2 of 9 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.5.1"
+version = "0.5.2"
```
oxinabox, line 3: this is that rare change that don't need a release, since it is only changing tests

test/rulesets/LinearAlgebra/dense.jl
```diff
@@ -1,118 +1,115 @@
...
+            rrule_test(dot, randn(), (x, x̄), (y, ȳ))
         end
     end
     @testset "inv" begin
-        rng, N = MersenneTwister(123456), 3
-        B = generate_well_conditioned_matrix(rng, N)
-        frule_test(inv, (B, randn(rng, N, N)))
-        rrule_test(inv, randn(rng, N, N), (B, randn(rng, N, N)))
+        N = 3
+        B = generate_well_conditioned_matrix(Random.GLOBAL_RNG, N)
```
nickrobinson251, line 30: we should fix this in ChainrulesTestUtils https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/27#issuecomment-621417499

mattBrzezinski: Yeah that makes more sense! I'll make the changes and re-open that CR and we can release that version and make the necessary changes in here.

mattBrzezinski:
Just going to wait until https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/27 goes in, then I'll make the change to call the new function and we should be good to go!

Issue #154, opened 2020-01-18
generic frule for constructors?


oxinabox:
Like for https://github.com/JuliaDiff/ChainRules.jl/issues/153
there is a generic frule for default constructors.
It relies on the arguments being the same as the fields so would have to detect that.

Its basically:

```
function frule(::Type{P}, args..., _, dargs...) where P
     y  = P(args...)
     dy = Composite{P}(; zip(fieldnames(P, dargs)...)
end
```

the more serious implementation might be:
```
function frule(::Type{P}, all_args...) where P
     nargs = fieldcount(P)
     length(all_args) == 2nargs + 1 || return nothing
     args = @inbounds all_args[1:nargs]
     all(typeof.(args) .== fieldtypes(P)) || return nothing
     dargs = @inbounds all_args[end-nargs : end]  # skip dself as constructors never functors
     y  = P(args...)
     dy = Composite{P}(; zip(fieldnames(P, dargs)...)
end
```

But this may well need to be written as a generated function if it doesn't constant fold good.

oxinabox:
Related:
1. We need to have mutable composite types. https://github.com/JuliaDiff/ChainRulesCore.jl/issues/105
2. We need a way to work out the differential type for a given primal type, so we can declare the differential arrays etc (this can default for Any) https://github.com/JuliaDiff/ChainRulesCore.jl/issues/106

Issue #153, opened 2020-01-18
generic `rrule` for all `getfield`


oxinabox:
If I am correct: the rrule for `getfield` is always:

```
function rrule(::typeof(getfield), val::P, field::Symbol) where P
    getfield_pullback(dy) = NO_FIELDS, Composite{P}(; (name => dy,)...)
    return getfield(val, field), getfield_pullback
end
```

willtebbutt:
Looks correct to me 🤷‍♂

cossio:
There is a typo, `name` should be `:field`. 

Shouldn't this also return zeros for the other fields (besides the requested `field` for which the gradient is 1)?

willtebbutt:
> Shouldn't this also return zeros for the other fields (besides the requested field for which the gradient is 1)?

`Composite` handles this lazily -- not specifying a field is essentially the same as setting it to `Zero` :)

PR #182, opened 2020-05-01
Add rules for all Symmetric/Hermitian constructors

### LINKED ISSUE #178: handle `uplo` argument for Symmetric
Zygote handles the 2 arg function for Symmetric
ChainRules rightnow only handles the 1 arg function

sethaxen:
Fixes #178 by porting the rules from Zygote, along with the `Hermitian` rules (see https://github.com/FluxML/Zygote.jl/blob/0b3e32d5c0f8e5ef3b061a6f84ed5505f5a202a0/src/lib/array.jl#L454-L491).

Tests currently do not pass I think because ChainRulesTestUtils seems to interpret the `:U` variable as a variable of name `U`, which does not exist. We may also need FiniteDifference v0.10.0 for the complex tests.

### DIFF
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -78,6 +78,13 @@ function rrule(T::Type{<:LinearAlgebra.HermOrSym}, A::AbstractMatrix, uplo)
     return Ω, HermOrSym_pullback
 end
 
+function frule((_, ΔA), TM::Type{<:Matrix}, A::LinearAlgebra.HermOrSym)
+    return TM(A), TM(_symherm_forward(A, ΔA))
+end
+function frule((_, ΔA), ::Type{Array}, A::LinearAlgebra.HermOrSym)
+    return Array(A), Array(_symherm_forward(A, ΔA))
+end
+
 function rrule(TM::Type{<:Matrix}, A::TA) where {TA<:LinearAlgebra.HermOrSym}
     function Matrix_pullback(ΔΩ)
         T∂A = _symhermtype(TA){eltype(ΔΩ),typeof(ΔΩ)}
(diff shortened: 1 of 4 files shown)


sethaxen:
I left out the Symmetric -> Matrix and Matrix -> Symmetric conversions at https://github.com/FluxML/Zygote.jl/blob/0b3e32d5c0f8e5ef3b061a6f84ed5505f5a202a0/src/lib/array.jl#L493-L496 but could add those as well. Has there been discussion on a generic `rrule` for such conversions? There were some in various Zygote issues but they didn't get very far.

oxinabox:
> Has there been discussion on a generic rrule for such conversions?
I don't think so, unless it is covered by  #153, or #154
open an issue?

sethaxen:
Blocked by #198

sethaxen:
Okay, this is ready for review. I also unified the constructors and also added `frule`s and rules for conversion to a `Matrix`.

Since between this and #193 there's a disproportionately large amount of `Symmetric` and `Hermitian` code in `structured.jl`, I wonder if it's worth splitting that into its own `symmetric.jl` file, which would also mirror the file structure in LinearAlgebra.

willtebbutt:
I'll be reviewing tomorrow morning, please don't merge before then.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -63,18 +63,84 @@ function rrule(::typeof(*), D::Diagonal{<:Real}, V::AbstractVector{<:Real})
...
-    function Symmetric_pullback(ȳ)
-        return (NO_FIELDS, @thunk(_symmetric_back(ȳ)))
+function frule((_, ΔA, _), T::Type{<:LinearAlgebra.HermOrSym}, A::AbstractMatrix, uplo)
+    return T(A, uplo), T(ΔA, uplo)
+end
+
+function rrule(T::Type{<:LinearAlgebra.HermOrSym}, A::AbstractMatrix, uplo)
+    Ω = T(A, uplo)
+    function HermOrSym_pullback(ΔΩ)
+        return (NO_FIELDS, @thunk(_symherm_back(T, ΔΩ, Ω.uplo)), DoesNotExist())
```
willtebbutt, line 76: Do we need a thunk here? The cotangent w.r.t. A is the only meaningful, so I would imagine that it would always get used somewhere, but my intuition for this isn't great, perhaps yours is better.

sethaxen: Probably not? I had thought that the convention was to always thunk in reverse unless 1) there's only one cotangent (not counting the `NO_FIELDS`) or 2) sometimes for scalar functions (e.g. those made with `@scalar_rule`). I personally think modifying (1) to be "there's only one cotangent that a user could reasonably want" makes perfect sense.

willtebbutt: @oxinabox do you agree? If so I'll open a PR to the docs to update guidance.

willtebbutt:
@sethaxen please feel free to merge this whenever. It's good to go IMO.

PR #185, opened 2020-05-06
Add rules for FastMath

### LINKED ISSUE #174: FastMath operations
It would be good to have support for `@fastmath`, especially the functions that `ccall` libm.

Not sure if it's the right approach here, but [in Zygote](https://github.com/FluxML/Zygote.jl/blob/359e586766129878ca0e56121037ed80afda6289/src/lib/number.jl#L77) we did this by looping over the usual `DiffRules`.

This is a bit of a blocker for https://github.com/FluxML/Zygote.jl/pull/366, because we want to remove DiffRules as part of that, but currently doing so would cause a regression on this.

oxinabox:
Closes #174 
(cc @mikeinnes)

Needs tests still.
The code has tests, in that it is still the same basic source as before, so the base cases are tested.
there are no tests of the fast math versions though.
Those tests can be generated in the same way however, if that is acceptable?

---
This PR also adds some missing tests for things that have fast math varients:
  - `expm1`
  - binary `+` and `-`, `/`, `\`, `max`, `min`

They are added in the trivial way, by adding them to the loops identical to others of their class.

### DIFF
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -25,7 +25,7 @@ const FASTABLE_AST = quote
                 rrule_test(sincos, Δz, (x, x̄))
             end
         end
-    end  # Trig
+    end
 
     @testset "exponents" begin
         for x in (-0.1, 6.4)
@@ -51,11 +51,9 @@ const FASTABLE_AST = quote
     @testset "Unary complex functions" begin
         for x in (-4.1, 6.4)
             test_scalar(abs, x)
-
             test_scalar(angle, x)
             test_scalar(abs2, x)
             test_scalar(conj, x)
-
         end
     end
 
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -1,0 +1,107 @@
+# Add tests to the quote for functions with  FastMath varients.
+const FASTABLE_AST = quote
+    @testset "Trig" begin
+        @testset "Basics" for x = (Float64(π)-0.01, Complex(π, π/2))
+            test_scalar(sin, x)
+            test_scalar(cos, x)
+            test_scalar(tan, x)
+        end
+        @testset "Hyperbolic" for x = (Float64(π)-0.01, Complex(π-0.01, π/2))
+            test_scalar(sinh, x)
+            test_scalar(cosh, x)
+            test_scalar(tanh, x)
+        end
+        @testset "Inverses" for x = (0.5, Complex(0.5, 0.25))
+            test_scalar(asin, x)
+            test_scalar(acos, x)
+            test_scalar(atan, x)
+        end
+        @testset "Multivariate" begin
+            @testset "sincos" begin
+                x, Δx, x̄ = randn(3)
+                Δz = (randn(), randn())
+
+                frule_test(sincos, (x, Δx))
+                rrule_test(sincos, Δz, (x, x̄))
+            end
+        end
+    end  # Trig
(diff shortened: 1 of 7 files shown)


oxinabox:
Generated via:

```julia
using MacroTools
using ChainRulesCore
src_code = String(read(joinpath(@__DIR__, "old_base.jl")))
src_code = "begin $src_code end"
src_ast = Meta.parse(src_code)
block_asts = [MacroTools.striplines(x) for x in src_ast.args if x isa Expr]  # don't want anything that don't have subexpressions
transformed_asts_once = Base.FastMath.make_fastmath.(block_asts)
is_fastable = block_asts .!= transformed_asts_once
transformed_asts_once[is_fastable] .|> println

open(joinpath(@__DIR__, "fastable.jl"), "w") do fh
    println.(Ref(fh), block_asts[is_fastable])
end

# check against:
transformed_asts_once[is_fastable]


open(joinpath(@__DIR__, "base.jl"), "w") do fh
    println.(Ref(fh), block_asts[.!is_fastable])
end

transformed_asts_once[.!is_fastable]
```

Then a bunch of manual stuff.
The compiler helpfully tells you when you define things twice, so that makes it easier to find where there were rules generated that were for things that did not have fastmath varients

test/rulesets/Base/fastmath_able.jl
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    @testset "Unary complex functions" begin
+        for x in (-4.1, 6.4)
+            test_scalar(abs, x)
+
+            test_scalar(angle, x)
+            test_scalar(abs2, x)
+            test_scalar(conj, x)
+
```
mattBrzezinski, line 58:
```suggestion
```

test/rulesets/Base/fastmath_able.jl
```diff
@@ -0,0 +1,107 @@
...
+                test_scalar(log10, x)
+                test_scalar(log1p, x)
+            end
+        end
+    end
+
+    @testset "Unary complex functions" begin
+        for x in (-4.1, 6.4)
+            test_scalar(abs, x)
+
```
mattBrzezinski, line 54:
```suggestion
```

test/rulesets/Base/fastmath_able.jl
```diff
@@ -0,0 +1,107 @@
...
+
+
+    @testset "sign" begin
+        @testset "at points" for x in (-1.1, -1.1, 0.5, 100)
+            test_scalar(sign, x)
+        end
+
+        @testset "Zero over the point discontinuity" begin
+            # Can't do finite differencing because we are lying
```
mattBrzezinski, line 86: Curious, what do you mean by this?

oxinabox: There is some math involve but basically the defintion of derivative is taking limit of small steps around a point.
It  is consider to be not defined if when you take those steps to the left and to the right you get a different answer.
In this case you get `lim x->0 +1/x` and `lim x-> 0 -1/x`
which we can say is something like positive and negativ infinity.

There is some extension of the math for deritivatives called the subgradient method.
and one of the outcomes of it is that it is good to treat derivatives that are not defined,
but that are such that from one side the limit is positive and from the other it is negative as being zero.
This is because one of the main uses of the dervative is to find local minima,
which typically are defined as being derivative zero at that point.
But for in cases where the dervative at that point is not define then it is also true that if the derivative on either side is opposite in sign it is also a local minima

oxinabox:
@mattBrzezinski  is this good to go?

PR #190, opened 2020-05-08
Add rules for evalpoly

sethaxen:
~Most of the pushforward and pullback for `evalpoly` can be written as more calls to `evalpoly`, which is very efficient.~

Edit: This is true in the scalar case but not the matrix. In the current implementation, both the matrix and scalar polynomials have forward and reverse rules on the same order as `evalpoly` itself.

~Currently all but one of the tests fail for various reasons. I think the testers in ChainRulesTestUtils don't like when one of the inputs is a `Tuple`.~ Also, I don't expect the complex tests to pass until this package uses FiniteDifferences v0.10.0.

Edit: Oh, and I see that ChainRules doesn't test against 1.4 yet at all.

~@oxinabox I also tested against https://github.com/FluxML/Zygote.jl/pull/366 and may have hit a bug where Zygote doesn't like that the adjoint of `p` is a `Tuple`, but I'm not sure.~

Edit: Now that I'm using `Composite`, it works fine in Zygote.

```julia
julia> using Zygote

julia> x = randn();

julia> pv = randn(10);

julia> p = Tuple(pv);

julia> evalpoly(x, pv)
126.48423475727726

julia> evalpoly(x, p)
126.48423475727726

julia> Zygote.gradient(evalpoly, x, pv) # seems to work fine
(1722.5531441128583, [1.0, 1.7894268524990125, 3.2020484604445225, 5.72983149812255, 10.253114343035136, 18.347198127169843, 32.83096899687731, 58.748617516574825, 105.12635373135284, 188.1159202721925])

julia> Zygote.gradient(evalpoly, x, p) # I guess not happy about the adjoint of p being a tuple?
ERROR: MethodError: no method matching conj(::NTuple{10,Float64})
Closest candidates are:
  conj(::Missing) at missing.jl:100
  conj(::ForwardDiff.Dual) at /Users/saxen/.julia/packages/ForwardDiff/cXTw0/src/dual.jl:367
  conj(::Real) at number.jl:167
  ...
Stacktrace:
 [1] wrap_chainrules_output at /Users/saxen/projects/Zygote.jl/src/compiler/chainrules.jl:49 [inlined]
 [2] map(::typeof(Zygote.wrap_chainrules_output), ::Tuple{ChainRulesCore.Zero,ChainRulesCore.Thunk{ChainRules.var"#51#56"{Float64,Float64,NTuple{10,Float64}}},ChainRulesCore.Thunk{ChainRules.var"#52#57"{Float64,Float64,NTuple{10,Float64}}}}) at ./tuple.jl:159
 [3] wrap_chainrules_output at /Users/saxen/projects/Zygote.jl/src/compiler/chainrules.jl:50 [inlined]
 [4] wrap_chainrules_pullback at /Users/saxen/projects/Zygote.jl/src/compiler/chainrules.jl:86 [inlined]
 [5] ZBack at /Users/saxen/projects/Zygote.jl/src/compiler/chainrules.jl:99 [inlined]
 [6] (::Zygote.var"#41#42"{Zygote.ZBack{ChainRules.var"#evalpoly_pullback#55"{Float64,NTuple{10,Float64}}}})(::Float64) at /Users/saxen/projects/Zygote.jl/src/compiler/interface.jl:45
 [7] gradient(::Function, ::Float64, ::Vararg{Any,N} where N) at /Users/saxen/projects/Zygote.jl/src/compiler/interface.jl:54
 [8] top-level scope at REPL[8]:1
```

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -129,15 +129,18 @@ if VERSION ≥ v"1.4"
             end
             push!(exs, :(y1 = $ex))
             Expr(:block, exs..., :(y1, ($(vars...), y1)))
-        else # fallback when can't generate code
-            N = length(p)
-            y = one(x) * p[N]
-            ntuple(N) do i
-                i == 1 && return y
-                return y = muladd(x, y, p[N - i + 1])
-            end
-            y, ys
+        else
+            _evalpoly_intermediates_fallback(x, p)
         end
+    end
+    function _evalpoly_intermediates_fallback(x, p::Tuple)
+        N = length(p)
+        y = one(x) * p[N]
+        ys = ntuple(N) do i
+            i == 1 && return y
+            return y = muladd(x, y, p[N - i + 1])
+        end
+        return y, ys
     end
     function _evalpoly_intermediates(x, p)
         N = length(p)
@@ -155,9 +158,7 @@ if VERSION ≥ v"1.4"
     #     1) x is a UniformScaling, pᵢ is a matrix
     #     2) x is a matrix, pᵢ is a UniformScaling
     function _evalpoly_back(x, p::Tuple, ys, Δy)
-        x′ = x'
-        ∂x = zero(Δy)
-        ∂p = if @generated
+        return if @generated
             exs = []
             vars = []
             ex = :(Δy)
@@ -169,15 +170,27 @@ if VERSION ≥ v"1.4"
                 push!(exs, :(∂x = muladd($∂yi, ys[$(N - i)]', ∂x)))
                 ex = :(x′ * $∂yi)
             end
-            Expr(:block, exs..., :($(vars...), $ex))
-        else # fallback when can't generate code
-            ∂yi = Δy
-            N = length(p)
-            ntuple(N) do i
-                i == 1 && return ∂yi
-                ∂x = muladd(∂yi, ys[N - i + 1]', ∂x)
-                return ∂yi = x′ * ∂yi
-            end
+            Expr(
+                :block,
+                :(x′ = x'),
+                :(∂x = zero(Δy)),
+                exs...,
+                :(∂p = ($(vars...), $ex)),
+                :(∂x, Composite{typeof(p),typeof(∂p)}(∂p)),
+            )
+        else
+            _evalpoly_back_fallback(x, p, ys, Δy)
+        end
+    end
+    function _evalpoly_back_fallback(x, p::Tuple, ys, Δy)
+        x′ = x'
+        ∂x = zero(Δy)
+        ∂yi = Δy
+        N = length(p)
+        ∂p = ntuple(N) do i
+            i == 1 && return ∂yi
+            ∂x = muladd(∂yi, ys[N - i + 1]', ∂x)
+            return ∂yi = x′ * ∂yi
         end
         return ∂x, Composite{typeof(p),typeof(∂p)}(∂p)
     end
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -169,22 +169,23 @@ if VERSION ≥ v"1.4"
             exs = []
             vars = []
             N = length(p.parameters)
-            push!(vars, :(_evalpoly_backp(p[1], Δy))) # ∂p1
             for i in 2:(N - 1)
                 ∂pi = Symbol("∂p", i)
                 push!(vars, ∂pi)
-                push!(exs, :(∂yi = x′ * ∂yi))
                 push!(exs, :(∂x = _evalpoly_backx(x, ys[$(N - i)], ∂x, ∂yi)))
                 push!(exs, :($∂pi = _evalpoly_backp(p[$i], ∂yi)))
+                push!(exs, :(∂yi = x′ * ∂yi))
             end
-            push!(vars, :(_evalpoly_backp(p[$N], x′ * ∂yi))) # ∂pN
+            push!(vars, :(_evalpoly_backp(p[$N], ∂yi))) # ∂pN
             Expr(
                 :block,
                 :(x′ = x'),
                 :(∂yi = Δy),
+                :(∂p1 = _evalpoly_backp(p[1], ∂yi)),
                 :(∂x = _evalpoly_backx(x, ys[$(N - 1)], ∂yi)),
+                :(∂yi = x′ * ∂yi),
                 exs...,
-                :(∂p = ($(vars...),)),
+                :(∂p = (∂p1, $(vars...))),
                 :(∂x, Composite{typeof(p),typeof(∂p)}(∂p)),
             )
         else
@@ -195,16 +196,18 @@ if VERSION ≥ v"1.4"
         x′ = x'
         ∂yi = Δy
         N = length(p)
+        ∂p1 = _evalpoly_backp(p[1], ∂yi)
         ∂x = _evalpoly_backx(x, ys[N - 1], ∂yi)
+        ∂yi = x′ * ∂yi
         ∂p = (
-            _evalpoly_backp(p[1], ∂yi), # ∂p1
+            ∂p1,
             ntuple(N - 2) do i
-                ∂yi = x′ * ∂yi
                 ∂x = _evalpoly_backx(x, ys[N-i-1], ∂x, ∂yi)
                 ∂pi = _evalpoly_backp(p[i+1], ∂yi)
+                ∂yi = x′ * ∂yi
                 return ∂pi
             end...,
-            _evalpoly_backp(p[N], x′ * ∂yi), # ∂pN
+            _evalpoly_backp(p[N], ∂yi), # ∂pN
         )
         return ∂x, Composite{typeof(p),typeof(∂p)}(∂p)
     end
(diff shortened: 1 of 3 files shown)


src/rulesets/Base/base.jl
```diff
@@ -88,3 +88,55 @@ function rrule(::typeof(identity), x::Tuple)
...
+    end
+
+    # This is a geometric progression, that is ∂p = Δy * (I, x, x², …, xⁿ)'
+    # TODO: Handle case where x is a matrix and p is UniformScaling
+    function _evalpoly_backp(Δy, x, p::Tuple)
+        N = length(p)
+        ∂p = ntuple(_ -> Δy, N)
+        x′ = x'
+        for i in 2:N
+            ∂p = Base.setindex(∂p, ∂p[i - 1] * x′, i)
```
oxinabox, line 128: TIL `Base.setindex` on `Tuple`

src/rulesets/Base/base.jl
```diff
@@ -88,3 +88,55 @@ function rrule(::typeof(identity), x::Tuple)
...
+        end
+        return ∂p
```
oxinabox, line 130: Pullbacks can't return `Tuple` as a `Tuple` is not a valid differential type,
because `Tuple` do not define `+` or `zero`
Instread one should return a `Composite`

```suggestion
        return Composite{typeof(p)}(∂p...)
```
or
```suggestion
        return Composite{typeof(p), typeof(∂p)}(∂p)
```

src/rulesets/Base/base.jl
oxinabox: Is this actually faster than
either:
```julia
ntuple(N) do i
    Δy*x′^i
end
```
or
```julia
last_val = Δy
tt = ntuple(N-1) do i
    last_val*=x′
end
∂p = (Δy, tt...)
```

or
```julia
last_val = Δy
∂p = ntuple(N) do i
    i==1 && return  Δy 
    last_val*=x′
end
```

src/rulesets/Base/base.jl
```diff
@@ -88,3 +88,55 @@ function rrule(::typeof(identity), x::Tuple)
...
+        @inbounds q = (1:(N - 1)) .* view(p, 1:(N - 1))
+        return q
+    end
+
+    # This is a geometric progression, that is ∂p = Δy * (I, x, x², …, xⁿ)'
+    # TODO: Handle case where x is a matrix and p is UniformScaling
+    function _evalpoly_backp(Δy, x, p::Tuple)
+        N = length(p)
+        ∂p = ntuple(_ -> Δy, N)
+        x′ = x'
```
oxinabox, line 126: Should this really be adjoint and not transpose?

```suggestion
        x′ = transpose(x)
```

test/rulesets/Base/base.jl
```diff
@@ -121,6 +121,28 @@
...
+                    (randn(T, 5), randn(T, 5)),
+                )
+                rrule_test(
+                    evalpoly, randn(T), (randn(T), randn(T)),
+                    (Tuple(randn(T, 5)), Tuple(randn(T, 5))),
+                )
+            end
+        end
+    end
+end
```
oxinabox, line 144:
```suggestion
```

test/rulesets/Base/base.jl
```diff
@@ -121,6 +121,28 @@
         )
     end
 
+if VERSION ≥ v"1.4"
+    @testset "evalpoly" begin
```
oxinabox, line 125: If we do the following we can follow the same indenting everywhere
```suggestion
    VERSION ≥ v"1.4" && @testset "evalpoly" begin
```

src/rulesets/Base/base.jl
sethaxen: Quite a bit:
```julia
    function _evalpoly_backp(Δy, x, p::Tuple)
        N = length(p)
        ∂p = ntuple(_ -> Δy, N)
        x′ = x'
        for i in 2:N
            ∂p = Base.setindex(∂p, ∂p[i - 1] * x′, i)
        end
        return ∂p
    end
    function _evalpoly_backp2(Δy, x, p::Tuple)
        N = length(p)
        x′ = x'
        ∂p = ntuple(N) do i
            Δy * x′^(i - 1)
        end
        return ∂p
    end
    function _evalpoly_backp3(Δy, x, p::Tuple)
        N = length(p)
        x′ = x'
        last_val = Δy
        tt = ntuple(N - 1) do i
            last_val *= x′
        end
        ∂p = (Δy, tt...)
        return ∂p
    end
    function _evalpoly_backp4(Δy, x, p::Tuple)
        N = length(p)
        x′ = x'
        last_val = Δy
        ∂p = ntuple(N) do i
            i == 1 && return Δy 
            return last_val *= x′
        end
        return ∂p
    end
```

```julia
julia> using BenchmarkTools

julia> Δy, x, p = randn(), randn(), Tuple(randn(10));

julia> @btime _evalpoly_backp(Δy, x, p);
  53.628 ns (1 allocation: 96 bytes)

julia> @btime _evalpoly_backp2(Δy, x, p);
  261.738 ns (1 allocation: 96 bytes)

julia> @btime _evalpoly_backp3(Δy, x, p);
  1.356 μs (34 allocations: 704 bytes)

julia> @btime _evalpoly_backp4(Δy, x, p);
  745.799 ns (23 allocations: 464 bytes)
```

src/rulesets/Base/base.jl
sethaxen: Assuming ChainRules is following the same conventions with complex numbers as Zygote does, then yes this should be.

src/rulesets/Base/base.jl
sethaxen: > Pullbacks can't return `Tuple` as a `Tuple` is not a valid differential type,
> because `Tuple` do not define `+` or `zero`
> Instread one should return a `Composite`

Ah good point, that's what I was missing.

>(in contrast, our, or at least my, opinion on the handling of complex numbers is much less strong. But does still disagree with Zygote.)

Can I read about those disagreements somewhere?

sethaxen: Does the input differential for `p` in the `frule` likewise need to be a `Composite` when `p` is a tuple?

src/rulesets/Base/base.jl
oxinabox: Nice, I am going to start using this

src/rulesets/Base/base.jl
oxinabox: It should be be a `Composite` yes, but it is less clear to me if it should be constrained to be.

src/rulesets/Base/base.jl
oxinabox: ChainRules does not follow the same convention a Zygote.
Its a thing we could change, but right now we work with the gradient of complex numbers, not its conjugate.

I don't actually know if we have an issue for that, feel free to open one on ChainRulesCore if we don't.

src/rulesets/Base/base.jl
sethaxen: @oxinabox Actually I realized I was just lucky here that I picked `length(p) == 10`. if `length(p) > 10`, the version with `setindex` has _terrible_ performance.

sethaxen:
This PR is basically feature complete, and tests pass. It's still marked as WIP because:
- Pending clarification in https://github.com/JuliaDiff/ChainRulesCore.jl/issues/159, the two adjoints may need to be made transposes.
- Complex tests are currently disabled until FD v0.10.0 is supported. They'll also depend on https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/33, since we use Horner's method for the polynomial, not the method that base uses, so they'll be approximately equal but not exactly. I don't think this should hold up this PR.
- Matrix polynomials where either `x` or elements of `p` are `UniformScaling` are not supported. There are currently no rules for `UniformScaling`, so I don't think that should hold up this PR.

sethaxen:
Yeah, the weird `muladd` in a macro giving an approximately equal but not equal result as `muladd` not in a macro on Travis issue came up again. No idea why. But tests pass now. Thanks!

manuelbb-upb:
I just found a [question on discourse](https://discourse.julialang.org/t/why-does-zygote-gradient-not-work-on-polynomials-jl/57031) asking why Zygote won't differentiate a polynomial created with `Polynomials.jl`.
I hope this is the right place to comment: I guess it could be due to the custom rules for `evalpoly`.
Take the example code from the original question:
```julia 
using Zygote
using Polynomials
p=Polynomial([1.0,0.0,3.0,4.0])
@show derivative(p)(5.0)
@show gradient(p,5.0)
```
This errors because we need a `similar` method for `p` in `_evalpoly_intermediate(x, p)`. No problem:
```
Base.similar( :: Polynomial, etype = eltype(p), len = length(p.coeffs)) = similar( p.coeffs, etype, len )
```
Now `Zygote.gradient` can calculate the gradient, but gives the wrong numerical result.
However,
```
gradient( evalpoly, 5.0, Tuple( p.coeffs ) )
```
gives the right result.

This might be due to the fact that the indexing for `p` starts at 0 and `_evalpoly_intermediate(x,p)`, where `p` is not a `Tuple`, assumes indexing starts at 1.
```julia
p_tup = Tuple(p.coeffs)
p[0] == p_tup[1] # true
p[1] == p_tup[2] # true
# and so forth
```

Issue #191, opened 2020-05-17
Representing (co)tangents of structured matrices


sethaxen:
This follows up a discussion on Slack.
@sethaxen:
> The pushforward of a function that produces a `Symmetric` matrix should also produce a `Symmetric` matrix.
> Is it also true that the pullback of a function that takes a `Symmetric` matrix should produce a `Symmetric` matrix?

@ChrisRackauckas:
> yes

@sethaxen:
>Okay, then what about when the input is `Symmetric{<:Real}`, but the pullback is passed an `AbstractMatrix{<:Complex}`? Should the pullback then produce a `Hermitian{<:Complex}` or a `Symmetric{<:Real}`, or something else?

@willtebbutt:
> I’ve been wondering about this.
> My old answer would have been (assuming that the symmetric matrix looks at the upper triangular of the underlying matrix): add the lower triangle to the upper triangle (don’t touch the diagonal), and represent the cotangent via the upper triangle, the rationale being that the lower triangle of the data backing the symmetric matrix never gets used.
> My newer thinking is that passing an asymmetric cotangent (I’m unclear whether we want represent this literally as a Symmetric or a `Composite{<:Symmetric}` ) to the pullback (e.g. a plain old `AbstractMatrix` ) should just error. My reasoning for this is that if you’ve somehow managed to get an asymmetric cotangent, then the downstream operations have been implemented incorrectly (e.g. `getindex(::Symmetric, ::Int, ::Int)` must be wrong and treating a `Symmetric` matrix as a general `Matrix` or something). Nothing in the Julia ecosystem currently implements this AFAICT though.

[...]

PR #193, opened 2020-05-17
Symmetric/Hermitian matrix function rules

sethaxen:
This PR adapts Zygote's adjoint rules for ~power series functions (analytic functions)~ matrix functions of real `Symmetric` and `Hermitian` inputs, originally introduced in https://github.com/FluxML/Zygote.jl/pull/355 (derivation [here](https://gist.github.com/sethaxen/000d164e515014fdda70601be1ecfb56))
It additionally adds `frule`s for the same functions. `Hermitian` below also includes `Symmetric{<:Real}`.

A few notes (`Hermitian` below also includes `Symmetric{<:Real}`):
- ~It also adds rules for `eigen` and `eigvals`, since they share a number of utility functions with the power series rules. The `eigen` rule should probably be a fallback for #179.~ Moved to #323 
- The rules assume that the (co)tangents are of the same type as their corresponding primal. It also assumes they are `Hermitian`, etc, not `Composite{<:Hermitian}` (relates #191).
- For the moment, it only adapts Zygote's rules for the type-stable functions, i.e. those whose scalar codomain is always real when their domain is real. (relates https://github.com/JuliaLang/julia/issues/35667).
- The above combined with `Hermitian` matrices having real eigenvalues means that the `rrule`s are not dependent on the conjugation convention of the cotangent.

To-do:
- [x] Write tests (this will include resolving https://github.com/FluxML/Zygote.jl/issues/608, since it relates these same rules). [This gist](https://gist.github.com/sethaxen/fa67e541c4a2a5e773b475349ed87fb9) tests most of the cases for Zygote's versions.

Because only a few utility functions differ between the `Hermitian{<:Complex}` case and the `Symmetric{<:Real}` and `Hermitian{<:Real}` cases, I don't think not having FD v0.10.0 support to test the rules with complex inputs should hold up this PR.


sethaxen:
These rules will probably not be usable with Zygote without some modification even after https://github.com/FluxML/Zygote.jl/pull/366 is merged, due to Zygote not enforcing type constraints. e.g.

```julia
julia> using Zygote, ChainRules

julia> A = Symmetric(randn(3, 3));

julia> Zygote.pullback(sum ∘ exp, A)[2](1.0)[1]
3×3 Array{Float64,2}:
 16.1831    4.00247     2.91251
  4.00247  -0.0429689  -0.331864
  2.91251  -0.331864   -0.546545

julia> rrule(exp, A)[2](ones(size(A)))[2] |> unthunk # this is identical, and hey, we have type constraints!
3×3 Symmetric{Float64,Array{Float64,2}}:
 16.1831    4.00247     2.91251
  4.00247  -0.0429689  -0.331864
  2.91251  -0.331864   -0.546545

julia> Zygote.pullback(sum ∘ exp, A)[2](im)[1]
3×3 Array{Complex{Float64},2}:
 0.0+16.1831im  0.0+4.00247im    0.0+2.91251im
 0.0+4.00247im  0.0-0.0429689im  0.0-0.331864im
 0.0+2.91251im  0.0-0.331864im   0.0-0.546545im

julia> rrule(exp, A)[2](im*ones(size(A)))[2] |> unthunk # that's...not even close
3×3 Symmetric{Complex{Float64},Array{Complex{Float64},2}}:
 0.0-6.6388im   0.0+5.58432im   0.0+5.86321im
 0.0+5.58432im  0.0-0.608647im  0.0-0.610593im
 0.0+5.86321im  0.0-0.610593im  0.0-1.26019im
```

I'm not certain if there's an easy fix to make the two compatible.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -47,6 +47,225 @@ end
...
+_realifydiag!(A::LinearAlgebra.RealHermSym) = A
+
+_realifydiag(A) = A - _pureimag(Diagonal(A))
+_realifydiag(A::LinearAlgebra.RealHermSym) = A
+
+# constrain B to have same Symmetric/Hermitian type as A
+_symherm(A::Symmetric, B) = Symmetric(B, Symbol(A.uplo))
+_symherm(A::Hermitian, B) = Hermitian(B, Symbol(A.uplo))
+_symherm(f, B::AbstractMatrix{<:Real}, uplo) = Symmetric(B, uplo)
+_symherm(f, B::AbstractMatrix{<:Complex}, uplo) = Hermitian(B, uplo)
```
nickrobinson251, line 128: Few thoughts 
- I wonder if this 3-arg version (where the first arg is not used) would be clearer as just another 2-arg version
`_symherm(B::AbstractMatrix{<:Complex}, uplo::Union{Char, Symbol})`
- I wonder if the mutating version should be a different function (since it does a subtly different thing, rather than the same thing but in-place)
- If we have `_symherm!`, then we want the mutated arg to be first, i.e. `_symherm!(B, A)` since it's `B` that (potentially) get's mutated. In which case we'd want to change the order for all methods, to be e.g. 
  ```julia
  # constrain B to have same Symmetric/Hermitian type as A
  _symherm(B::Symmetric, A) = Symmetric(B, Symbol(A.uplo))
  ```
- I wonder if we only actually need the `uplo` version? and then just call `_symherm(B, A.uplo)` instead of `_symherm(B, A)`? Or is getting `B::Hermition` if `A isa Hermitian` important (even if `eltype(B) <: Real`)?

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -47,6 +47,225 @@ end
...
+            end
+            return _symherm(A, U * U′∂AU * U')
+        end
+        return NO_FIELDS, ∂A
+    end
+    return sincosA, sincos_pullback
+end
+
+# difference quotient, i.e. Pᵢⱼ = (f(λᵢ) - f(λⱼ)) / (λᵢ - λⱼ), with f'(λᵢ) when i==j
+Base.@propagate_inbounds function _diffquot(f, i, j, λ, fλ, df_dλ)
```
nickrobinson251, line 253: is `f` needed here (since we get `fλ` and `df_dλ`)?

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -47,6 +47,225 @@ end
...
+        ∂A = Thunk() do
+            ∂F = Composite{typeof(F)}(values = Δλ)
+            return unthunk(back(∂F)[2])
+        end
+        return NO_FIELDS, ∂A
+    end
+    return λ, eigvals_pullback
+end
+
+_nonzero(x) = abs(x) > eps(eltype(x)) ? x : eps(eltype(x)) * sign(x)
```
nickrobinson251, line 105: can you add a comment about this function? right now i'm not sure if it has the expected behaviour or not, 
since it won't make `0.0` non-zero (because `sign(0.0) === 0.0`)

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -47,6 +47,225 @@ end
 _symmetric_back(ΔΩ) = UpperTriangular(ΔΩ) + LowerTriangular(ΔΩ)' - Diagonal(ΔΩ)
 _symmetric_back(ΔΩ::Union{Diagonal,UpperTriangular}) = ΔΩ
 
+#####
+##### `Symmetric{<:Real}`/`Hermitian` eigendecomposition
```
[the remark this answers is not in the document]
sethaxen: In base, all of the primal methods are defined in `LinearAlgebra/src/symmetric.jl`. Would it be better to mirror that by moving these (and in another PR the pullbacks for `Symmetric`) into a new file, `LinearAlgebra/src/symmetric.jl`?

src/rulesets/LinearAlgebra/structured.jl
sethaxen: These are great points, thanks! Yeah, this gets to the funniness of base. For one of these functions `f` (e.g. `exp`), then we have `y = f(x::Hermitian{<:Real})::Symmetric{<:Real}`, and `y.uplo` is always `'U'`, even if `x.uplo` is `'L'`,
The exception is the (not yet implemented) type-unstable functions, for which we have a different set of rules, hence `f` as an argument for future dispatch:
```julia
_symherm(f, B::AbstractMatrix{<:Real}, uplo) = Symmetric(B, uplo)
_symherm(f, B::AbstractMatrix{<:Complex}, uplo) = Hermitian(B, uplo)
```
 However, we want outputs of pushforwards and pullbacks to exactly match the types and have the same `uplo` as their corresponding primals, so we also have
```julia
_symherm(A::Symmetric, B) = Symmetric(B, Symbol(A.uplo))
_symherm(A::Hermitian, B) = Hermitian(B, Symbol(A.uplo))
```
And lastly, for some reason the functions in base zero the imaginary component of the `.data` field in-place, hence `_symherm!`.

The point is well-taken though, and I can both clean up the signatures and document some of this logic.

src/rulesets/LinearAlgebra/structured.jl
sethaxen: No, not anymore.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -47,6 +47,225 @@ end
...
+        _setdiag!(K, Zero())
+        ∂U = U * K
+        return Composite{typeof(F)}(values = ∂λ, vectors = ∂U)
+    end
+    return F, ∂F
+end
+
+function rrule(::typeof(eigen), A::LinearAlgebra.RealHermSymComplexHerm)
+    F = eigen(A)
+    function eigen_pullback(ΔF)
```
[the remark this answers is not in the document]
sethaxen: We could do this, but I think it's best to avoid typing for reasons other than dispatch. Unless this is necessary to make sure something else works correctly?

src/rulesets/LinearAlgebra/structured.jl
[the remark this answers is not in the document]
willtebbutt: That would be ideal.

sethaxen:
I'd like to polish this off here to ultimately replace Zygote's version. The main decision that blocked this was the age-old debate of whether covectors should be projected before being pulled back (see https://github.com/JuliaDiff/ChainRules.jl/pull/193#issuecomment-629775530). i.e. if one of these rules takes real inputs and produces a real output, but someone tries to pull back a complex `Hermitian` matrix, then what should we do? I am now in the "discard the imaginary part before pulling back" camp. And I also think all covectors should be projected to be Hermitian or Symmetric depending on the structure of the output of the primal before pulling back. This differs from how Zygote's versions of these rules behaves and needs to be decided before proceeding.

willtebbutt:
FWIW I'm also in your camp on this @sethaxen .

sethaxen:
Okay, this PR is ready for review now. I removed the rule for `^`. That should be handled in its own PR, as handling it correctly will add a bit of code (requiring implementing a pullback of `Base.power_by_squaring`)

Code coverage says three methods aren't covered, but that's a miscount, as they are covered via integration tests. Probably being inlined out or something.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+            return Y, $(Symbol("$(func)_pullback"))
+        end
+    end
+end
+
+function frule((_, ΔA), ::typeof(sincos), A::LinearAlgebra.RealHermSymComplexHerm)
+    ΔA isa AbstractZero && return sincos(A), ΔA
+    sinA, (λ, U, sinλ, cosλ) = _matfun(sin, A)
+    cosA = _symhermtype(sinA)((U * Diagonal(cosλ)) * U')
+    tmp = ΔA * U
```
oxinabox, line 335:
```suggestion
    tmp = ΔA * U  # We will overwrite this matrix several times to hold different values
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+end
+
+function rrule(::typeof(sincos), A::LinearAlgebra.RealHermSymComplexHerm)
+    sinA, (λ, U, sinλ, cosλ) = _matfun(sin, A)
+    cosA = typeof(sinA)((U * Diagonal(cosλ)) * U', sinA.uplo)
+    Y = (sinA, cosA)
+    sincos_pullback(ΔY::AbstractZero) = (NO_FIELDS, ΔY)
+    function sincos_pullback(ΔY::Composite)
+        ΔsinA, ΔcosA = ΔY
+        ΔsinA isa AbstractZero && ΔcosA isa AbstractZero && return NO_FIELDS, ΔsinA + ΔcosA
```
oxinabox, line 353:
```suggestion
    function sincos_pullback((ΔsinA, ΔcosA)::Composite)
        ΔsinA isa AbstractZero && ΔcosA isa AbstractZero && return NO_FIELDS, ΔsinA + ΔcosA
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+        if eltype(A) <: Real
+            ∂sinA, ∂cosA = real(ΔsinA), real(ΔcosA)
+        else
+            ∂sinA, ∂cosA = ΔsinA, ΔcosA
+        end
+        if ∂cosA isa Zero
```
oxinabox, line 359: Why not
```suggestion
        if ∂cosA isa AbstractZero
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+            Ā = _matfun_frechet(sin, A, sinA, ∂sinA, (λ, U, sinλ, cosλ))
+        elseif ∂sinA isa Zero
+            Ā = _matfun_frechet(cos, A, cosA, ∂cosA, (λ, U, cosλ, -sinλ))
+        else
+            tmp = ∂sinA * U
```
oxinabox, line 364:
```suggestion
            tmp = ∂sinA * U  # we will overwrite this with various temporary values during this computation
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+    return Y, sincos_pullback
+end
+
+# compute the matrix function f(A), returning also a cache of intermediates for computing
+# the pushforward or pullback.
+function _matfun(f, A::LinearAlgebra.RealHermSymComplexHerm)
+    λ, U = eigen(A)
+    if all(λi -> _isindomain(f, λi), λ)
+        fλ_df_dλ = map(λi -> frule((Zero(), One()), f, λi), λ)
+    else # promote to complex if necessary
```
oxinabox, line 384:
```suggestion
    else  # promote to complex if necessary
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+        end
+        _hermitrize!(Ā)
+        ∂A = typeof(A)(Ā, A.uplo)
+        return NO_FIELDS, ∂A
+    end
+    return Y, sincos_pullback
+end
+
+# compute the matrix function f(A), returning also a cache of intermediates for computing
+# the pushforward or pullback.
```
oxinabox, line 379:
```suggestion
# the pushforward or pullback.
# Note any function `f` used with this **must** have a `frule` defined on it.
```

What do you thing about making this a docstring?

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+    fA = (U * Diagonal(fλ)) * U'
+    Y = if eltype(A) <: Real
+        Symmetric(fA)
+    elseif eltype(fλ) <: Complex
+        fA
+    else
+        Hermitian(fA)
+    end
+    cache = (λ, U, fλ, df_dλ)
+    return Y, cache
```
oxinabox, line 398:
```suggestion
    return Y, (λ, U, fλ, df_dλ)
```
or

```suggestion
    intermediates = (λ, U, fλ, df_dλ)
    return Y, intermediates
```

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+    else
+        PΔ .= β .* PΔ .+ _diffquot.(f, λ, λ', fλ, transpose(fλ), ∂fλ, transpose(∂fλ)) .* Δ
+        return PΔ
+    end
+end
+
+_isindomain(f, x) = true
+_isindomain(::Union{typeof(acos),typeof(asin)}, x::Real) = -1 ≤ x ≤ 1
+_isindomain(::typeof(acosh), x::Real) = x ≥ 1
+_isindomain(::Union{typeof(log),typeof(sqrt)}, x::Real) = x ≥ 0
```
oxinabox, line 446: I really want a package that just knows about this kind of thing
`SpecialFunctionProperties.jl`

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -288,6 +454,18 @@ _symhermtype(::Type{<:Symmetric}) = Symmetric
 _symhermtype(::Type{<:Hermitian}) = Hermitian
 _symhermtype(A) = _symhermtype(typeof(A))
 
+function _realifydiag!(A)
+    for i in axes(A, 1)
```
oxinabox, line 458: Why this and not `diaginds` ?
Is faster? or clearer?

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -283,4 +283,225 @@
...
     end
+
+    @testset "Symmetric/Hermitian matrix functions" begin
+        # generate random matrices of type TA in the domain of f
+        function rand_matfun_input(f, TA, T, uplo, n, hermout)
+            U = Matrix(qr(randn(T, n, n)).Q)
+            if hermout # f(A) will also be a TA
+                λ = if f in (acos, asin, atanh)
+                    2 .* rand(real(T), n) .- 1
+                elseif f in (log,sqrt)
```
oxinabox, line 294:
```suggestion
                elseif f in (log, sqrt)
```

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -283,4 +283,225 @@
...
+            @testset "frule" begin
+                @testset for uplo in (:L, :U), hermout in (true, false)
+                    A, ΔA = rand_matfun_input(f, TA, T, uplo, n, hermout), TA(randn(T, n, n), uplo)
+                    Y = f(A)
+                    istypestable = try
+                        @inferred f(A)
+                        true
+                    catch ErrorException
+                        false
+                    end
```
oxinabox, line 327: can we extract this out into a helper function?

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -283,4 +283,225 @@
...
+            @testset "rrule" begin
+                @testset for uplo in (:L, :U), hermout in (true, false)
+                    A = rand_matfun_input(f, TA, T, uplo, n, hermout)
+                    Y = f(A)
+                    istypestable = try
+                        @inferred f(A)
+                        true
+                    catch ErrorException
+                        false
+                    end
```
oxinabox, line 383: Can we extract this into a helper function?

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -283,4 +283,225 @@
...
+                        TY = if T <: Real
+                            Union{Symmetric{Complex{T}},Symmetric{T}}
+                        else
+                            Union{Matrix{T},Hermitian{T}}
+                        end
+                        Y_ad, back = @inferred Tuple{TY,Any} rrule(f, A)
+                    end
+                    @test Y_ad == Y
+                    @test typeof(Y_ad) === typeof(Y)
+                    :uplo in propertynames(Y) && @test Y_ad.uplo == Y.uplo
```
oxinabox, line 401: Sigh `hasproperty` is not in 1.0.
Is it in Compat.jl?

oxinabox (commented):
Nice.
I am going to have to come back to this later to finish reviewing the tests properly.
and I am yet to check the source.
Should I get a person who knows math check it, or has it already been checked by someone who knows math when it went in Zygote?

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+# operators and applications to the theory of perturbations.
+# Amer. Math. Soc. Transl., Series 2, 47:1–30, 1965.
+# Stabilization for almost-degenerate matrices due to
+# S. D. Axen, 2020. Representing Ensembles of Molecules.
+# Appendix D: Automatic differentation rules for power series functions of diagonalizable matrices
+# https://escholarship.org/uc/item/6s62d8pw
+# These rules are more stable for degenerate matrices than applying the chain rule to the
+# rules for `eigen`.
+
+for func in (:exp, :log, :sqrt, :cos, :sin, :tan, :cosh, :sinh, :tanh, :acos, :asin, :atan, :acosh, :asinh, :atanh)
```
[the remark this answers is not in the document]
sethaxen: `_matfun` calls the corresponding `frule`s for `sin` and `cos`, both of which internally call `sincos`, so we get that benefit for free.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+    @eval begin
+        function frule((_, ΔA), ::typeof($func), A::LinearAlgebra.RealHermSymComplexHerm)
+            ΔA isa AbstractZero && return $func(A), ΔA
+            Y, cache = _matfun($func, A)
+            Ȳ = _matfun_frechet($func, A, Y, ΔA, cache)
+            # If ΔA was hermitian, then ∂Y has the same structure as Y
+            ∂Y = if ishermitian(ΔA) && (isa(Y, Symmetric) || isa(Y, Hermitian))
```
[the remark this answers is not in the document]
sethaxen: The matrix functions when called on a `LinearAlgebra.RealHermSymComplexHerm` that wraps a `StridedMatrix` will always return `Union{Symmetric,Hermitian,Matrix}`. If someone wraps a `Diagonal`, all of the matrix functions error at `eigen!`. The only way one could get a `Diagonal` would be to implement their own diagonal matrix type, so I think we're safe here.

src/rulesets/LinearAlgebra/symmetric.jl
sethaxen: very nice

src/rulesets/LinearAlgebra/symmetric.jl
sethaxen: Makes sense. I intend to overload these functions to support other matrix types in LinearAlgebra for future PRs.

src/rulesets/LinearAlgebra/symmetric.jl
sethaxen: It was in an earlier benchmark, but I just double-checked, and I'm not seeing that now. I'll switch to `diagind`, it's safer.

test/rulesets/LinearAlgebra/symmetric.jl
sethaxen: Yeah, I can do that. Actually, we should have this in ChainRulesTestUtils anyways, since it doesn't make sense to check if an `frule` or `rrule` is type-stable if its primal isn't (its pullback can still be though).

test/rulesets/LinearAlgebra/symmetric.jl
sethaxen: It does, although for some reason using `hasproperty` still caused an error. I'll try again.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+        end
+
+        function rrule(::typeof($func), A::LinearAlgebra.RealHermSymComplexHerm)
+            Y, cache = _matfun($func, A)
+            $(Symbol("$(func)_pullback"))(ΔY::AbstractZero) = (NO_FIELDS, ΔY)
+            function $(Symbol("$(func)_pullback"))(ΔY)
+                # for Hermitian Y, we don't need to realify the diagonal of ΔY, since the
+                # effect is the same as applying _hermitrize! at the end
+                ∂Y = eltype(Y) <: Real ? real(ΔY) : ΔY
+                # for matrix functions, the pullback is related to the pushforward by an adjoint
```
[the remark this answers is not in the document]
sethaxen: No, a different reason. I don't have a name for the property, but is a property of any function that can be written as a converging power series with real coefficients. I haven't posted the proof anywhere or seen it before, but I'm sure it could be argued from some generic property. For Hermitian matrices, it can be shown by applying the usual inner product trick in the ChainRules docs to derive the pullback.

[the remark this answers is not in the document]
sethaxen: Perhaps @antoine-levitt knows of this property.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+                Ā = _matfun_frechet($func, A, Y, ∂Y', cache)
+                # the cotangent of Hermitian A should be Hermitian
+                ∂A = typeof(A)(eltype(A) <: Real ? real(Ā) : Ā, A.uplo)
+                _hermitrize!(∂A.data)
```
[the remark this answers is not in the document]
sethaxen: Yeah, same thing. I can do that.

sethaxen:
> ...and I am yet to check the source.
> Should I get a person who knows math check it, or has it already been checked by someone who knows math when it went in Zygote?

Unfortunately the Daleckii-Krein paper is not available on the internet. I've cited it because everywhere the Frechet derivative appears in this form they cite that paper, and I'm waiting on my library to send me a copy of it to confirm. These rules are a fairly straightforward application of the chain rule and the usual rules for hermitian `eigen` to the primal functions in LinearAlgebra. The only "fancy" thing that this and the Zygote version introduces is the stabilization of the difference quotient using the Taylor series. But this also is not new; @antoine-levitt noted at the time that he uses the same trick (note, I use a slightly different bound here than in Zygote). That error bound could probably be tightened to be more robust for matrices with very large or small eigenvalues, but I think that could be a future PR.

src/rulesets/LinearAlgebra/symmetric.jl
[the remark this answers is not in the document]
antoine-levitt: sorry, what property are you talking about?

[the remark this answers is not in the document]
sethaxen: > sorry, what property are you talking about?

For a matrix function `f` that can be written in terms of a converging power series with real coefficients,
its pullback `f^*` is related to its pushforward `f_*` by an adjoint. Specifically, if `Y=f(A)`, `ΔA` is a tangent of `A`, and `ΔY` is a cotangent of `Y` (adopting ChainRules'/Zygote's conventions for how a cotangent is represented), then `(f^*)_{Y} (ΔY) = (f_*)_{A'} (ΔY)`. The property means we can write the pullback for any matrix function in terms of its `frule`.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -279,6 +279,172 @@ function rrule(::typeof(svdvals), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS
...
+    return ∂Y
+end
+
+# difference quotient, i.e. Pᵢⱼ = (f(λⱼ) - f(λᵢ)) / (λⱼ - λᵢ), with f'(λᵢ) when λᵢ=λⱼ
+function _diffquot(f, λi, λj, fλi, fλj, ∂fλi, ∂fλj)
+    T = Base.promote_typeof(λi, λj, fλi, fλj, ∂fλi, ∂fλj)
+    Δλ = λj - λi
+    iszero(Δλ) && return T(∂fλi)
+    # handle round-off error using Maclaurin series of (f(λᵢ + Δλ) - f(λᵢ)) / Δλ wrt Δλ
+    # and approximating f''(λᵢ) with forward difference (f'(λᵢ + Δλ) - f'(λᵢ)) / Δλ
```
antoine-levitt, line 423: oh that's more clever than what I do and gets you eps^2/3 (I only get eps^1/2), nice! Of course this is all assuming that all quantities are order 1 (or else you need to be careful about relative errors rather than absolute ones), but it's fine.

src/rulesets/LinearAlgebra/symmetric.jl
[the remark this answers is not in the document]
antoine-levitt: I'd have to check but isn't it basically f(A') = f(A)'?

antoine-levitt:
Regarding reference, I'd just cite Higham, pretty sure he discusses DK theorem somewhere.

src/rulesets/LinearAlgebra/symmetric.jl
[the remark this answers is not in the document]
sethaxen: I got there through a much more complicated route (writing the power series as a recurrence and working out the corresponding pushforwards and pullbacks as recurrences yields this relation), but I'll check to see if we can get there from that property as well. That would be nice, haha.

src/rulesets/LinearAlgebra/symmetric.jl
sethaxen: That's right, yeah. TBH, error analysis is not my thing. I plan to come back and try to improve this in the future. For the moment, this seems to be fine. I haven't been able to construct a random almost-degenerate matrix for which the pushforward/pullback disagrees with finite differences so far.

antoine-levitt: well it'll agree to eps^2/3 if what you wrote is correct. I can check the math if you want

sethaxen: That'd be great if you have the time!

src/rulesets/LinearAlgebra/symmetric.jl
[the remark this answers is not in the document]
antoine-levitt: hm no it looks more complicated than that, it's df(A)' = df(A') or something like that, I'm not sure it follows from f(A)' = f(A') (but maybe it does, haven't checked carefully). I use this trick in the hermitian case because it means that the differential is self-adjoint on the space of hermitian matrices equipped with the Frobenius metric, but I didn't know about it in the non-hermitian case, it's cute.

src/rulesets/LinearAlgebra/symmetric.jl
antoine-levitt: Yup, checks out. The approximation (f(l1) - f(l2))/(l1-l2) ~= (f'(l1) + f'(l2))/2 is accurate to order dl^2. The roundoff error when computing (f(l1) - f(l2))/(l1-l2) is order eps/dl. So it's advantageous to switch to the approximation when eps/dl ~= dl^2 => dl = cbrt(eps). The worst case accuracy is indeed eps^2/3.

sethaxen: Awesome! Thanks for checking that!

test/rulesets/LinearAlgebra/symmetric.jl
sethaxen: I added the helper function to https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/103 and will use it in this PR.

sethaxen:
I made a few changes to ensure that the rules do the right thing when the wrapped array is immutable. This isn't tested though.

src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -347,28 +344,25 @@ function rrule(::typeof(sincos), A::LinearAlgebra.RealHermSymComplexHerm)
...
-            ∂sinA, ∂cosA = ΔsinA, ΔcosA
+            ΔsinA, ΔcosA = real(ΔsinA), real(ΔcosA)
         end
-        if ∂cosA isa AbstractZero
-            Ā = _matfun_frechet(sin, A, sinA, ∂sinA, (λ, U, sinλ, cosλ))
-        elseif ∂sinA isa Zero
-            Ā = _matfun_frechet(cos, A, cosA, ∂cosA, (λ, U, cosλ, -sinλ))
+        if ΔcosA isa AbstractZero
+            Ā = _matfun_frechet(sin, A, sinA, ΔsinA, (λ, U, sinλ, cosλ))
+        elseif ΔsinA isa Zero
```
oxinabox, line 354: why is this one a `Zero` and not an `AbstractZero` as above?

oxinabox (approved):

src/rulesets/LinearAlgebra/symmetric.jl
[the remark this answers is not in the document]
sethaxen: Seems like it would be worth me writing this up somewhere.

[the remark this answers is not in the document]
antoine-levitt: OK, so let's do the proof for A^n, n ≥ 1. Then we can pass to arbitrary functions by linearity.

df(A)⋅dA = ∑_{k=0^n-1} A^k dA A^(n-1-k), so let's look at the adjoint of the linear operator L(A) : dA -> A^k1 dA A^k2, and the result follows again by linearity
<L(A) dA, dB> = tr((L(A) dA)' dB) = tr(A^k2' dA' A^k1' dB) = tr(dA' A^k1' dB A^k2') so adj(L(A)) : dB -> A^k1' dB A^k2' so adj(L(A)) = L(A')

That's probably close to the derivation you had?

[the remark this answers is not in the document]
sethaxen: Okay yes, but yours is _much_ simpler and more concise than mine. Nice!

sethaxen:
Blocked by https://github.com/JuliaDiff/ChainRulesCore.jl/pull/279 and need to stabilize output type for `jvp`.

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -329,13 +329,7 @@
...
                     A, ΔA = rand_matfun_input(f, TA, T, uplo, n, hermout), TA(randn(T, n, n), uplo)
                     Y = f(A)
-                    istypestable = try
-                        @inferred f(A)
-                        true
-                    catch ErrorException
-                        false
-                    end
-                    if istypestable
+                    if ChainRulesTestUtils._is_inferrable(f, A)
```
oxinabox, line 332: I don't love that we are using an explictly nonexported function.
Do we even need to be testing `@inferred` here, or can we rely on `frule_test` to do that for us later

sethaxen: Unfortunately we cannot use `frule_test` or `rrule_test` here because we need to work around the outputs of some functions being type-unstable. (and `rrule`'s won't agree with FiniteDifferences for `Symmetric`/`Hermitian`). I copied over a simplified `is_inferrable`.

Issue #195, opened 2020-05-20
Incorrect chain rule for `abs2`


MasonProtter:
this rule is incorrect for complex arguments: https://github.com/JuliaDiff/ChainRules.jl/blob/2d8b2656c16a38df1a01d698a0bd729d4624ac76/src/rulesets/Base/fastmath_able.jl#L40

It should be restricted to `<:Real` and then preferably have a non-holomorphic complex version defined.

oxinabox:
Cool want to make a PR?

PR #196, opened 2020-05-20
Correct chainrules for abs2, abs, conj and angle

MasonProtter:
Closes https://github.com/JuliaDiff/ChainRules.jl/issues/195. 

Pending some more thoughts in https://discourse.julialang.org/t/taking-complex-autodiff-seriously-in-chainrules/39317/49 and / or an issue here, we should consider adding something along the lines of 
```
@scalar_rule abs2(z::Complex) (z', z)
```

______

Current state of the PR is described here: https://github.com/JuliaDiff/ChainRules.jl/pull/196#issuecomment-649090280


MasonProtter:
Seth Axen pointed out that the Zygote chainrules PR does this: https://github.com/FluxML/Zygote.jl/blob/bf913a2a8ed616242e2f5378fbe598b289dd550a/src/lib/number.jl#L26-L30

to get correct answers. I think this is a reasonable way to go about it rather than using Wirtinger definitions.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,16 @@ let
 
         # Unary complex functions
         @scalar_rule abs(x::Real) sign(x)
-        @scalar_rule abs2(x) 2x
+        @scalar_rule abs2(x::Real) 2x
         @scalar_rule angle(x::Real) Zero()
         @scalar_rule conj(x::Real) One()
+        function frule((Δx,), abs, x::ComplexF64)
```
sethaxen, line 43: Can this be generalized for `Complex`? I see what I shared was too specific.
```suggestion
        function frule((Δx,), abs, x::Complex)
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,16 @@ let
...
+            Ω = abs(x)
+            return Ω, (real(x) * real(Δx) + imag(x) * imag(Δx)) / Ω
+        end
+        function frule((Δx,), abs2, x::ComplexF64)
```
sethaxen, line 47: Here as well?
```suggestion
        function frule((Δx,), abs2, x::Complex)
```

YingboMa:
Given ChainRules.jl does not support complex rules. Maybe it is better to remove or comment out complex rules.

willtebbutt:
> Given ChainRules.jl does not support complex rules. Maybe it is better to remove or comment out complex rules.

This is incorrect. We definitely support complex rules.

YingboMa:
> This is incorrect. We definitely support complex rules.

I don't see `Wirtinger` definition at all.

willtebbutt:
We just don't use Wirtingers. There's nothing stopping your from writing rules that work with complex numbers though.

YingboMa:
But are `abs` and `abs2` complex analytic?

willtebbutt:
> But are abs and abs2 complex analytic?

I don't believe so, but that's not the point. I'm not arguing that our `@scalar_rule` won't do the correct thing in the non-analytic case. It was a more general comment, not meant to be contraversial.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,16 @@ let
 
         # Unary complex functions
         @scalar_rule abs(x::Real) sign(x)
-        @scalar_rule abs2(x) 2x
+        @scalar_rule abs2(x::Real) 2x
         @scalar_rule angle(x::Real) Zero()
         @scalar_rule conj(x::Real) One()
+        function frule((Δx,), abs, x::Complex)
```
MasonProtter, line 43: @sethaxen  shouldn't this be

```julia
frule((_, Δx), abs, x::Complex)
```
?

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,16 @@ let
...
+            Ω = abs(x)
+            return Ω, (real(x) * real(Δx) + imag(x) * imag(Δx)) / Ω
+        end
+        function frule((Δx,), abs2, x::Complex)
```
MasonProtter, line 47: similarly, shouldn't this be 
```julia
 frule((_, Δx), abs2, x::Complex)
```

src/rulesets/Base/fastmath_able.jl
nickrobinson251: "needs tests"?

MasonProtter:
Does anyone know what the corresponding `rrule`s should be?

src/rulesets/Base/fastmath_able.jl
sethaxen: @MasonProtter Of course, yes, you're right. 🤦

src/rulesets/Base/fastmath_able.jl
sethaxen: Also 🤦

sethaxen:
> Does anyone know what the corresponding `rrule`s should be?

They should be the ones you linked to above in Zygote. Assuming Zygote's conjugation conventions of course.

YingboMa:
> I don't believe so, but that's not the point. I'm not arguing that our @scalar_rule won't do the correct thing in the non-analytic case. It was a more general comment, not meant to be contraversial.

I don't understand how can you define complex rules without using 2x2 matrix or Wirtinger, if the function is not complex analytic.

sethaxen:
> > I don't believe so, but that's not the point. I'm not arguing that our @scalar_rule won't do the correct thing in the non-analytic case. It was a more general comment, not meant to be contraversial.
> 
> I don't understand how can you define complex rules without using 2x2 matrix or Wirtinger, if the function is not complex analytic.

I believe all of these rules in ChainRules assume that the tangents and cotangents are derivatives of the primal with respect to a real scalar or a real scalar with respect to a primal, respectively. Or equivalently, which is why no Wirtinger is needed. If complex differentiation is what is needed, you just call the pushforward/pullback twice to fill the Jacobian. This is basically what Zygote does: https://fluxml.ai/Zygote.jl/latest/complex/

YingboMa:
I see. That makes sense. But then how can we warn the user that the function is not analytic? I don't think silently giving the wrong answer is a good idea.

MasonProtter:
> I see. That makes sense. But then how can we warn the user that the function is not analytic? I don't think silently giving the wrong answer is a good idea.

It's not the wrong answer. It's just that naively asking for `pullback(1)` doesn't allow you to derive the whole Jacobian, unlike in the holomorphic case.

It's like how in `ForwardDiff2` if you had `f(::Vector)::Vector` and did
```julia
D(f)([1, 0])
```
you wouldn't know the full Jacobian. You need `D(f)([0, 1])` as well. 

In this case, the whole Jacobian can be obtained from `pullback(1), pullback(im)`.

It would however be a good idea to make this thing more clear in the docs somehow.

YingboMa:
It's not a documentation issue. The information that the function is not holomorphic is never forwarded. So an AD system doesn't know it needs special handling for the non-holomorphic case.

sethaxen:
> It's not a documentation issue. The information that the function is not holomorphic is never forwarded. So an AD system doesn't know it needs special handling for the non-holomorphic case.

That's fair. In this approach, you need to check if the function is holomorphic by checking the Cauchy-Riemann equations. And it'll be a bit wasteful (although you can reuse the pullback when not mutating). But on the upside, the rules are simpler, and I don't think there's anything preventing future implementation of Wirtinger derivatives or something equivalent.

In case you didn't see it, this discussion is relevant: https://discourse.julialang.org/t/taking-complex-autodiff-seriously-in-chainrules

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,30 @@ let
...
         @scalar_rule angle(x::Real) Zero()
         @scalar_rule conj(x::Real) One()
+        function frule((_, Δz), ::typeof(abs), z::Complex)
+            Ω = abs(z)
+            return Ω, (real(z) * real(Δz) + imag(z) * imag(Δz)) / Ω
+        end
+        function rrule(::typeof(abs), z::Complex)
+            Ω = abs(z)
+            function ∂abs(Δz)
+                return real(Δz)*conj(z)/Ω
```
nickrobinson251, line 50: Need to also return the derivative with respect to the fields of function itself, which since it isn't a closure or callable struct is the `NO_FIELDS` constant (which is actually just `Zero`).

```suggestion
                return (NO_FIELDS, real(Δz)*conj(z)/Ω)
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -37,9 +37,30 @@ let
...
+            end
+            return Ω, ∂abs
+        end
+
+        function frule((_, Δz), ::typeof(abs2), z::Complex)
+            return abs2(z), 2 * (real(z) * real(Δz) + imag(z) * imag(Δz))
+        end
+        function rrule(::typeof(abs2), z::Complex)
+            function ∂abs2(Δz)
+                return real(Δz)*conj(z)
```
nickrobinson251, line 60:
```suggestion
                return (NO_FIELDS, real(Δz)*conj(z))
```

test/rulesets/Base/fastmath_able.jl
```diff
@@ -49,7 +49,7 @@ const FASTABLE_AST = quote
     end
 
     @testset "Unary complex functions" begin
-        for x in (-4.1, 6.4)
+        for x in (-4.1, 6.4, 3 + im)
```
nickrobinson251, line 52: this loop now tests we have `frule`s/`rrule`s for `angle` and `conj` on `Complex` numbers, which we don't yet, so we need to miss those out :)

MasonProtter:
Thanks for the helping hand, Nick

MasonProtter:
Looks like FiniteDifferences.jl doesn't do the right thing for complex inputs on, non-holomorphic functions. 
```julia
julia> fdm(central_fdm(5, 1), abs, 3 + im)
0.9486832980506245
```

willtebbutt:
Very good point. Alternatively we could (should?) restrict `fdm` to operate on the reals.

MasonProtter:
`fdm` knows how to handle Holomorphic functions correctly, so I think it might feel onerous to make that restriction:
```julia
julia> jacobian(central_fdm(5,1), ((x, y),) -> sin(x + im*y), [3.0, 1.0])
([-1.5276382501169548 0.16584440191898575; -0.16584440191912198 -1.5276382501165395],)

julia> fdm(central_fdm(5, 1), sin, 3 + im)
-1.527638250116479 - 0.165844401918999im
```
Not sure what the right way to handle this would be.

willtebbutt:
Hmm. I'm not sure. I guess I view `fdm` as the computational primitive in `FiniteDifferences`, for lack of a better description. It might make more sense to remove it from the user-facing API and instead insist that everyone uses `jacobian` or something. Either way, if someone could open an issue about this, that would be great.

ettersi:
> Looks like FiniteDifferences.jl doesn't do the right thing for complex inputs on non-holomorphic functions.
> 
> ```julia
> julia> fdm(central_fdm(5, 1), abs, 3 + im)
> 0.9486832980506245
> 
> julia> jacobian(central_fdm(5,1), ((x, y),) -> abs(x + im*y), [3.0, 1.0])
> ([0.9486832980506245 0.31622776601677055],)
> ```

The issue here is that defining finite differences in the complex plane is not as straightforward as it is on the real line. `fdm(..., f, x+im*y)` seems to return `df/dx` here, which I would argue is the most sensible out of all possible definitions (it is the straightforward generalisation of real finite differences, after all). The only cause of confusion then is that `df/dx` is not a quantity we usually want to compute, but I don't think it is the job of FiniteDifferences.jl to second-guess your intentions and then return something other than what you asked for. 

Eventually, this takes us back to the question of whether the above is an issue in documentation (`fdm` should clearly specify that it returns `df / dx` even for complex arguments), or whether `fdm` should simply not be defined for complex arguments. I am tending towards the latter here. Finite differences only make sense as directional derivatives, and since the complex plane is two-dimensional there is no canonical derivative direction here. 

However, I also believe that there are instances where we cannot reasonably protect the user from shooting their own feet. For example, I do not think that `frule` and `rrule` should check whether the function is holomorphic and issue a warning if not, since this would directly interfere with evaluating Jacobians as 
```
[
    real(back(1))     imag(back(1))
    real(back(im))    imag(back(im))
]
```
In these cases, all we can do is have a big red warning somewhere in the documentation stating that when you do complex differentiation, you should think carefully about what exactly you want to compute and whether the piece of code that you use actually does what you think it does.

ettersi:
The test errors on this branch are another instance of confusion regarding expectations of what a piece of code should do and what it actually does. The issue here is that
```
frule((nothing, 1.0), (x + im*y) -> u(x,y) + im*v(x,y), x + im*y)
```
returns `du / dx + im * dv / dx`, but 
```
rrule((x + im*y) -> u(x,y) + im*v(x,y), x + im*y) [2] (1.0)
```
returns `du / dx + im * du / dy`. 

Possible solutions are:
 - adapt `test_scalar` to handle the complex case properly, or
 - do not use `test_scalar` for non-holomorphic functions. 

I would be inclined to go with the first solution, but maybe someone more familiar with this code base might have different opinions.

willtebbutt:
I agree regarding `fdm` -- it's really set up for evaluating directional derivatives, and it's a "happy coincidence" that for holomorphic functions it returns the correct thing. The case with `frule`s and `rrule`s is different though: if you evaluate the `frule` or `rrule` of a non-holomorphic function, you should _always_ get the correct answer. As far as I'm concerned there's no discussion to be had about this. My suspicion is that if you try and implement an `frule` / `rrule` using the `@scalar_rule` macro for a non-holomorphic function, your implementation will just be wrong.

ettersi:
I agree that `frule` and `rrule` should work even for non-holomorphic functions, and I believe we agree that "correct" means to follow the "treat `Complex(x,y)` as `[x,y]`"-convention (please object if not). I further agree that I cannot think of a non-holomorphic function which can be implemented correctly using `@scalar_rule`, that Mason's PR is needed and implements the chain rules for `abs` and `abs2` correctly. What is needed to move forward is a fix for the tests of `abs` and `abs2` with complex arguments.

willtebbutt:
Agreed -- @MasonProtter the appropriate thing to do here is to use `frule_test` and `rrule_test` from `ChainRulesTestUtils` (also used elsewhere within `ChainRules`, rather than `test_scalar`. `test_scalar` has the same set of assumptions as `@scalar_rule`, so probably isn't appropriate here.

MasonProtter:
Sure thing, sorry I missed that standard.

MasonProtter:
Okay, I think this is approximately the right way to do the testing of complex functions (once it's refactored to meet guidelines and be more modular). 

Playing around with this has revealed that the way we handle holomorphic functions like `sin` is missing a `conj` in `rrule`:
```julia
using ChainRules, FiniteDifferences

function jacobian_via_frule(f,z)
    fz,df_dx = frule((Zero(), 1),f,z)
    fz,df_dy = frule((Zero(),im),f,z)
    return [
        real(df_dx)  real(df_dy)
        imag(df_dx)  imag(df_dy)
    ]
end
function jacobian_via_rrule(f,z)
    fz, pullback = rrule(f,z)
    _,du_dz = pullback( 1)
    _,dv_dz = pullback(im)
    return [
        real(du_dz)  imag(du_dz)
        real(dv_dz)  imag(dv_dz)
    ]
end
function jacobian_via_fdm(f, z::Union{Real, Complex})
    j = jacobian(central_fdm(5,1), ((x, y),) -> f(x + im*y), [(real ∘ float)(z)
                                                              (imag ∘ float)(z)])[1]
    if size(j) == (2,2)
        j
    elseif size(j) == (1, 2)
        [j 
         false false]
    else
        error("Invalid Jacobian size $(size(j))")
    end
end

let f = sin, z = 1 - 0.5im
    j1 = jacobian_via_frule(f, z)
    j2 = jacobian_via_rrule(f, z)
    j3 = jacobian_via_fdm(  f, z)
    @show j1
    @show j2 
    @show j3
    @show j1 ≈ j2 ≈ j3
end;

#+RESULTS:
: j1 = [0.6092589091577942 -0.4384865798925953; 0.4384865798925953 0.6092589091577942]
: j2 = [0.6092589091577942 0.4384865798925953; -0.4384865798925953 0.6092589091577942]
: j3 = [0.6092589091577078 -0.4384865798925346; 0.43848657989261214 0.6092589091577243]
: j1 ≈ j2 ≈ j3 = false
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,86 @@ let
 
 
         # Unary complex functions
-        @scalar_rule abs(x::Real) sign(x)
-        @scalar_rule abs2(x) 2x
-        @scalar_rule angle(x::Real) Zero()
-        @scalar_rule conj(x::Real) One()
+        ## abs
+        function frule((_, Δx), ::typeof(abs), x::Real)
```
sethaxen, line 40: Why are `frule`s and `rrule`s for `Real` inputs necessary? Doesn't the `@scalar_rule` macro work in those cases?

sethaxen:
> Playing around with this has revealed that the way we handle holomorphic functions like `sin` is missing a `conj` in `rrule` at least according to the conventions being set in [JuliaDiff/ChainRulesCore.jl#167](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/167)

That's correct. This will be fixed by https://github.com/JuliaDiff/ChainRulesCore.jl/pull/170.

> Okay, I think this is approximately the right way to do the testing of complex functions (once it's refactored to meet guidelines and be more modular).

With the perspective in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/167 the `frule`s and `rrule`s for all `ℂᵐ → ℂⁿ` are handled the same way as for the equivalent `ℝ²ᵐ → ℝ²ⁿ`. So I don't think it should be necessary to explicitly build the Jacobian in order to test these functions. One should be able to use the usual `ChainRulesTestUtils.frule_test` and `ChainRulesTestUtils.rrule_test` functions. Perhaps what is needed is for `ChainRulesTestUtils.test_scalar` to call those methods internally, just like `@scalar_rule` internally creates an `rrule` method and an `frule` method. This way `test_scalar` could be used for non-holomorphic complex functions. What do you think?

src/rulesets/Base/fastmath_able.jl
MasonProtter: Because `@scalar_rule` is still wrong for `x::Real` and `Δx::Complex`
```julia
using ChainRules
@scalar_rule abs(x::Real) sign(x)

jacobian_via_fdm(abs, 2.0)

#+RESULTS:
: 2×2 Array{Float64,2}:
:  1.0  -1.11022e-15
:  0.0   0.0
```
```julia
jacobian_via_frule(abs, 2.0)

#+RESULTS:
: 2×2 Array{Float64,2}:
:  1.0  0.0
:  0.0  1.0

```

sethaxen: Good point!

MasonProtter:
Yeah, that sounds good, I didn't know about the details of `ChainRulesTestUtils.jl`. I think there might be a problem with `jvp` though:
```julia
using FiniteDifferences
jvp(central_fdm(5,1), abs2, (3.0 + im, 0.25))

#+RESULTS:
: 1.9999999999995874
```
```julia
jvp(central_fdm(5,1), abs2, (3.0 + im, 0.25 + 0im))

#+RESULTS:
: 1.4999999999999445
```

sethaxen:
> I think there might be a problem with `jvp` though:

`jvp` and `j′vp` in general expect the (co)tangents to be of the same type as the primals. Usually they fail with an error if the inputs when converted to `ℝⁿ` have different lengths (at least `j′vp` does), but it looks like that isn't happening here. So your second one is doing the right thing. Your first example really should loudly fail though. Worth opening an issue on FD.

On further thought, it probably makes the most sense for `test_scalar` to parallel `@scalar_rule` as it currently does, so you could use `frule_test` and `rrule_test` as @willtebbutt suggested above.

MasonProtter:
> `jvp` and `j′vp` in general expect the (co)tangents to be of the same type as the primals. Usually they fail with an error if the inputs when converted to `ℝⁿ` have different lengths (at least `j′vp` does), but it looks like that isn't happening here. So your second one is doing the right thing. Your first example really should loudly fail though. Worth opening an issue on FD.

Hm, I'm not sure that's necessarily sound. For instance, if you are using units, i.e. `x::Unitful.Quantity` then `f(x)` won't have the same units as `df/dx`. 

But more generally, I think that in AD systems we'll somewhat generically end up in situations where we might try to use a `Δf` or `Δx` that has a different type than `f(x)` or `x` so we should be testing for those cases in ChainRules.jl

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,86 @@ let
...
+            return Ω, abs_pullback
+        end
+
+        ## abs2
+        function frule((_, Δx), ::typeof(abs2), x::Real)
+            return abs2(x), 2x * real(Δx)
+        end
+        function frule((_, Δz), ::typeof(abs2), z::Complex)
+            return abs2(z), 2 * (real(z) * real(Δz) + imag(z) * imag(Δz))
+        end
```
ettersi, line 68: How about merging these two methods to 
```suggestion
        function frule((_, Δz), ::typeof(abs2), z)
            return abs2(z), 2 * (real(z) * real(Δz) + imag(z) * imag(Δz))
        end
```
I would say we should only treat the real case separately if doing so allows for some performance optimisations (as it does e.g. for `abs` rules).

MasonProtter: For `Float64`, I think something funny happens that lets it do this in only one clock cycle:
```julia
let Δz = Ref(1.0 + im), x = Ref(4.0) 
    @btime f1($Δz[], $x[])
    @btime f2($Δz[], $x[])
end

#+RESULTS:
:   1.329 ns (0 allocations: 0 bytes)
:   1.329 ns (0 allocations: 0 bytes)
: 8.0
```
But generally, since the compiler can't know that `imag(Δz)` isn't `Inf` or `NaN`, I think the specialized rules for `Real` will generally be more performant. But yeah, it's also not a big deal, so if it's too much extra cruft we can remove it.

sethaxen:
> > `jvp` and `j′vp` in general expect the (co)tangents to be of the same type as the primals. Usually they fail with an error if the inputs when converted to `ℝⁿ` have different lengths (at least `j′vp` does), but it looks like that isn't happening here. So your second one is doing the right thing. Your first example really should loudly fail though. Worth opening an issue on FD.
> 
> Hm, I'm not sure that's necessarily sound. For instance, if you are using units, i.e. `x::Unitful.Quantity` then `f(x)` won't have the same units as `df/dx`.
> 
> But more generally, I think that in AD systems we'll somewhat generically end up in situations where we might try to use a `Δf` or `Δx` that has a different type than `f(x)` or `x` so we should be testing for those cases in ChainRules.jl. For instance, if you have something like `some_complicated_complex_function(x) + some_real_function(x)`, you might end up having a real `x` but complex `Δx`.
> 
> For instance, without testing such a case, it would have been hard for me to discover this problem: [#196 (comment)](https://github.com/JuliaDiff/ChainRules.jl/pull/196#discussion_r444566361)

Yeah, finite differences as a method is only well defined I think for real-to-real functions. In cases where the inputs/outputs can be mapped to real space, you can overload the `FiniteDifferences.to_vec` methods to do just that or compose the function you're testing with such maps. But this needs to be done for every non-real type, and its not a general solution. For cases like the one you're trying to test, the `rrule_test` and `frule_test` functions won't work, because you need to pass either a different function or different inputs to FDM than to ChainRules. e.g.

```julia
@test jvp(central_fdm(5,1), abs2, (3.0 + im, 0.25 + 0im)) ≈ frule((Zero(), 0.25), abs2, 3.0 + im)
```

In the Unitful case, you'd probably need to do something like this:

```julia
julia> dist(v, t) = v * t;

julia> jvp(central_fdm(5, 1), t -> dist(10u"m/s", t * u"s") / 1.0u"m", (1, 1))
10.00000000000002
```

There are a few cases like this already in ChainRules (see e.g. https://github.com/JuliaDiff/ChainRules.jl/blob/master/test/rulesets/LinearAlgebra/structured.jl#L51), where one needs to test the rule in a custom way, which I agree is what I think you'd need to do here. And I see that the Jacobian approach you have implemented is a convenient way to do that, so I'm fine with this.

src/rulesets/Base/fastmath_able.jl
ettersi: Ah, I see. Thanks for your explanation! I think we should keep your version, then.

ettersi:
> finite differences as a method is only well defined I think for real-to-real functions

`FiniteDifferences.jvp` is defined for any function `f : X -> Y` via 
```
lim_{h -> 0} ( f(x + Δx*h) - f(x) ) / h
```
as long as `X` and `Y` are vector spaces.

So I would say that 
```
jvp(central_fdm(5,1), abs2, (3.0 + im, 0.25))
```
should work, and similarly the Unitful example should work when changed to 
```
jvp(central_fdm(5, 1), t -> dist(10u"m/s", t), (1u"s", 1u"s"))
```
and the result should be 10u"m".

willtebbutt:
> Yeah, that sounds good, I didn't know about the details of ChainRulesTestUtils.jl. I think there might be a problem with jvp though:

I agree with Seth that this looks like a bug -- please do raise an issue about this, your example should error.

The current design is limited (as @sethaxen points out) in that it's not possible to "promote" a tangent vector of lower dimensionality to be the same dimensionality as the vector, so it's not possible to promote a real tangent to be complex if the input to the function is complex. It's not clear to me whether or not there's a good general solution to this, but I would personally be content with being sure that situations like the one you've encountered will error.

> FiniteDifferences.jvp is defined for any function f : X -> Y via

Yes. FiniteDifferences.jl is fine with this, and this is exactly the definition that's employed to compute `jvp`s. Whether or not it plays nicely with `Unitful`s out of the box is an implementation detail. The current design requires that any type can be mapped to a plain `Vector{<:Real}`, because we can always add them and scale them. The alternative approach would be to assume that any tangent vector can be scaled and added to the corresponding vector.

MasonProtter:
Okay, I opened an issue here: https://github.com/JuliaDiff/FiniteDifferences.jl/issues/90

MasonProtter:
Okay, I think this is now ready for review / merge. Here is a summary of the current changes and implementation. 

1) `@scalar_rule` is insufficient for non-holomorphic functions, even if the evaluation point is real, the sensitivity can be complex which gives an incorrect answer. Instead, `frule`s and `rrule`s need to be written according to the conventions and analysis in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/167 

I define rules for `abs`, `abs2`, `conj` and `angle` which are of the following form:
```julia
function frule((_, Δz), ::typeof($f), z) 
    Δx, Δy = reim(Δz)
    return $f(z), ∂u_∂x * Δx + ∂u_∂y * Δy + im * (∂v_∂x * Δx + ∂v_∂y * Δy)
end

function rrule(::typeof($f), z)
    function abs_pullback(Δf)
        Δu, Δv = real(Δf) 
        return (NO_FIELDS, Δu * ∂u_∂x + Δv * ∂v_∂x + im * (Δu * ∂u_∂y + Δv * ∂v_∂u))
    end
    return $f(z), abs_pullback
end
```
where I take `f(x + im * y) = u(x, y) + im * v(x, y)`. I also define real versions of these rules that in some cases will make their evaluation a bit quicker. These only exist for optimization purposes and if it's felt that they're unnecessary, I can remove them.

One unfortunate (but I think unavoidable) consequence of adopting these correct rules is:
```julia
julia> rrule(angle, 1.0)[2](1)
(Zero(), 0.0 + 1.0im)
``` 
because ` ∂/∂y angle(x + im*y) = ∂/∂y atan(x, y) = x/(x^2 + y^2)` which is `1/x` when `y=0`. This means that casting to complex can happen in reverse mode AD involving `angle` even if everything is real. This won't happen in forward mode though. 
```julia
julia> frule((0, 1), angle, 10.0)[2]
0.0
```
Arguably, this could just be said as a case where if you're upset about the performance hit of complex arithmetic, don't use `angle` on real numbers. :man_shrugging: 

2) `test_scalar` has similar limitations to `@scalar_rule`, so I define testing functions here: https://github.com/JuliaDiff/ChainRules.jl/pull/196/files#diff-d42146e529268fad091f175cb53fea59R2-R39. The idea here is that rather than use `frule_test` and `rrule_test`, I construct the Jacobian of the function via fdm and from the `rrule` and `frule` and compare the Jacobian directly. I do this because of some trickyness with FiniteDifferences.jl and because I want to test cases where the sensitivity is complex but the function evaluation point is real, which can be a very pernicious source of bugs.

If wanted, I can try to upstream and generalize these tests to ChainRulesTestUtils.jl, but I haven't had time to do that yet and may not for a little while.

sethaxen:
> One unfortunate (but I think unavoidable) consequence of adopting these correct rules is:
> 
> ```julia
> julia> rrule(angle, 1.0)[2](1)
> (Zero(), 0.0 + 1.0im)
> ```
> 
> because ` ∂/∂y angle(x + im*y) = ∂/∂y atan(x, y) = x/(x^2 + y^2)` which is `1/x` when `y=0`. This means that casting to complex can happen in reverse mode AD involving `angle` even if everything is real. This won't happen in forward mode though.
> 
> ```julia
> julia> frule((0, 1), angle, 10.0)[2]
> 0.0
> ```
> 
> Arguably, this could just be said as a case where if you're upset about the performance hit of complex arithmetic, don't use `angle` on real numbers. 🤷‍♂️

In general, the tangent will be the derivative of a primal wrt an upstream real. Likewise, the cotangent will in general be the derivative of a downstream real wrt a primal. Consequently, the (co)tangent of a real strictly speaking should always be a real.

This is easier to enforce in forward mode; one can 1) use a type constraint 2) check that each `frule` is passed a valid tangent and/or 3) just have the `frule` do the right thing assuming that the passed tangent is valid. ChainRules seems to do a mixture of (2) and (3). So e.g. in your case with the frule of `angle(::Real)`, you could assume that the tangent is a real if everything else in the AD system is working correctly, but you could also explicitly call `real` to ensure that it is. Of course, if someone defines a rule for an upstream function that does the wrong thing, then everything downstream could silently fail.

In reverse mode, it's harder, because while the primal function might be specialized for the input type, the `rrule` often is not, so the pullback might take a "wider" cotangent type than the primal output type and consequently return a wider cotangent for the primal input than the input primal type. I don't think we have a good way to handle this yet. We *sometimes* enforce a constraint on the pullback's output (see e.g. the structured matrix rules), but sometimes not. In this case, if you have a real primal and are passed a complex cotangent, it should be completely safe to call `real` on it. Which since in your case would always return a 0, you can just return a `Zero()`. My opinion is that this is the right thing to do here, because in general if all of your computations involve reals, then the pullback should not introduce complex numbers unless absolutely necessary. 

Someone please correct me if I am missing something here.

> 1. I do this because of some trickyness with FiniteDifferences.jl and because I want to test cases where the sensitivity is complex but the function evaluation point is real, which can be a very pernicious source of bugs.

FiniteDifferences v0.9 did not handle complex numbers correctly, but FD v0.10 does. Once we've bumped compat for v0.10.0, we should be able to use the usual testing functions (see e.g. https://github.com/JuliaDiff/ChainRules.jl/pull/211)

I'm about to review the code, but given that ChainRules is switching conjugation conventions with https://github.com/JuliaDiff/ChainRulesCore.jl/pull/167, and this adopts the conventions from that PR, my thinking is that this should be marked as a breaking change. And perhaps once merged we should delay bumping the version number and releasing until https://github.com/JuliaDiff/ChainRulesCore.jl/pull/170 is merged and ChainRules has bumped its compat accordingly. Thoughts @oxinabox @willtebbutt?

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
+            return abs(x), sign(x) * real(Δx)
+        end
+        function frule((_, Δz), ::typeof(abs), z::Complex)
+            Ω = abs(z)
+            return Ω, (real(z) * real(Δz) + imag(z) * imag(Δz)) / Ω
+        end
+        
+        function rrule(::typeof(abs), x::Real)
+            function abs_pullback(Δf)
+                return (NO_FIELDS, real(Δf)*sign(x))
```
sethaxen, line 50:
```suggestion
            function abs_pullback(ΔΩ)
                return (NO_FIELDS, real(ΔΩ)*sign(x))
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
+            end
+            return abs(x), abs_pullback
+        end
+        function rrule(::typeof(abs), z::Complex)
+            Ω = abs(z)
+            function abs_pullback(Δf)
+                Δu = real(Δf)
```
sethaxen, line 57:
```suggestion
            function abs_pullback(ΔΩ)
                Δu = real(ΔΩ) 
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
+            function abs_pullback(Δf)
+                return (NO_FIELDS, real(Δf)*sign(x))
+            end
+            return abs(x), abs_pullback
+        end
+        function rrule(::typeof(abs), z::Complex)
+            Ω = abs(z)
+            function abs_pullback(Δf)
+                Δu = real(Δf) 
+                return (NO_FIELDS, Δu*z/Ω)
```
sethaxen, line 58: Should we special-case when `z=0` here? e.g. for `sign`, the pushforward/pullback is technically undefined at `z=0`, but following the subgradient convention, they are given a value of `0`.

```suggestion
                return (NO_FIELDS, Δu*z/ifelse(iszero(z), one(Ω), Ω)
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
-        @scalar_rule abs2(x) 2x
-        @scalar_rule angle(x::Real) Zero()
-        @scalar_rule conj(x::Real) One()
+        ## abs
+        function frule((_, Δx), ::typeof(abs), x::Real)
+            return abs(x), sign(x) * real(Δx)
+        end
+        function frule((_, Δz), ::typeof(abs), z::Complex)
+            Ω = abs(z)
+            return Ω, (real(z) * real(Δz) + imag(z) * imag(Δz)) / Ω
```
sethaxen, line 45: Same question here as below regarding subgradient convention.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
+            function conj_pullback(Δf)
+                return (NO_FIELDS, conj(Δf))
+            end
+            return conj(z), conj_pullback
+        end
+
+        ## angle
+        function frule((_, Δz), ::typeof(angle), x::Real)
+            Δx, Δy = reim(Δz)
+            return angle(x), Δy/x
```
sethaxen, line 99: Same question here and below regarding subgradient convention.

src/rulesets/Base/fastmath_able.jl
MasonProtter: Hm. I'm not very familiar with this 'subgradient convention', but I must say it makes me uneasy. Personally, I think it'd rather just encounter a `NaN` than get zero gradient, but I can see how producing `NaN` willy-nilly in an AD system could be frustrating.

sethaxen: @oxinabox wrote a little bit about it here: https://github.com/JuliaDiff/ChainRules.jl/pull/185#discussion_r421675694.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,88 @@ let
...
+        function frule((_, Δz), ::typeof(angle), z::Complex)
+            x,  y  = reim(z)
+            Δx, Δy = reim(Δz)
+            return angle(z), (-y*Δx + x*Δy)/abs2(z)  
+        end
+        function rrule(::typeof(angle), x::Real)
+            function angle_pullback(Δf)
+                Δu, Δv = reim(Δf)
+                return (NO_FIELDS, im*Δu/x)
+            end
```
ettersi, line 110: I would say the correct definition is this:
```suggestion
            angle_pullback(Δf::Real) = (NO_FIELDS, Zero())
            function angle_pullback(Δf::Complex)
                Δu, Δv = reim(Δf)
                return (NO_FIELDS, im*Δu/x)
            end
```
Of course, one may argue what the result for `x = 0` should be, but that's orthogonal to the issue of how to handle `x::Real`, `Δf::Union{Real,Complex}`.

MasonProtter: If you do the math though, even for real `x` and real `Δf`, the pullback is still complex.

src/rulesets/Base/fastmath_able.jl
oxinabox: The short version of the subgradient convention is that we claim derivative is zero at local minima, regardless of if it is otherwise differentiable or not.
Because its useful to people doing optimization.
Its not something i feel that strongly about though.

For real numbers the common alternative in AD is to instead chose 1 or -1, which would be the program deriviative if going through control flow,
and the function had being implmented wioth a `if x <= 0` etc.
Which I think while acceptable is just worse.

For complex numbers, I don't know.
I am not sure the optimization usecase even applies.

I will not block this PR over this.

src/rulesets/Base/fastmath_able.jl
sethaxen: > If you do the math though, even for real `x` and real `Δf`, the pullback is still complex.

I don't think this conforms with the convention laid out in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/167. Namely, we treat complex numbers as just another way of representing `ℝ²`. So just as a `ℝ → ℝ` function's pullback shouldn't take an `ℝ²` covector and definitely shouldn't produce an `ℝ²` covector, that should be the same here. Doesn't your argument rely on treating `x::Real` as actually being embedded in the complex numbers? See also https://github.com/JuliaDiff/ChainRules.jl/issues/212.

Another case where this comes up is `sign` (see https://github.com/JuliaDiff/ChainRules.jl/pull/211). `sign` for a complex number normalizes it to the unit circle in the complex plane. Hence, at `z = 1 + 0im`, the cotangent will be parallel to the imaginary axis, i.e. `0 + Δy * im`. If we treat `sign(x::Real)` as embedded in the complex numbers, then our cotangent would look like this as well, when it should be `0`.

I wrote some thoughts on this here: https://github.com/JuliaDiff/ChainRules.jl/issues/212

src/rulesets/Base/fastmath_able.jl
sethaxen: Yeah, I found a few different conventions in ChainRules (e.g. `mod` uses `NaN` where it is not differentiable). So I also will not block over this.

MasonProtter: Optimization examples certainly do apply, especially in C -> R functins like `abs`. After some reflection, I think I can get behind using this subgradient convention here. I'll make the appropriate changes.

MasonProtter: @sethaxen Is there a reason you recommend `Δu*z/ifelse(iszero(z), one(Ω), Ω)` rather than 
```juia
ifelse(iszero(z), zero(Ω), Δu*z/Ω)
```
? I guess your version is better behaved under promotion?

src/rulesets/Base/fastmath_able.jl
MasonProtter: Yeah, I find that convincing, makes sense.

src/rulesets/Base/fastmath_able.jl
sethaxen: > I guess your version is better behaved under promotion?

Yeah that's right, it makes it type-stable.

MasonProtter:
Okay, so this PR now adopts the subgradient convention where in the situations that might cause functions like the gradient of `angle` `abs` to give `NaN` for non-nan inputs (i.e. `z = 0`), we instead return an appropriate zero.

It also has changed to the point of view for `angle` that it does *not* treat the reals as embedded in the complex plane, so `frule` and `rrule` of `angle` now give `Zero()` when `x::Real, Δx::Real, ΔΩ::Real`.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
-        @scalar_rule abs2(x) 2x
-        @scalar_rule angle(x::Real) Zero()
-        @scalar_rule conj(x::Real) One()
+        ## abs
+        function frule((_, Δx), ::typeof(abs), x::Real)
+            return abs(x), sign(x) * real(Δx)
+        end
+        function frule((_, Δz), ::typeof(abs), z::Complex)
+            Ω = abs(z)
+            return Ω, (real(z) * real(Δz) + imag(z) * imag(Δz)) / ifelse(iszero(z), one(Ω), Ω)
```
ettersi, line 45:
```suggestion
            return Ω, (real(z) * real(Δz) + imag(z) * imag(Δz)) / ifelse(iszero(z), one(Ω), Ω)
            # `ifelse` is applied only to denominator to ensure type-stability. 
```
Hope I am not putting the wrong words into your mouth.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            function abs_pullback(ΔΩ)
+                return (NO_FIELDS, real(ΔΩ)*sign(x))
+            end
+            return abs(x), abs_pullback
+        end
+        function rrule(::typeof(abs), z::Complex)
+            Ω = abs(z)
+            function abs_pullback(ΔΩ)
+                Δu = real(ΔΩ)
+                return (NO_FIELDS, Δu*z/ifelse(iszero(z), one(Ω), Ω))
```
ettersi, line 58:
```suggestion
                return (NO_FIELDS, Δu*z/ifelse(iszero(z), one(Ω), Ω))
                # `ifelse` is applied only to denominator to ensure type-stability. 
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            end
+            return abs2(x), abs2_pullback
+        end
+        function rrule(::typeof(abs2), z::Complex)
+            function abs2_pullback(ΔΩ)
+                Δu = real(ΔΩ)
+                return (NO_FIELDS, 2real(ΔΩ)*z)
+            end
+            return abs2(z), abs2_pullback
+        end
```
ettersi, line 83: As far as I can tell, this one is exactly the same as the `Real` version, and line 79 is unnecessary.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            function conj_pullback(ΔΩ)
+                return (NO_FIELDS, conj(ΔΩ))
+            end
+            return conj(z), conj_pullback
+        end
+
+        ## angle
+        function frule((_, Δz), ::typeof(angle), x::Real)
+            Δx, Δy = reim(Δz)
+            return angle(x), Δy/ifelse(iszero(x), one(x), x)
```
ettersi, line 99:
```suggestion
            return angle(x), Δy/ifelse(iszero(x), one(x), x) 
            # `ifelse` is applied only to denominator to ensure type-stability. 
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            Δx, Δy = reim(Δz)
+            return angle(x), Δy/ifelse(iszero(x), one(x), x) 
+        end
+        function frule((_, Δz)::Tuple{<:Any, <:Real}, ::typeof(angle), x::Real)
+            return angle(x), Zero()
+        end
+        function frule((_, Δz), ::typeof(angle), z::Complex)
+            x,  y  = reim(z)
+            Δx, Δy = reim(Δz)
+            return angle(z), (-y*Δx + x*Δy)/ifelse(iszero(z), one(z), abs2(z))
```
ettersi, line 107:
```suggestion
            return angle(z), (-y*Δx + x*Δy)/ifelse(iszero(z), one(z), abs2(z))  
            # `ifelse` is applied only to denominator to ensure type-stability. 
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            Δx, Δy = reim(Δz)
+            return angle(z), (-y*Δx + x*Δy)/ifelse(iszero(z), one(z), abs2(z))  
+        end
+        function rrule(::typeof(angle), x::Real)
+            function angle_pullback(ΔΩ::Real)
+                return (NO_FIELDS, Zero())
+            end
+            function angle_pullback(ΔΩ)
+                Δu, Δv = reim(ΔΩ)
+                return (NO_FIELDS, im*Δu/ifelse(iszero(x), one(x), x))
```
ettersi, line 115:
```suggestion
                return (NO_FIELDS, im*Δu/ifelse(iszero(x), one(x), x))
                # `ifelse` is applied only to denominator to ensure type-stability. 
```

src/rulesets/Base/fastmath_able.jl
```diff
@@ -36,10 +36,94 @@ let
...
+            end
+            return angle(x), angle_pullback 
+        end
+        function rrule(::typeof(angle), z::Complex)
+            function angle_pullback(ΔΩ)
+                x,  y  = reim(z)
+                Δu, Δv = reim(ΔΩ)
+                return (NO_FIELDS, (-y + im*x)*Δu/ifelse(iszero(z), one(z), abs2(z)))
```
ettersi, line 123:
```suggestion
                return (NO_FIELDS, (-y + im*x)*Δu/ifelse(iszero(z), one(z), abs2(z)))
                # `ifelse` is applied only to denominator to ensure type-stability. 
```

src/rulesets/Base/fastmath_able.jl
MasonProtter: Good catch!

PR #199, opened 2020-05-24
Rules for cross-product

sethaxen:
Adds rules for the cross product.
It includes a rule for chunked forward mode, following https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/37, where the chunked tangent vector matrices are assumed to have the size `(m, n)`, where `n` is the length of the input vectors and `m` is the number of chunks. Does this look right, @oxinabox @YingboMa?

Pending clarification on complex conjugation conventions ( https://github.com/JuliaDiff/ChainRulesCore.jl/issues/159) the `rrule` is restricted to real input vectors.

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -26,12 +26,6 @@ end
 function frule((_, Δa, Δb), ::typeof(cross), a::AbstractVector, b::AbstractVector)
     return cross(a, b), cross(Δa, b) .+ cross(a, Δb)
 end
-# chunked mode
-function frule((_, Δa, Δb)::Tuple{<:Any,<:AbstractMatrix,<:AbstractMatrix},
-               ::typeof(cross), a::AbstractVector, b::AbstractVector)
-    aₓ, bₓ = _crossmat(a), _crossmat(b)
-    return aₓ * b, Δa * bₓ .- Δb * aₓ
-end
 
 # TODO: support complex vectors
 function rrule(::typeof(cross), a::AbstractVector{<:Real}, b::AbstractVector{<:Real})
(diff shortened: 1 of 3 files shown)


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -19,6 +19,32 @@ function rrule(::typeof(dot), x, y)
...
 
+#####
+##### `cross`
+#####
+
+function frule((_, Δa, Δb), ::typeof(cross), a::AbstractVector, b::AbstractVector)
+    return cross(a, b), cross(Δa, b) .+ cross(a, Δb)
+end
+# chunked mode
+function frule((_, Δa, Δb)::Tuple{<:Any,<:AbstractMatrix,<:AbstractMatrix},
```
sethaxen, line 30: There aren't  (I think) any examples of dispatching on `frule` sensitivities yet. I don't think this is the right way to do it (doesn't support `Composite`).

willtebbutt: Yeah, I don't think that we've formalised the API for this properly yet. [This issue](https://github.com/JuliaDiff/ChainRulesCore.jl/issues/92) is still being addressed. It's gone a bit stale now, which I suspect is because the FD2 crew are okay with where it is currently.

YingboMa: I am not working on FD2 at the moment, but I do plan to rewrite it sometime.

sethaxen:
There were a few ways to go with this:
1) merge as is, assuming the chunked forward mode will follow the convention used here
2) wait until chunked API is decided and then adapt and merge
3) removed the chunked version from this PR for now and merge.

I'm thinking (3) is the right way to go.

PR #201, opened 2020-05-25
Add rrules for triu/tril

sethaxen:


sethaxen:
> Would be good to ensure that the cotangents are of the correct type.

In general for structured matrices our pullbacks don't ensure the right type (e.g. the `rrule` for `UpperTriangular` force the cotangent to be a `Matrix`, and the rrule for `diag` forces the cotangent to be `Diagonal`, but those types are only correct when the input is a `Matrix` or a `Diagonal`).

willtebbutt:
That's a fair point. We've really not got our act together on this front at the minute (do we implement things for `AbstractMatrix`s ever, or do we stick with much more strict sub-typing?), so I don't see what issue adding this one could cause for now.

PR #207, opened 2020-06-10
Remove constraints on primal from Composites for SVD and Cholesky

### LINKED ISSUE #205: SVD is broken in Zygote
e.g.
```julia
gradient(x -> sum(svd(x).S), rand(10,10))
```
fails due to Zygote's NamedTuples not being supported by the `rrule` for this function.

Would be nice to have a type alias for `Composite`/`NamedTuple` so that rrules can be generic across both. Perhaps for now we can just remove the type restriction, since no other adjoint for `SVD` is possible anyway.

oxinabox:
closes #205 
the constraint on getting the primal right is not actually needed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.6.3"
+version = "0.6.4"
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Can confirm this solves it, thanks for the quick fix!

PR #211, opened 2020-06-24
Fix rules for sign with complex input

sethaxen:
The current rule for `sign(z::Complex)` is incorrect, as `sign(z::Complex) = z / abs(z)`. This PR adds rules for the complex `sign`. It also reimplements the rules for real inputs, to ensure that they do the right thing if passed complex (co)tangents (following https://github.com/JuliaDiff/ChainRules.jl/pull/196#discussion_r444566361).

The tests in this PR require #198 to pass.

### DIFF
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -113,6 +113,8 @@ let
     # we do this second so it overwrites anything we included by mistake in the fastable
 end
 
+# the jacobian for `sign` is symmetric; `_sign_jvp` gives both J * Δz and Jᵀ * ΔΩ for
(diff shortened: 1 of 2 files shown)


src/rulesets/Base/fastmath_able.jl
```diff
@@ -88,3 +112,7 @@ let
     eval(fastable_ast)  # Get original definitions
     # we do this second so it overwrites anything we included by mistake in the fastable
 end
+
+_sign_jvp(Ω, absz, Δ) = Ω * ((imag(Δ) * real(Ω) - real(Δ) * imag(Ω)) / absz)im
```
willtebbutt, line 116: A brief comment here explaining why this function is here / what it does would be helpful.

PR #216, opened 2020-06-28
Add missing complex tests and rules

### LINKED ISSUE #215: Missing complex tests/rules for scalar functions
Scalar functions with general rules but no complex tests:
- [ ] `identity`
- [ ] `one`
- [ ] `zero`
- [ ] `deg2rad`
- [ ] `rad2deg`
- [x] `abs2` (handled by #196)
- [ ] `+`
- [ ] `-`
- [ ] `/`
- [ ] `\`
- [ ] `^`
- [ ] `inv`
- [ ] `muladd`
- [ ] `sqrt`
- [x] `sign` (handled by #211)
- [ ] `sincos`
- [ ] `exp2`
- [ ] `exp10`
- [ ] `expm1`
- [ ] `log`
- [ ] `log2`
- [ ] `log10`
- [ ] `log1p`

Scalar functions defined/tested on reals that can be generalized to/tested on complex:
- [ ] `real`
- [ ] `imag`
- [ ] `adjoint`
- [ ] `hypot`
- [ ] `exp`
- [x] `abs` (handled by #196)
- [x] `conj` (handled by #196)
- [x] `angle` (handled by #196)

sethaxen:
This PR fixes #215 by adding missing complex tests and where necessary generalizing rules and adding missing rules. New rules are
- `adjoint(::Number)`
- `real(::Number)`
- `imag(::Complex)`
- `Complex(::Number)`
- `Complex(::Number, ::Number)`
- `hypot(::Complex)`
- `hypot(::T, ::T) where T<:Union{Real,Complex}`
- `exp(::Number)`

It also simplifies the rules for `abs`, `angle`, and `sign` with shared utility functions.

I believe as of this PR, all scalar functions in Base for which we have complex rules should be tested on complex inputs.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -14,13 +14,26 @@ function rrule(::typeof(adjoint), z::Number)
     return (z', adjoint_pullback)
 end
 
+# `real`
+
+@scalar_rule real(x::Real) One()
+
+frule((_, Δz), ::typeof(real), z::Number) = (real(z), real(Δz))
+
+function rrule(::typeof(real), z::Number)
+    # add zero(z) to embed the real number in the same number type as z
+    real_pullback(ΔΩ) = (NO_FIELDS, real(ΔΩ) + zero(z))
+    return (real(z), real_pullback)
+end
+
+# `imag`
+
 @scalar_rule imag(x::Real) Zero()
 @scalar_rule hypot(x::Real) sign(x)
 
 
 @scalar_rule fma(x, y, z) (y, x, One())
 @scalar_rule muladd(x, y, z) (y, x, One())
-@scalar_rule real(x::Real) One()
 @scalar_rule rem2pi(x, r::RoundingMode) (One(), DoesNotExist())
 @scalar_rule(
     mod(x, y),
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -37,8 +37,23 @@ function rrule(::typeof(imag), z::Complex)
     return (imag(z), imag_pullback)
 end
 
+# `hypot`
+
 @scalar_rule hypot(x::Real) sign(x)
 
+function frule((_, Δz), ::typeof(hypot), z::Complex)
+    Ω = hypot(z)
+    ∂Ω = _realconjtimes(z, Δz) / ifelse(iszero(Ω), one(Ω), Ω)
+    return Ω, ∂Ω
+end
+
+function rrule(::typeof(hypot), z::Complex)
+    Ω = hypot(z)
+    function hypot_pullback(ΔΩ)
+        return (NO_FIELDS, (real(ΔΩ) / ifelse(iszero(Ω), one(Ω), Ω)) * z)
+    end
+    return (Ω, hypot_pullback)
+end
 
 @scalar_rule fma(x, y, z) (y, x, One())
 @scalar_rule muladd(x, y, z) (y, x, One())
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -3,8 +3,17 @@
 
 @scalar_rule one(x) zero(x)
 @scalar_rule zero(x) zero(x)
-@scalar_rule adjoint(x::Real) One()
 @scalar_rule transpose(x) One()
+
+# `adjoint`
+
+frule((_, Δz), ::typeof(adjoint), z::Number) = (z', Δz')
+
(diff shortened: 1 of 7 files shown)


src/rulesets/Base/base.jl
```diff
@@ -3,15 +3,80 @@
...
+end
+
+# `imag`
+
 @scalar_rule imag(x::Real) Zero()
+
+frule((_, Δz), ::typeof(imag), z::Number) = (imag(z), imag(Δz))
+
+function rrule(::typeof(imag), z::Complex)
+    imag_pullback(ΔΩ) = (NO_FIELDS, real(ΔΩ) * im)
```
sethaxen, line 36: I think we could generalize this to `z::Number` with this
```suggestion
function rrule(::typeof(imag), z::Number)
    # multiply by `one(z)` so for non-complex z, cotangent is similar to z
    imag_pullback(ΔΩ) = (NO_FIELDS, (real(ΔΩ) * im) * one(z))
```

sethaxen:
The one failure on Julia 1.4.2 is odd. `muladd(x,y,z)`'s rules are defined using `@scalar_rule`, which literally implements the primal for `frule` as `muladd(x,y,z)`, yet somehow the equality check against `muladd(x,y,z)` fails for complex inputs (the values are still approximately equal). https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/46 relaxes the equality check, but that still doesn't explain why it fails here. I cannot reproduce this locally.

src/rulesets/Base/base.jl
willtebbutt: I agree that this should be generalised to `Real`, it might be best to implement a separate method as `imag(x::Real)` always returns `zero(x)`, hence the `pullback` should also just return `zero(x)`.

willtebbutt:
Hmm yeah, that is very strange... checking for approximate equality is fine though imho.

src/rulesets/Base/base.jl
sethaxen: There already is a `@scalar_rule` implementation for `imag`. My idea here is about generalizing beyond `Complex` to `Number`. But I don't think this is useful. How often does one define a non-real, non-complex type that is a valid input to `imag`? There isn't even an `imag(::Number)` method.

willtebbutt: Oh I see. I wouldn't worry about this -- sticking with `Complex` type constraint is fine.

src/rulesets/Base/utils.jl
```diff
@@ -0,0 +1,13 @@
...
+@inline _realconjtimes(x::Real, y::Complex) = x * real(y)
+@inline _realconjtimes(x::Complex, y::Real) = real(x) * y
+@inline _realconjtimes(x::Real, y::Real) = x * y
+
+# imag(conj(x) * y) avoiding computing the real part if possible
+@inline _imagconjtimes(x, y) = imag(conj(x) * y)
+@inline _imagconjtimes(x::Complex, y::Complex) = -imag(x) * real(y) + real(x) * imag(y)
+@inline _imagconjtimes(x::Real, y::Complex) = x * imag(y)
+@inline _imagconjtimes(x::Complex, y::Real) = -imag(x) * y
+@inline _imagconjtimes(x::Real, y::Real) = Zero()
```
oxinabox, line 13: should these all be `AbstractComplex` ?

sethaxen: I don't think there is an `AbstractComplex` in base.

sethaxen:
Okay this should be ready for review.

@MasonProtter I ended up simplifying the rules for `abs`, `angle`, and `sign` with shared utility functions. They pass the same test suite, and locally their runtime seem to be the same as the specialized versions you implemented, but it'd be nice if you could double check.

MasonProtter:
Looks good to me!

src/rulesets/Base/fastmath_able.jl
```diff
@@ -126,7 +107,30 @@ let
...
-        @scalar_rule hypot(x::Real, y::Real) (x / Ω, y / Ω)
+
+        # `hypot`
+
+        function frule(
+                (_, Δx, Δy),
+                ::typeof(hypot),
+                x::T,
+                y::T,
+            ) where {T<:Union{Real,Complex}}
```
willtebbutt, line 118:
```suggestion
            (_, Δx, Δy),
            ::typeof(hypot),
            x::T,
            y::T,
        ) where {T<:Union{Real,Complex}}
```

PR #217, opened 2020-06-29
Generalize and test the rrules for sum

sethaxen:
This PR adds the missing test for `rrule(sum, abs2, x)`. It also release the constraints on `x` for the `sum` `rrule`s to be `Real`.

### DIFF
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -3,7 +3,7 @@
 #####
 
 function frule((_, ẋ), ::typeof(sum), x; dims=:)
-    return sum(x, dims=dims), sum(ẋ, dims=dims)
+    return sum(x; dims=dims), sum(ẋ; dims=dims)
 end
 
 function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
@@ -29,7 +29,7 @@ function frule(
     ∂y = if dims isa Colon
         2 * real(dot(x, ẋ))
     elseif VERSION ≥ v"1.2" # multi-iterator mapreduce introduced in v1.2
-        2 * mapreduce(_realconjtimes, +, x, ẋ; dims=dims);
+        2 * mapreduce(_realconjtimes, +, x, ẋ; dims=dims)
     else
         2 * sum(_realconjtimes.(x, ẋ); dims=dims)
     end
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -26,7 +26,7 @@ function rrule(
 ) where {T<:Union{Real,Complex}}
     y = sum(abs2, x; dims=dims)
     function sum_abs2_pullback(ȳ)
-        return (NO_FIELDS, DoesNotExist(), @thunk(2ȳ .* x))
+        return (NO_FIELDS, DoesNotExist(), @thunk(2 .* real.(ȳ) .* x))
     end
     return y, sum_abs2_pullback
 end
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -18,6 +18,22 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
     return y, sum_pullback
 end
 
+function frule(
+    (_, _, ẋ),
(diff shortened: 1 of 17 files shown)


willtebbutt:
Think this just needs a patch-version bump and we're good to go.

sethaxen:
> Think this just needs a patch-version bump and we're good to go.

We could either merge this now without a version bump, and it would be part of the large v0.7.0 release with fixes for complex rules, or (since these rules were previously constrained to the reals) we could wait until v0.7.0 is finished and merged and then make this v0.7.1.

willtebbutt:
I'm fine with either. Probably 0.7.1 would be better for the sake of keeping things nicely chopped up. @oxinabox do you have an opinion on this?

oxinabox:
Yeah i would rather wait til we release 0.7.0 before releasing any extra changes that are not strictly required.
Because having small incremental releases helped with bug tracking.
and 0.7 is a big set of changes intrinsitcally

sethaxen:
Thanks!

src/rulesets/Base/mapreduce.jl
```diff
@@ -2,8 +2,8 @@
 ##### `sum`
 #####
 
-function frule((_, ẋ), ::typeof(sum), x)
-    return sum(x), sum(ẋ)
+function frule((_, ẋ), ::typeof(sum), x; dims=:)
+    return sum(x, dims=dims), sum(ẋ, dims=dims)
```
willtebbutt, line 6:
```suggestion
    return sum(x; dims=dims), sum(ẋ; dims=dims)
```

src/rulesets/Base/mapreduce.jl
```diff
@@ -18,6 +18,22 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
...
+    ::typeof(sum),
+    ::typeof(abs2),
+    x::AbstractArray{T};
+    dims=:,
+) where {T<:Union{Real,Complex}}
+    y = sum(abs2, x; dims=dims)
+    ∂y = if dims isa Colon
+        2 * real(dot(x, ẋ))
+    else
+        2 * mapreduce(_realconjtimes, +, x, ẋ; dims=dims);
```
willtebbutt, line 32:
```suggestion
        2 * mapreduce(_realconjtimes, +, x, ẋ; dims=dims)
```

test/rulesets/Base/mapreduce.jl
```diff
@@ -1,51 +1,38 @@
...
-            end
-            @testset "keyword arguments" begin
-                n = 4
-                rrule_test(sum, randn(T, n, 1), (randn(T, n, n+1), randn(T, n, n+1)); fkwargs=(dims=2,))
+        sizes = (3, 4, 7)
+        @testset "dims = $dims" for dims in (:, 1)
+            fkwargs = (dims=dims,)
+            @testset "Array{$N, $T}" for N in eachindex(sizes), T in (Float64, ComplexF64)
+                s = sizes[1:N]
+                x, ẋ, x̄ = randn(T, s...), randn(T, s...), randn(T, s...)
```
willtebbutt, line 8:
```suggestion
                x = randn(T, s...)
                ẋ = randn(T, s...)
                x̄ = randn(T, s...)
```

test/rulesets/Base/mapreduce.jl
```diff
@@ -1,51 +1,38 @@
...
-                n = 4
-                rrule_test(
-                    sum,
-                    randn(n, 1),
-                    (abs2, nothing),
-                    (randn(T, n, n+1), randn(T, n, n+1));
-                    fkwargs=(dims=2,),
-                )
+        sizes = (3, 4, 7)
+        @testset "dims = $dims" for dims in (:, 1)
```
willtebbutt, line 19: This block repeats quite a lot from the `sum` tests above. Any chance you could figure out how to merge them to avoid the duplicate code?

sethaxen: I tried, but it just made the tests less readable. If we add more `sum(::Function, x)` type rules, those tests could be trivially merged with the `sum(abs2, x)` test though. Also, this way we can use `frule_test` for `sum` still.

willtebbutt: Fair enough.

Issue #214, opened 2020-06-27
Missing tests for BLAS rules


sethaxen:
Rules for the following functions are completely untested:
- `BLAS.dot`
- `BLAS.nrm2`
- `BLAS.asum`

Rules for the following functions are not constrained to real inputs but are untested for complex inputs:
- `BLAS.gemv`
- `BLAS.gemm`

Issue #210, opened 2020-06-22
Change convention for complex differentiation


sethaxen:
This issue tracks the necessary changes after https://github.com/JuliaDiff/ChainRulesCore.jl/issues/167 is merged:
- ChainRulesCore:
   - [x] Update `ChainRulesCore.@scalar_rule` to conjugate the partials in `rrule`. (breaking change) (https://github.com/JuliaDiff/ChainRulesCore.jl/pull/170)
- ChainRulesTestUtils:
    - [x] Bump ChainRulesCore compat  (https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/44)
    - [x] Update `ChainRulesTestUtils.test_scalar` to conjugate the derivative when testing `rrule`. (breaking change)(https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/44)
- ChainRules:
   - [x] Add missing tests for rules that take complex arguments (#214, #215)
   - [x] Bump ChainRulesCore, ChainRulesTestUtils, and FiniteDifferences compats (breaking change) (#198)
   - [x] Update non-scalar `rrule`s to follow the new convention (breaking change).
- Zygote:
   - [x] Bump ChainRules compat (https://github.com/FluxML/Zygote.jl/pull/721)
   - [x] Update Zygote's adjoints to stop conjugating to/from ChainRules (https://github.com/FluxML/Zygote.jl/pull/721)

I have local implementations of most of these already.

willtebbutt:
Happy to help out anywhere that it would be useful, but you're probably the best placed to deal with these.

oxinabox:
>  Update Zygote's adjoints to stop conjugating to/from ChainRules (breaking change)

This is not breaking for Zygote.
Its a nonbreaking feature to change what version of ChainRule's it supports

sethaxen:
Good point, thanks. Yes, if Zygote bumps the compat and stops conjugating to/from ChainRules at the same time, this is not a breaking change. I have updated.

PR #218, opened 2020-06-29
Fix complex array rules

sethaxen:
As part of #210, PR adds missing complex tests and fixes where necessary for the array rules of the following functions, which lack type constraints:
- `dot`
- `det`
- `logdet`
- `Symmetric`

 It does not release any type constraints. It also does not handle the BLAS rules (#214). This PR requires https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/47.

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -80,21 +80,6 @@
         end
     end
 
-    @testset "matmul *(x, y)" begin
-        x, y = rand(3, 2), rand(2, 5)
-        z, pullback = rrule(*, x, y)
-
-        @test z == x * y
-
-        z̄ = rand(3, 5)
-        (ds, dx, dy) = pullback(z̄)
-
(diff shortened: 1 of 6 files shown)


test/rulesets/Base/base.jl
```diff
@@ -80,21 +80,6 @@
         end
     end
 
-    @testset "matmul *(x, y)" begin
```
willtebbutt, line 83: Has this been superseded by some other tests somewhere?

sethaxen: There's already a matrix multiplication `rrule` test in the `dense.jl` file in `test`, so this is redundant. And the rule itself is defined in `dense.jl` in `src`.

sethaxen:
This is technically a breaking change, as it adopts the new complex rule conventions and is part of #210, so it should keep the current `-DEV` version number.

willtebbutt:
Good point.

PR #223, opened 2020-07-01
MersenneTwister

willtebbutt:
Ideally this would be covered with a nograd macro along the lines of https://github.com/JuliaDiff/ChainRulesCore.jl/issues/150. Unfortunately I've not implemented that feature yet, so here we are.


willtebbutt:
> Are you sure you don't want to implement that first and matching test utils for it?

No, but I've done this now. And this gives us a prime example of a no-grad-able thing to look at when someone (I'm thinking maybe @glennmoy ) gets around to implementing a no-grad thing.

PR #226, opened 2020-07-06
Add rules for specialized norm functions and normalize

### LINKED ISSUE #204: Add rrule for infinite-norm
The current `rrule` for `norm` fails when `p=Inf`, for which the norm is equivalent to `maximum(abs, x)`:

```julia
julia> using LinearAlgebra

julia> x = Float64[1, 2, -3];

julia> norm(x, Inf)
3.0

julia> _, back = ChainRules.rrule(norm, x, Inf);

julia> ∂self, ∂x, ∂p = back(1.0);

julia> ∂x() # expected: [0.0, 0.0, -1.0]
3-element Array{Float64,1}:
   0.0
 NaN
 NaN

julia> ∂p() # expected: 0.0 or Zero()
NaN
```

This really just fails for any very large `p` (e.g. also fails for `p=1e3`). It works fine for `p=100`:

```julia
julia> norm(x, 1e2)
3.0

julia> _, back = ChainRules.rrule(norm, x, 1e2);

julia> ∂self, ∂x, ∂p = back(1.0);

julia> ∂x()
3-element Array{Float64,1}:
  5.820975652447899e-48
  3.689481639869744e-18
 -1.0

julia> ∂p()
0.0
```

The actual implementation of `norm` in LinearAlgebra special-cases `p` of 0, 1, 2, `-Inf`, and `Inf`, falling back to `normp` if another `p` is provided (see https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L604-L619). Should the `rrule` for `norm` here instead be replaced by `rrules` for these functions? I see 2 downsides: 1) more code 2) pullback of `p` for these special `p`'s would be `Zero()`, and in general would not be type-stable when pulling back through `norm`.
On the upside, the pullbacks for these types can be implemented a little more efficiently.

sethaxen:
This PR adds ~`frule`s~ (edit: it only adds the `frule` for the scalar function and `norm` without `p`) and `rrule`s of the various generic norm functions (and friends) in LinearAlgebra:
- `norm`
- `norm2`
- `norm1`
- `normInf`
- `normMinusInf`
- `normp`
- `normalize`
- ~`normalize!` (`frule` only)~

The `frule`s are patterned after the primal functions and work on iterators. The `rrule`s are restrict to `AbstractArray` primal inputs because I don't see how to generically write `rrule`s for iterators. This PR replaces the current `rrule` for `norm` to dispatch to the other rules. The rules use the subgradient convention throughout. ~I'm still working on the test suite.~

This PR will fix #204 and https://github.com/FluxML/Zygote.jl/issues/663.


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -282,22 +282,349 @@ end
...
-    function norm_pullback(ȳ)
-        u = y^(1-p)
-        ∂A = @thunk ȳ .* u .* abs.(A).^p ./ A
-        ∂p = @thunk ȳ * (u * sum(a->abs(a)^p * log(abs(a)), A) - y * log(y)) / p
-        (NO_FIELDS, ∂A, ∂p)
+function frule((_, Δx, Δp), ::typeof(norm), x, p::Real)
+    return if isempty(x)
+        z = float(norm(zero(eltype(x))))
+        z, zero(z * zero(Δp) + z * zero(Δx))
+    elseif p == 2
```
oxinabox, line 289: Arguably might want to use `===` here. If we are willing to commit to that require
The compiler has historically been better about constant folding  that (but i haven't checked recently).
and it is immune to invalidations i guess.

sethaxen: Hm, I can't see a difference in benchmarks, but I'm not certain the best way to check on constant folding or invalidations. The arguments for using `==` that I can see are 1) it mirrors the primal function and 2) it enables the more efficient methods to be called even when the user calls `norm(x, 2.0)` instead of `norm(x, 2)` .

willtebbutt:
@sethaxen Would you be willing to restrict the types to e.g. `StridedArray` wherever you currently have `AbstractArray` to prevent the rule from acting on things like `Diagonal` and producing dense pullbacks?

I'm generally trying to move our codebase away implementing rules for `AbstractArray`s -- we'll need a big breaking release at some point where we break a load of Zygote stuff / have to implement a load of Zygote-specific patches in Zygote, but for now it seems reasonable to restrict new stuff to try and limit the pain of the transition.

nickrobinson251:
> Would you be willing to restrict the types to e.g. StridedArray wherever you currently have AbstractArray to prevent the rule from acting on things like Diagonal and producing dense pullbacks?
> I'm generally trying to move our codebase away implementing rules for `AbstractArray`s 

@willtebbutt can you open an issue about this (if there's not one already)? It'd be good to track this work, and we may also want to add some guidance about this in the docs

sethaxen:
> @sethaxen Would you be willing to restrict the types to e.g. `StridedArray` wherever you currently have `AbstractArray` to prevent the rule from acting on things like `Diagonal` and producing dense pullbacks?
>
> I'm generally trying to move our codebase away implementing rules for `AbstractArray`s -- we'll need a big breaking release at some point where we break a load of Zygote stuff / have to implement a load of Zygote-specific patches in Zygote, but for now it seems reasonable to restrict new stuff to try and limit the pain of the transition.

I have some thoughts on this I'll add to the issue you created when I find some time. Things like `Diagonal` and `AbstractTriangular` already do the right thing in the pullback (because the pullback broadcasts over the primal input), but I'd need to double check for primals like `Symmetric`.

sethaxen:
Okay @willtebbutt, I restricted the `rrule`s to primal types that I know do the right thing, pending resolution of #232.

I think all that's left is to test `norm(::Union{Adjoint{Vector},Transpose{Vector}}[,p])`.

simeonschaub:
Bump. I just wanted this, looks like this just needs a rebase.

sethaxen:
Hmm, yeah, I need to remember why I paused this. I think because with the type constraints on the `rrule`, it wouldn't be picked up by `Zygote` (which has a more generic `AbstractArray` adjoint), which was the original motivation for this PR. Still, there's more than Zygote, so this will be good to have. I'll try to wrap it up in the next few days.

simeonschaub:
Do you mean this one? https://github.com/FluxML/Zygote.jl/blob/a04adb8ceb020ffda312677b37c2f64c6c2ab1ee/src/lib/array.jl#L439. That one is pretty nasty, there must be some better way to do this.

oxinabox:
We should delete the one from Zygote once we have this.
Every array that matters is eventually an wrapper of a DenseArray,
Or a FillArray, but one has to be very careful with their rules anyway.
and Zygote can work them out mostly without rules anyway.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -282,22 +282,396 @@ end
 ##### `norm`
```
oxinabox, line 282: Probably lets give have a `norm.jl` file, since this is looong

sethaxen:
This PR is now ready for review. A few notes:
- I removed almost all of the `frule`s. In almost every case, I was just hardcoding what a sane AD should do, and because the primal and `frule` were so intertwined, it would be annoying to maintain. I did keep the `frule` for `norm(x)` and `norm2(x)` though, because there the efficient implementation does not fuse with the primal.
- This PR depends on https://github.com/JuliaDiff/FiniteDifferences.jl/pull/118.
- CI won't run until #309 is merged

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -0,0 +1,248 @@
...
+    else
+        signx = x isa Real ? sign(x) : x * pinv(y)
+        _realconjtimes(signx, Δx)
+    end
+    return y, ∂y
+end
+
+function rrule(
+    ::typeof(norm),
+    x::Union{StridedArray,LinearAlgebra.AbstractTriangular,Diagonal},
```
willtebbutt, line 27:
```suggestion
    x::Union{StridedArray, LinearAlgebra.AbstractTriangular, Diagonal},
```

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -0,0 +1,248 @@
...
+        end
+        ∂p = @thunk _normp_back_p(x, p, y, Δy)
+        return (NO_FIELDS, ∂x, ∂p)
+    end
+    norm_pullback(::Zero) = (NO_FIELDS, Zero(), Zero())
+    return y, norm_pullback
+end
+function rrule(
+    ::typeof(norm),
+    x::Union{StridedArray,LinearAlgebra.AbstractTriangular,Diagonal},
```
willtebbutt, line 57:
```suggestion
    x::Union{StridedArray, LinearAlgebra.AbstractTriangular, Diagonal},
```

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -0,0 +1,248 @@
...
+            _norm2_back(x, y, Δy)
+        end
+        return (NO_FIELDS, ∂x)
+    end
+    norm_pullback(::Zero) = (NO_FIELDS, Zero())
+    return y, norm_pullback
+end
+function rrule(
+    ::typeof(norm),
+    x::Union{LinearAlgebra.TransposeAbsVec,LinearAlgebra.AdjointAbsVec},
```
willtebbutt, line 73:
```suggestion
    x::Union{LinearAlgebra.TransposeAbsVec, LinearAlgebra.AdjointAbsVec},
```

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -0,0 +1,248 @@
...
+    return y, norm_pullback
+end
+
+#####
+##### `normp`
+#####
+
+function rrule(
+    ::typeof(LinearAlgebra.normp),
+    x::Union{StridedArray,LinearAlgebra.AbstractTriangular,Diagonal},
```
willtebbutt, line 106:
```suggestion
    x::Union{StridedArray, LinearAlgebra.AbstractTriangular, Diagonal},
```

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -0,0 +1,248 @@
...
+    return ∂p
+end
+
+#####
+##### `normMinusInf`/`normInf`
+#####
+
+function rrule(
+    ::typeof(LinearAlgebra.normMinusInf),
+    x::Union{StridedArray,LinearAlgebra.AbstractTriangular,Diagonal},
```
willtebbutt, line 147:
```suggestion
    x::Union{StridedArray, LinearAlgebra.AbstractTriangular, Diagonal},
```

willtebbutt (approved):
I'm approving subject to the things that you note above being resolved as I don't want to hold you back, and deciding whether to release a new minor version (if we decide to do that, then we should probably -DEV briefly, and bundle it in with #311  to avoid making multiple breaking releases.

I've made some style suggestions to make the type unions more readable.

We should discuss whether or not it's a breaking change though. Before, would you have expected things to break if someone passed in a type outside of the new constraints?

sethaxen:
> We should discuss whether or not it's a breaking change though. Before, would you have expected things to break if someone passed in a type outside of the new constraints?

Not necessarily. The rule _should_ have worked for any array type whose differential could be represented by scaling the array via broadcast. So it would have worked for `Hermitian`/`Symmetric` as well as `StaticArray`s, but would probably have failed for others, in particular for arrays like `UnitLowerTriangular`, where the diagonal of the differential should be zero but would not have been.

willtebbutt:
Hmmm interesting. So it probably fixes some potential bugs, but might remove some functionality. @nickrobinson251 any thoughts on the correct course of action here?

sethaxen:
@nickrobinson251 what are your thoughts on the version number here?

sethaxen:
Okay, this PR should be 100% complete, just pending input on choice of version number.

nickrobinson251:
I'd say this is a bug fix 👍 
Thanks a lot!

nickrobinson251 (approved):

oxinabox:
Agreed, bug-fixes are never considered breaking.
Plus, this didn't break Zygote's tests, and I know it won't break ReversePropagation's Tests sice that only uses our scalar rules.
I am pretty sure it won't break Nabla.

Also it's much easier to deal with something incorrectly called non-breaking (revert, tag a patch, unrevert tag a breaking).
There isn't really any way to undo something bring considered breaking when it wasn't

sethaxen:
Great! Thanks to you both!

PR #228, opened 2020-07-06
Fix signature for MersenneTwister

### LINKED ISSUE #227: no method matching MersenneTwister(::typeof(abs2), ::Array{Complex{Float64},1})
I suddenly got a mysterious error on CI from a call to Zygote.gradient
```
LoadError: MethodError: no method matching MersenneTwister(::typeof(abs2), ::Array{Complex{Float64},1})
```
I haven't touched the code that threw the error in a long time so something must have gone wild in Zygote or ChainRules
link to CI failure: https://github.com/baggepinnen/SpectralDistances.jl/runs/840227039#step:4:721
link to failing call: https://github.com/baggepinnen/SpectralDistances.jl/blob/master/test/test_diff.jl#L204 

Probably caused by https://github.com/JuliaDiff/ChainRules.jl/commit/697e7e4e168015d216c86863c5c205f954766ddc

```
julia version 1.4.2
Installed ChainRules ─────────────────── v0.7.5
Installed Zygote ─────────────────────── v0.5.2
Installed ZygoteRules ────────────────── v0.2.0
```

sethaxen:
This PR (probably) fixes #227. `typeof(MersenneTwister)` is `DataType`, so the rules introduced in #223 will be applied to many function calls.

@baggepinnen can you confirm this resolves the issue for you?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.5"
+version = "0.7.6"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Random/random.jl
+++ b/src/rulesets/Random/random.jl
@@ -1,6 +1,6 @@
-frule(Δargs, ::typeof(MersenneTwister), args...) = MersenneTwister(args...), Zero()
+frule(Δargs, ::Type{MersenneTwister}, args...) = MersenneTwister(args...), Zero()
 
-function rrule(::typeof(MersenneTwister), args...)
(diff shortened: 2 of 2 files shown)


simeonschaub:
Would it be possible to add a test for this?

nickrobinson251:
While fixing-up this code, would you mind also changing `MersenneTwister_rrule` -> `MersenneTwister_pullback`, to match the style elsewhere in the package?

nickrobinson251:
I think this fix is pretty hard to test. We'd need to test that we didn't accidentally define a rule for `DataType`. And given this is breaking the Zygote tests I think it'd be good to merge the fix asap (presuming we've verified this does indeed fix the problem).

nickrobinson251:
This fixes the issue for me, so i'll just merge and make a release since it's affecting Zygote and Flux. Thanks @sethaxen!

Issue #229, opened 2020-07-06
Should we run Zygote tests as part of CI?


nickrobinson251:
Raised in https://github.com/JuliaDiff/ChainRules.jl/issues/227#issuecomment-654096195

We could copy what we do in ChainRulesCore for ChainRules integration tests already: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/master/.cirrus.yml Would be good to have a separate runner for this, so the test suite doesn't take forever to run.

The idea is to stress test new rules (e.g. check they're constrained to the right inputs as expected).

oxinabox:
Once ForwardDiff2 gets back up to the latest version of ChainRules we should add it also

oxinabox:
Potentially we should use Github actions instead of CirrusCI,
I think it is faster, esp if we have to end up doing a bunch of these reverse dependency tests.

PR #231, opened 2020-07-06
Add github action for Zygote CI

sethaxen:
This PR adds a Github Action (as proposed by @oxinabox in #229) that runs the test suite of the master branch of Zygote using the current version of ChainRules (master branch, PR, or tagged release).


.github/workflows/ZygoteCI.yml
```diff
@@ -0,0 +1,30 @@
...
+          version: ${{ matrix.julia-version }}
+          arch: x64
+      - uses: julia-actions/julia-buildpkg@latest
+      - name: Clone Zygote
+        uses: actions/checkout@v2
+        with:
+          repository: FluxML/Zygote.jl
+          path: "Zygote.jl"
+      - name: Run the tests
+        run: julia --project="Zygote.jl" -e "using Pkg; Pkg.develop(PackageSpec(path=\"./\")); Pkg.update(); Pkg.test()"
```
oxinabox, line 29: this dev's the current branch of ChainRules, right?

.github/workflows/ZygoteCI.yml
```diff
@@ -0,0 +1,30 @@
...
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: Julia v${{ matrix.julia-version }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: [1.0, 1.4]
```
oxinabox, line 14: Lets just do:
```suggestion
        julia-version: [1]
```

I think Zygote is dropping support for the LTS soon and I think 1 gets the latest stable release?

.github/workflows/ZygoteCI.yml
```diff
@@ -0,0 +1,30 @@
+name: ZygoteCI
```
oxinabox, line 1:
```suggestion
name: IntegrationTestZygote 
```

.github/workflows/ZygoteCI.yml
sethaxen: I think so, but to make sure, I'll introduce something that will definitely break Zygote and then if it fails revert before merging.

.github/workflows/IntegrationTestZygote.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+
+jobs:
+  test:
+    name: Julia v${{ matrix.julia-version }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: [1]
+        os: [ubuntu-latest]
```
simeonschaub, line 15: I don't think the matrix is necessary here, since you can just hard code the values, but doesn't really matter,

sethaxen: Yeah, it's just there in case we ever want to add other versions/oses, and it currently spawns the same one job we would get if we just hard-coded them.

simeonschaub: Makes sense!

PR #233, opened 2020-07-08
fix division return type 

CarloLucibello:
Fix https://github.com/FluxML/Zygote.jl/issues/727. 

Still have to add tests. Where are `rrule_test`, `frule_test` and `scalar_test` defined? Do they check types?

### DIFF
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -133,7 +133,7 @@ let
 
         @scalar_rule x + y (One(), One())
         @scalar_rule x - y (One(), -1)
-        @scalar_rule x / y (inv(y), -((x / y) / y))
+        @scalar_rule x / y (one(Ω) / y, -(Ω / y))
         #log(complex(x)) is required so it gives correct complex answer for x<0
         @scalar_rule(x ^ y,
             (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(complex(x))),
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -133,7 +133,7 @@ let
 
         @scalar_rule x + y (One(), One())
         @scalar_rule x - y (One(), -1)
-        @scalar_rule x / y (one(Ω) / y, -(Ω / y))
+        @scalar_rule x / y (one(x) / y, -(Ω / y))
         #log(complex(x)) is required so it gives correct complex answer for x<0
         @scalar_rule(x ^ y,
             (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(complex(x))),
@@ -141,7 +141,7 @@ let
         # x^y for x < 0 errors when y is not an integer, but then derivative wrt y
         # is undefined, so we adopt subgradient convention and set derivative to 0.
         @scalar_rule(x::Real ^ y::Real,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(convert(typeof(Ω), ifelse(x ≤ 0, one(x), x)))),
+            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x)))),
         )
         @scalar_rule(
             rem(x, y),
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -113,16 +113,16 @@ let
         function frule(
             (_, Δx, Δy),
             ::typeof(hypot),
-            x::T1,
-            y::T2,
-        ) where {T1<:Union{Real,Complex}, T2<:Union{Real,Complex}}
+            x::T,
+            y::T,
+        ) where {T<:Union{Real,Complex}}
             Ω = hypot(x, y)
             n = ifelse(iszero(Ω), one(Ω), Ω)
             ∂Ω = (_realconjtimes(x, Δx) + _realconjtimes(y, Δy)) / n
             return Ω, ∂Ω
         end
 
-        function rrule(::typeof(hypot), x::T1, y::T2) where {T1<:Union{Real,Complex}, T2<:Union{Real,Complex}}
+        function rrule(::typeof(hypot), x::T, y::T) where {T<:Union{Real,Complex}}
             Ω = hypot(x, y)
             function hypot_pullback(ΔΩ)
                 c = real(ΔΩ) / ifelse(iszero(Ω), one(Ω), Ω)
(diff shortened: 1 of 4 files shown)


nickrobinson251:
> Where are rrule_test, frule_test and scalar_test defined?

They are [defined in ChainRulesTestUtils.jl](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/23c78a6d9168dd843340399d5df7f292dba97356/src/testers.jl#L62-L205). 

(This is a separate package [to keep dependencies separate](https://www.juliadiff.org/ChainRulesCore.jl/dev/FAQ.html#How-do-I-test-my-rules?-1))

> Do they check types?

No, I think they only test via comparing to Finite Differencing

willtebbutt:
Yeah, and `FiniteDifferencing` is still quite bad at preserving the types correctly, which is a bit of a shame.

@CarloLucibello in terms of testing this rule, the correctness tests should be fine as is, and it would be sufficient for you to add some basic regression tests on the type of the differential. Could you also please open an issue on `ChainRulesTestUtils` about this? It seems like the kind of thing that we should be testing for all of the time, and is probably part of a more general "testing that you've produced a valid differential type" in the rules.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -133,7 +133,7 @@ let
 
         @scalar_rule x + y (One(), One())
         @scalar_rule x - y (One(), -1)
-        @scalar_rule x / y (inv(y), -((x / y) / y))
+        @scalar_rule x / y (one(Ω) / y, -(Ω / y))
```
sethaxen, line 136: Because `d/dx (x) = one(x)`, shouldn't this be `one(x)`?  It probably doesn't matter in general but could in some cases. e.g. if `x` is real but `y` is complex, your version will have more operations than this one.
```suggestion
        @scalar_rule x / y (one(x) / y, -(Ω / y))
```

CarloLucibello: I did that because the following thought crossed my mind: that a closure referencing only Ω  could be more memory efficient in some situations than one referencing both x and Ω. I have no idea if in practice this is actually true, could be?

test/rulesets/Base/fastmath_able.jl
```diff
@@ -131,13 +131,25 @@ const FASTABLE_AST = quote
...
 
-        @testset "$f(x::$T, y::$T)" for f in (/, +, -, hypot), T in (Float64, ComplexF64)
+        @testset "$f(x::$T, y::$T)" for f in (/, +, -, hypot), T in (Float32, Float64, ComplexF64)
             x, Δx, x̄ = 10rand(T, 3)
             y, Δy, ȳ = rand(T, 3)
             Δz = randn(typeof(f(x, y)))
 
-            frule_test(f, (x, Δx), (y, Δy))
-            rrule_test(f, Δz, (x, x̄), (y, ȳ))
+            if T != Float32
```
CarloLucibello, line 139: had to exclude this case because test tolerance is too strict for Float32

CarloLucibello:
I can't find a rule for `\` (which may need a similar change, see https://github.com/JuliaDiff/DiffRules.jl/pull/46)

sethaxen:
> I can't find a rule for `\` (which may need a similar change, see [JuliaDiff/DiffRules.jl#46](https://github.com/JuliaDiff/DiffRules.jl/pull/46))

Here it is: https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/rulesets/Base/base.jl#L141

src/rulesets/Base/fastmath_able.jl
```diff
@@ -113,16 +113,16 @@ let
         function frule(
             (_, Δx, Δy),
             ::typeof(hypot),
-            x::T,
-            y::T,
-        ) where {T<:Union{Real,Complex}}
+            x::T1,
+            y::T2,
+        ) where {T1<:Union{Real,Complex}, T2<:Union{Real,Complex}}
```
sethaxen, line 118: Is this signature just changed so that the tests in the loop pass? If so, maybe it's better to just take that test out of the loop. The signature we had here is the one that is FastMath compatible. There is a `hypot` that takes mixed arguments, but it promotes before calling this one, so currently we would just let the AD handle that. Same with the `rrule`.

test/rulesets/Base/fastmath_able.jl
sethaxen: Could you decrease the tolerance using the keyword arguments of the testers?
e.g. something like `frule_test(f, (x, Δx), (y, Δy); atol=0, rtol = sqrt(eps(T)))`

src/rulesets/Base/fastmath_able.jl
sethaxen: Hm maybe? I don't know enough about the performance of closures.

DhairyaLGandhi: I think performance wise those should not pose too much of a difference

CarloLucibello:
zygote failure is unrelated

src/rulesets/Base/fastmath_able.jl
```diff
@@ -133,15 +133,15 @@ let
...
+        @scalar_rule x / y (one(Ω) / y, -(Ω / y))
         #log(complex(x)) is required so it gives correct complex answer for x<0
         @scalar_rule(x ^ y,
             (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(complex(x))),
         )
         # x^y for x < 0 errors when y is not an integer, but then derivative wrt y
         # is undefined, so we adopt subgradient convention and set derivative to 0.
         @scalar_rule(x::Real ^ y::Real,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(ifelse(x ≤ 0, one(x), x))),
+            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(convert(typeof(Ω), ifelse(x ≤ 0, one(x), x)))),
```
CarloLucibello, line 144: I did this because otherwise for integer x and float32  y I was obtaining
```julia
  Expression: frule((Zero(), Δx, Δy), f, 2, y) isa Tuple{T, T}
   Evaluated: (1.2687975f0, 0.44214729389061513) isa Tuple{Float32,Float32}
```
I'm not sure convert is the best way to do it or if we want to care about this case at all. If we don't I'll just remove this change and the corresponding test

src/rulesets/Base/fastmath_able.jl
sethaxen: Yeah, I still prefer for `one(x)`, but I won't hold this PR up over it.

src/rulesets/Base/fastmath_able.jl
sethaxen: Good catch. This approach looks right. Perhaps this is cleaner?
```suggestion
            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x)))),
```

CarloLucibello:
again Zygote failure unrelated. This should be ready to go

CarloLucibello:
merge and tag?

willtebbutt:
Please. Thanks for the contribution

PR #234, opened 2020-07-12
Add logabsdet and test det and logdet on numbers

sethaxen:
`det` and `logdet` were implemented for numbers but were untested. This PR adds number tests and removes the dangerous `inv` in their rules for number inputs. (see https://github.com/FluxML/Zygote.jl/issues/727#issuecomment-655361588).

It also adds rules for `logabsdet`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.9"
+version = "0.7.10"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


sethaxen:
The `rrule` for `logabsdet` differs from [Zygote's](https://github.com/FluxML/Zygote.jl/blob/v0.5.3/src/lib/array.jl#L370) because Zygote's discards the adjoint of the sign, which is relevant for complex matrices:

sethaxen:
One thing I _could_ add to this PR are similar rules for `det`, `logdet`, and `logabsdet` to the `logabsdet` rules in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/191 for the concrete `Matrix` type, where we know that `lu` is used for the primal and so we can reuse it for the inverse.

PR #236, opened 2020-07-15
add adjoint for reduce hvcat

piever:
Code is basically the same as @AzamatB's https://github.com/FluxML/Zygote.jl/pull/501 (plus tests), but I think now it makes more sense to have this here.

This fixes `reduce(hcat, As)` and `reduce(vcat, As)` for Zygote, as otherwise you'd get the mutation error.

### DIFF
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -40,12 +40,12 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
 end
 
 function rrule(::typeof(reduce), ::typeof(hcat), As::AbstractVector{<:AbstractVecOrMat})
-    function reduce_hcat_pullback(Ȳ)
+    function reduce_hcat_pullback(ΔY)
         sizes = size.(As, 2)
         cumsizes = cumsum(sizes)
         ∂As = map(cumsizes, sizes) do post, diff
             pre = post - diff + 1
-            Ȳ[:, pre:post]
+            return ΔY[:, pre:post]
         end
         return (NO_FIELDS, NO_FIELDS, ∂As)
     end
@@ -71,12 +71,12 @@ function rrule(::typeof(vcat), A::AbstractArray, Bs::AbstractArray...)
 end
 
 function rrule(::typeof(reduce), ::typeof(vcat), As::AbstractVector{<:AbstractVecOrMat})
-    function reduce_vcat_pullback(Ȳ)
+    function reduce_vcat_pullback(ΔY)
         sizes = size.(As, 1)
         cumsizes = cumsum(sizes)
         ∂As = map(cumsizes, sizes) do post, diff
             pre = post - diff + 1
-            Ȳ[pre:post, :]
+            return ΔY[pre:post, :]
         end
         return (NO_FIELDS, NO_FIELDS, ∂As)
     end
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -47,7 +47,7 @@ function rrule(::typeof(reduce), ::typeof(hcat), As::AbstractVector{<:AbstractVe
             pre = post - diff + 1
             return ΔY[:, pre:post]
         end
-        return (NO_FIELDS, NO_FIELDS, ∂As)
+        return (NO_FIELDS, DoesNotExist(), ∂As)
     end
     return reduce(hcat, As), reduce_hcat_pullback
 end
@@ -78,7 +78,7 @@ function rrule(::typeof(reduce), ::typeof(vcat), As::AbstractVector{<:AbstractVe
             pre = post - diff + 1
             return ΔY[pre:post, :]
         end
-        return (NO_FIELDS, NO_FIELDS, ∂As)
+        return (NO_FIELDS, DoesNotExist(), ∂As)
     end
     return reduce(vcat, As), reduce_vcat_pullback
 end
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -39,6 +39,19 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
     return hcat(A, Bs...), hcat_pullback
 end
 
+function rrule(::typeof(reduce), ::typeof(hcat), As::AbstractVector{<:AbstractVecOrMat})
+    function reduce_hcat_pullback(Ȳ)
+        sizes = size.(As, 2)
+        cumsizes = cumsum(sizes)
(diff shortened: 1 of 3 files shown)


src/rulesets/Base/array.jl
```diff
@@ -39,6 +39,19 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
...
 
+function rrule(::typeof(reduce), ::typeof(hcat), As::AbstractVector{<:AbstractVecOrMat})
+    function reduce_hcat_pullback(Ȳ)
+        sizes = size.(As, 2)
+        cumsizes = cumsum(sizes)
+        ∂As = map(cumsizes, sizes) do post, diff
+            pre = post - diff + 1
+            Ȳ[:, pre:post]
+        end
+        return (NO_FIELDS, NO_FIELDS, ∂As)
```
sethaxen, line 50: I think `DoesNotExist()` should be used here, as `NO_FIELDS` is reserved for the calling function. This is what we do for e.g. `sum(abs, x)`, but I could be wrong about this:

```suggestion
        return (NO_FIELDS, DoesNotExist(), ∂As)
```

src/rulesets/Base/array.jl
```diff
@@ -39,6 +39,19 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
     return hcat(A, Bs...), hcat_pullback
 end
 
+function rrule(::typeof(reduce), ::typeof(hcat), As::AbstractVector{<:AbstractVecOrMat})
+    function reduce_hcat_pullback(Ȳ)
```
sethaxen, line 43: Could you use our `ΔY` notation? I know we're not entirely consistent in this package, but since we've [documented the standard](http://www.juliadiff.org/ChainRulesCore.jl/stable/FAQ.html#%CE%94x,-x,-dx-1), probably new rules should follow suit.

```suggestion
    function reduce_hcat_pullback(ΔY)
```

src/rulesets/Base/array.jl
```diff
@@ -39,6 +39,19 @@ function rrule(::typeof(hcat), A::AbstractArray, Bs::AbstractArray...)
...
+        sizes = size.(As, 2)
+        cumsizes = cumsum(sizes)
+        ∂As = map(cumsizes, sizes) do post, diff
+            pre = post - diff + 1
+            Ȳ[:, pre:post]
```
sethaxen, line 48: Blue Style uses explicit returns:
```suggestion
            return ΔY[:, pre:post]
```

src/rulesets/Base/array.jl
```diff
@@ -57,6 +70,19 @@ function rrule(::typeof(vcat), A::AbstractArray, Bs::AbstractArray...)
     return vcat(A, Bs...), vcat_pullback
 end
 
+function rrule(::typeof(reduce), ::typeof(vcat), As::AbstractVector{<:AbstractVecOrMat})
+    function reduce_vcat_pullback(Ȳ)
+        sizes = size.(As, 1)
+        cumsizes = cumsum(sizes)
+        ∂As = map(cumsizes, sizes) do post, diff
+            pre = post - diff + 1
+            Ȳ[pre:post, :]
```
sethaxen, line 79:
```suggestion
            return ΔY[pre:post, :]
```

src/rulesets/Base/array.jl
```diff
@@ -57,6 +70,19 @@ function rrule(::typeof(vcat), A::AbstractArray, Bs::AbstractArray...)
     return vcat(A, Bs...), vcat_pullback
 end
 
+function rrule(::typeof(reduce), ::typeof(vcat), As::AbstractVector{<:AbstractVecOrMat})
+    function reduce_vcat_pullback(Ȳ)
```
sethaxen, line 74:
```suggestion
    function reduce_vcat_pullback(ΔY)
```

src/rulesets/Base/array.jl
```diff
@@ -57,6 +70,19 @@ function rrule(::typeof(vcat), A::AbstractArray, Bs::AbstractArray...)
...
+        sizes = size.(As, 1)
+        cumsizes = cumsum(sizes)
+        ∂As = map(cumsizes, sizes) do post, diff
+            pre = post - diff + 1
+            Ȳ[pre:post, :]
+        end
+        return (NO_FIELDS, NO_FIELDS, ∂As)
```
sethaxen, line 81:
```suggestion
        return (NO_FIELDS, DoesNotExist(), ∂As)
```

test/rulesets/Base/array.jl
```diff
@@ -34,6 +34,21 @@ end
...
+    H, pullback = rrule(reduce, hcat, [A, B, C])
+    @test H == reduce(hcat, [A, B, C])
+    H̄ = randn(3, 6)
+    (ds, dh, (dA, dB, dC)) = pullback(H̄)
+    @test ds == NO_FIELDS
+    @test dh == NO_FIELDS
+    @test dA ≈ view(H̄, :, 1:2)
+    @test dB ≈ view(H̄, :, 3:3)
+    @test dC ≈ view(H̄, :, 4:6)
+end
```
sethaxen, line 50: Are you able to use `ChainRulesTestUtils.rrule_test` here and below? It's safest to verify with finite differences in case the derivation is wrong.

src/rulesets/Base/array.jl
piever: This is the one style change where I'm not sure I agree.

```julia
help?> NO_FIELDS
search: NO_FIELDS

  NO_FIELDS

  Constant for the reverse-mode derivative with respect to a structure that
  has no fields. The most notable use for this is for the reverse-mode
  derivative with respect to the function itself, when that function is not a
  closure.

help?> DoesNotExist
search: DoesNotExist

  DoesNotExist() <: AbstractZero

  This differential indicates that the derivative does not exist. It is the
  differential for a Primal type that is not differentiable. Such an Integer,
  or Boolean (when not being used as a represention of a value that normally
  would be a floating point.)
```

A function (not a closure) is either "struct with no fields" or "not differentiable primal type": how can it be one as first argument to `rrule`, and the other as second argument?

src/rulesets/Base/array.jl
piever: Sure, I've taken the notation from `hcat` above: I guess `hcat` can be updated to follow the standard in a future PR.

piever:
Thanks for the review! I've addressed the comments, it should be good to go now.

piever:
I've just bumped the release. Could you squash and merge when tests pass? I'm not a JuliaDiff member so can't really merge myself.

PR #239, opened 2020-07-16
simplify sum rrule

piever:
I imagine this was an oversight (unless of course I'm missing some subtlety).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.9"
+version = "0.7.10"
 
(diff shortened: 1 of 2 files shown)


willtebbutt:
I agree that this looks like an oversight. I suspect that the current implementation will give the wrong behaviour for anything that 's not a `Number` -- could you add an `rrule_test` e.g. for summing over a vector of vectors to prevent regression?

piever:
> could you add an `rrule_test` e.g. for summing over a vector of vectors to prevent regression?

This `rrule` specializes on `AbstractArray{<:Number}`, so I imagine that's not possible (that is, there is no rule to test).

piever:
Sure, I've just increased the patch version number.

PR #240, opened 2020-07-17
Inplace getindex rrule

oxinabox:
Something @willtebbutt pointed out to me the other day
We can use ChainRules's ability to define an inplace addition operation (the InplaceThunk) when defining the gradient of getindex.
Which would be much more efficient than e.g retuning a onehot dense matrix to be summed, and marginally more efficient than returning a onehot sparse matrix to be summed.
It would also mean maybe we would not need to say that the differential for a primal type depends on the primal type and the operation.
Though we still might want to.

Right now Zygote won't use ChainRules's inplace accumulation stuff, but idk how hard it would be to enable it.
I think it might be fine or it might (only) break  nesting Zygote.
It would definitely require the deeper change over to ChainRules's types.

I am tempted to leave this WIP for a while and improve our abstractions, while trying to get Zygote to actually use this. (Though that is a bigger project as need Zygote to use ChainRules's trypes) 

- [x] Test single indexes
- [x] Make it so can test this with ChainRulesTestUtils (improve abstractions first?)


oxinabox:
For Zygote to benifit it needs https://github.com/FluxML/Zygote.jl/issues/603
and indeed this is itself another good motivation to do that.

test/rulesets/Base/array.jl
```diff
@@ -75,3 +75,10 @@ end
     @test dd isa DoesNotExist
     @test dv ≈ 27.0
 end
+
+@testset "getindex" begin
+    x = [1.0 2.0 3.0; 10.0 20.0 30.0]
+    x̄ = [1.4 2.5 3.7; 10.5 20.1 30.2]
+    rrule_test(getindex, 2.3, (x, x̄), (2, nothing))
+    rrule_test(getindex, 2.3, (x, x̄), (2, nothing), (1, nothing))
```
oxinabox, line 83:
```suggestion
    rrule_test(getindex, 2.3, (x, x̄), (2, nothing), (1, nothing))
    
    # overlapping indexes
    rrule_test(getindex, 2.3, (x, x̄), (2, nothing), (2, nothing))
    rrule_test(getindex, 2.3, (x, x̄), (2, nothing), (1, nothing), (2, nothing))
```

test/rulesets/Base/array.jl
```diff
@@ -75,3 +75,14 @@ end
...
+    
+    # overlapping indexes
```
oxinabox, line 85: These don't actually overlap and overlapping indexs don't seem possible for this form of `getindex` Still i would like to keep the extra tests
```suggestion
```

src/rulesets/Base/array.jl
```diff
@@ -102,3 +102,25 @@ function rrule(::typeof(fill), value::Any, dims::Int...)
...
+
+#####
+##### getindex
+#####
+
+function rrule(::typeof(getindex), x::Array{<:Number}, inds::Vararg{Int})
+    y = getindex(x, inds...)
+    function getindex_pullback(ȳ)
+        function getindex_add!(Δ)
+            Δ[inds...] = Δ[inds...] .+ ȳ
```
oxinabox, line 114: Out of scope for this PR probably but i wonder if the way to extend this for more things like `Array{AbstractArray}` is:

```suggestion
            Δ[inds...] = add!!(Δ[inds...], ȳ)
```

oxinabox:
Ok, seems like i need to do a bunch more methods of `getindex` to make this work for Nabla so this PR is going to get bigger.

willtebbutt:
This broadly LGTM. It would be good to add some tests involving arrays with an element type that's not a number -- perhaps an array of arrays?

oxinabox:
> This broadly LGTM. It would be good to add some tests involving arrays with an element type that's not a number -- perhaps an array of arrays?

Good Idea
Probably don't need to many of them since most of the tests are about indexing behavour.
Just some to make sure it goes ok

PR #245, opened 2020-08-17
Restrict types in logdet

willtebbutt:
Fixing this makes e.g.`Zygote` "just work" with`MvNormal` and `GenericMvT` in `Distributions` by preventing `ChainRules` from taking control of the way that `AbstractPDMats` compute `logdet`.


willtebbutt:
Zygote integration failure is unrelated to this change.

oxinabox:
Not a fan of the existance of `StridedMatrix`
Would rather if we could just say `Matrix`,
but we can't so approving it

Issue #232, opened 2020-07-07
Type Constraints


willtebbutt:
Various discussions have been had in various places about the correct kinds of types to implement `rrule`s for, but we've not discussed this in a central location. This problem probably occurs for some `frule`s, but doesn't seem as prevalent as in the `rrule` case.

## Problem Statement

The general theme of the problem is whether or not to view certain types as being "embedded" inside others or not, for the purpose of computing derivatives. For example, is a `Diagonal` matrix one that just _happens_ to be diagonal and is equally-well thought of as a `Matrix`, or is it really it's own thing? Similarly, is an `Integer` just a `Real`, or is it something else?

I will first attempt to demonstrate the implications of each choice with a couple of representative examples.

### `Diagonal` matrix multiplication

Consider the following `rrule` implementation:
```julia
function rrule(::typeof(*), X::AbstractMatrix{<:Real}, Y::AbstractMatrix{<:Real})
    function mul_AbstractMatrix_pullback(ΔΩ::AbstractMatrix{<:Real})
        return ΔΩ * Y', X' * ΔΩ
    end
    return X * Y, mul_AbstractMatrix_pullback
end
```
If `X` and `Y` are `Matrix{Float64}`s for example, then this is a perfectly reasonable implementation -- `ΔΩ` should also be a `Matrix{Float64}` if whoever is calling this rule is calling it correctly.

Things break down if `X` is a `Diagonal{Float64}`. The forwards-pass is completely fine, as is the computation of the cotangent for `Y`, `X' * ΔΩ`. However, the complexity of the cotangent representation / computation for `X` is now very concerning -- `ΔΩ * Y'` produces a `Matrix{Float64}`. Such a matrix is specified by `O(N^2)` numbers rather than `O(N)` required for `X`, and requires `O(N^3)`-time to compute, as opposed to the forwards-pass complexity `O(N^2)`. This breaks the promise that the forwards- and reverse-pass time- and memory-complexity of reverse-mode AD should be the same, in essence rendering the structure in a `Diagonal` matrix if used in an AD system where this is the only rule for multiplication of matrices.

[...]

PR #246, opened 2020-08-18
Add basic sparse handling

DhairyaLGandhi:
Questions: 

- [ ] how do I test things?
- [ ] Will this be the right test here?
- [x] Also this adds an `rrule` for a constructor, is that a problem?


willtebbutt:
>  how do I test things?

See other examples in this package, e.g. [here](https://github.com/JuliaDiff/ChainRules.jl/blob/d55540a7c836e96f9f5e46cabe0b5e49d789587c/test/rulesets/LinearAlgebra/dense.jl#L9)

>  Will this be the right test here?

Not clear what you mean by this. Could you clarify please?

>  Also this adds an rrule for a constructor, is that a problem?

`rrule`s for constructors are fine :)

DhairyaLGandhi:
> Not clear what you mean by this. Could you clarify please?
I feel like I'm using the API incorrectly, but none seemed to work

```julia
rrule_test(x -> sum(sparse(x)), rand(), (rand(3,3), rand()))
```
Should have, I thought, but maybe you could clarify

willtebbutt:
> rrule_test(x -> sum(sparse(x)), rand(), (rand(3,3), rand()))

> Should have, I thought, but maybe you could clarify

The first element of the last argument is the value at which you want to evaluate the `rrule`, call it `x`. That looks like it's fine in your case. The second element should be an appropriate differential for `x`, call it `dx`. Since `x` is a `Matrix{Float64}`, `dx` should be a `Matrix{Float64}` of the same size.

As regards the first argument, `rrule_test` literally only handles `rrule`s, and can't handle their composition with other things. Consequently, if you want to test `sparse`, you should only pass in `sparse`. The reason it can't handle composition with other functions is because you would need an AD to do this, and `ChainRules` is AD-agnostic. Call the first argument `f`.

Finally, the second argument has to be an appropriate differential for `f(x)`. In this case, this means that you need to define what it means to be an appropriate differential for a `SparseMatrixCSC{Float64,Int64}`. I'm guessing that some arbitrary `AbstractMatrix{Float64}` is _not_ generally appropriate because it will destroy the complexity guarantees associated with sparse arrays.

On that note, it might be worth reading #232 if you've not already -- there's a large discussion going on about the "right" way to think about this, which might be helpful. It would be great if you could chime in with your thoughts on the matter.

DhairyaLGandhi:
> The reason it can't handle composition with other functions is because you would need an AD to do this, and ChainRules is AD-agnostic

That would explain why `sparse` wouldn't work either. It would need to test `SparseMatrixCSC` directly, since there is no AD recursion into the stacks happening either. Thanks for pointing me to the issue, I do think having a robust testing pipeline is necessary for the different levels of abstractions, and this makes sense for ChainRules.

willtebbutt:
> That would explain why sparse wouldn't work either. It would need to test SparseMatrixCSC directly, since there is no AD recursion into the stacks happening either.

That's not quite true. IIUC you've implemented an rrule for `sparse` directly, so you _should_ be able to use CRTU (ChainRulesTestUtils) on it. The question is just what the differential of the output should be.

> Thanks for pointing me to the issue, I do think having a robust testing pipeline is necessary for the different levels of abstractions, and this makes sense for ChainRules.

Agreed. The issue in question has signficant ramifications for Zygote as well though -- there are currently loads of rules implemented there that are causing the kinds of problems discussed in that issue. But that's a separate issue.

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
+using SparseArrays
+
+function rrule(::Type{<:SparseMatrixCSC{T,N}}, arr) where {T,N}
```
nickrobinson251, line 3: we try to match the file structure of Base / stdlibs (althought we're not yet entire consistent on that). Since `SparseMatrixCSC` is defined in `SparseArrays/src/sparsematrix.jl`, can you please rename this file from `rulesets/SparseArrays/array.jl` -> `rulesets/SparseArrays/sparsematrix.jl`

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
...
+  function SparseMatrix_pullback(Δ)
```
nickrobinson251, line 4: please can you use 4-space indenting throughout, as per the style guide? https://github.com/invenia/BlueStyle#synopsis :)

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
...
+    (collect(Δ),)
```
nickrobinson251, line 5: Please follow [BlueStyle in always using a `return` keyword on long-form functions](https://github.com/invenia/BlueStyle#method-definitions)

Also, the first return should be the "derivative with respect to the fields of the function itself" https://www.juliadiff.org/ChainRulesCore.jl/dev/#Self-derivative-%CE%94self,-self,-self,-%E1%B9%A1elf-etc
```suggestion
        return NO_FIELDS, collect(Δ)
```
(I think `NO_FIELDS` is fine here for now?)

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
...
+  end
+  SparseMatrixCSC{T,N}(arr), SparseMatrix_pullback
```
nickrobinson251, line 7:
```suggestion
    return SparseMatrixCSC{T,N}(arr), SparseMatrix_pullback
```

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
...
+end
+
+function rrlue(::typeof(diagm), x)
+  diagm(x), d -> (diag(d),)
+end
```
nickrobinson251, line 12: `diagm` is defined in LinearAlegebra stdlib, not SpareArrays, so rules for it should go in `rulesets/LinearAlgebra/dense.jl`. Since there are other `rrules` for `diagm`, I think we should probably restrict the input to `v::AbstractVector`?

also, please can you change this to follow the ChainRules style of using named local functions for the pullback in an rrule?
https://www.juliadiff.org/ChainRulesCore.jl/dev/writing_good_rules.html#Code-Style

src/rulesets/SparseArrays/array.jl
```diff
@@ -0,0 +1,16 @@
...
+
+function rrlue(::typeof(issymmetric), x)
+  issymmetric(x), d -> (d,)
+end
```
nickrobinson251, line 16: similar comments to above: please move to LinearAlgebra, add type constraints and follow the style guide :)

nickrobinson251 (commented):
thanks for the contribution!

I think @willtebbutt is helping out with testing. i've left some on the `src` 👍

src/rulesets/SparseArrays/array.jl
DhairyaLGandhi: Done

src/rulesets/SparseArrays/array.jl
DhairyaLGandhi: Done

src/rulesets/SparseArrays/array.jl
DhairyaLGandhi: I've moved things, good to get another pair of eyes over it

sethaxen:
I have a current project for which I need some `rrule`s for `SparseMatrixCSC`. @DhairyaLGandhi are you planning to work on this in the near future? And if not, would you mind if I continued this work (in a new PR)?

DhairyaLGandhi:
Sure, please go ahead! Apologies I missed this comment in the GitHub blackhole. I need this for a project as well, so would love to get this through.

willtebbutt:
@sethaxen please ping me when ready for review.

src/rulesets/SparseArrays/sparsematrix.jl
```diff
@@ -0,0 +1,15 @@
+using SparseArrays
```
oxinabox, line 1: Not a full review, but just so it doesn't get missed, it is the pattern of this package, inline with BlueStyle,
to do all the `using`s at the top level file

CarloLucibello:
Just hit this problem in https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/113. 
Is there any interest in finishing up this PR soon (it has been open for a year and a half)? Otherwise, I'll create a new one

sethaxen:
@CarloLucibello i won't be able to work on this soon. Feel free to take over!

PR #249, opened 2020-08-20
Remove thunks on scalars

willtebbutt:
@dpsanders pointed out that our scalar multiplication implementation returns thunks. As a policy we don't thunk scalar operations for simple operations as the tradeoff isn't typically worth it -- these appeared to creep through.

I'll add a separate don't-thunk-by-default test to `ChainRulesTestUtils` to make sure that this doesn't creep back in, but still lets rule-implementers use thunk if they _really_ want to.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.13"
+version = "0.7.14"
 
 [deps]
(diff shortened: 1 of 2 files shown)


nickrobinson251:
Shall we also explicitly mention scalar operations in the docs on using Thunks? 

https://www.juliadiff.org/ChainRulesCore.jl/dev/writing_good_rules.html#Use-Thunks-appropriately

(Would be a separate or, to ChainRulesCore)

nickrobinson251 (approved):

PR #252, opened 2020-08-28
add nondiff rules

oxinabox:
Requires https://github.com/JuliaDiff/ChainRulesCore.jl/pull/207

@Keno  🎁 
Here is several hundred methods you don't have to AD any more.


oxinabox:
idk if these are worth testing.
Perhaps as long as the macro is well tested in ChainRulesCore, that is enough, and these don't need tests.
The main mistake that could be made with these is if the wrong things were included.
Which seems hard to test in a way that doesn't just have the same content as the code

src/rulesets/Base/nondiff.jl
```diff
@@ -0,0 +1,358 @@
...
+@non_differentiable Sys.isdragonfly(::Symbol)
+@non_differentiable Sys.isexecutable(::AbstractString)
+@non_differentiable Sys.isfreebsd(::Symbol)
+@non_differentiable Sys.isjsvm(::Symbol)
+@non_differentiable Sys.islinux(::Symbol)
+@non_differentiable Sys.isnetbsd(::Symbol)
+@non_differentiable Sys.isopenbsd(::Symbol)
+@non_differentiable Sys.isunix(::Symbol)
+@non_differentiable Sys.iswindows(::Symbol)
+@non_differentiable Sys.which(::AbstractString)
```
willtebbutt, line 358:
```suggestion
@non_differentiable Sys.which(::AbstractString)

```

src/rulesets/Random/random.jl
```diff
@@ -6,3 +6,16 @@ function rrule(::Type{MersenneTwister}, args...)
     end
     return MersenneTwister(args...), MersenneTwister_pullback
 end
+
+@non_differentiable Broadcast.broadcastable(::AbstractRNG)
+
+@non_differentiable Random.randexp(::AbstractRNG)
+@non_differentiable Random.randstring(::AbstractRNG)
+
```
willtebbutt, line 14:
```suggestion

@non_differentiable rand()
```

src/rulesets/Random/random.jl
```diff
@@ -6,3 +6,16 @@ function rrule(::Type{MersenneTwister}, args...)
...
+@non_differentiable rand(::AbstractRNG)
+@non_differentiable randn(::AbstractRNG)
```
willtebbutt, line 16:
```suggestion
@non_differentiable randn()
@non_differentiable randn(::AbstractRNG)
```

src/rulesets/Base/nondiff.jl
```diff
@@ -0,0 +1,358 @@
...
+
+@non_differentiable wait(::Base.Process)
+@non_differentiable wait(::Base.ProcessChain)
+@non_differentiable wait(::RawFD)
+@non_differentiable write(::IO, ::Any)
+@non_differentiable write(::IO, ::Any, ::Any,)
+@non_differentiable write(::IO, ::Any, ::Any, ::Any)
+@non_differentiable write(::IO, ::Any, ::Any, ::Any, ::Any)
+
+@non_differentiable Libc.free(::Any)
```
willtebbutt, line 341: Looks like `Libc` has quite a lot of methods that certainly don't appear differentiable that don't appear on this list. Is there a particular reason for the omission, or just because you've not had time to write them out yet?

willtebbutt (commented):
Very nice.

I'm not of the opinion that tests would be particularly helpful here. We'll probably want to add regression tests when we find issues with these rules / decide to add extra rules, but for now I think this is completely fine.

Things from `Base` `@nograd`ed in Zygote but not `@non_differentiable`d here:

- `Base.gc_num`
- `Base.time_ns`
- `Channel`
- `schedule`
- `floor`
- `ceil`
- `trunc`
- `round`
- `div`
- `eachindex`
- ` Base.OneTo`
- `axes`
- `Colon()`
- `ones`
- `zeros`
- `one`
- `zero`
- `all`
- `any`

Things in submodules:
- `Broadcast.combine_styles`
- `Broadcast.result_style`

These all look broadly correct to me. Maybe we want to implement them before merging this?

oxinabox:
I think the following we already have as returning `Zero`.

```
ones
zeros
one
zero
```

I am dubious about these:

- [ ] floor
- [ ] ceil
- [ ] trunc
- [ ] round
- [ ] div


We should definately have these:

- [ ] eachindex
- [ ] Base.OneTo
- [ ] axes
- [ ] Colon()
- [ ] all

Also i guess the other logical operators
- [ ]  `any`
- [ ] `!`
- [ ] `xor`

src/rulesets/Base/nondiff.jl
oxinabox: We only grabbed once that had particular types uses in their type signatures.
In particular we grabbed nothing with `Number` in the type-sig

willtebbutt:
> I am dubious about these:

Hmmm yeah, maybe. There are definitely some methods of them that involve clearly non-differentiable types that we can be pretty sure are `@non_differentiable`.

Assuming that no one is punning on them, I think I'm correct in saying that they all have zero derivative everywhere, except at the points where the functions change value. At those points the derivative is unbounded / undefined / whatever we're calling `1 / h` as `h -> 0`. So my opinion is that it's probably fine to `@non_differentiable` them.

src/rulesets/Base/nondiff.jl
willtebbutt: Oh interesting. Just looking through the methods there are definitely some good candidates e.g. `calloc`.

oxinabox:
Right but  `@non_differentiable` does `DoesNotExist` not `Zero()`

willtebbutt:
Good point.

oxinabox:
perhaps we can generalize the code.
Maybe in a follow up, or maybe now. 

Could be something like `@constant_rule f(x,y) --> DoesNotExist()` `@costant_rule f(x,y) --> Zero()`

willtebbutt:
Good point -- a constant rule would be helpful.

I vote that we get this in and then add a `@constant_rule` construct of some kind.

src/rulesets/Random/random.jl
oxinabox: This has no inputs.
Isn't it implict that it is not differentiable?

willtebbutt: Kind of. An AD system would have to figure out that `rand` has no internal data and deduce that it can ignore it. This way if the AD system isn't very good at doing that things are still okay.

oxinabox: I think I would like to cover all function without any args in a seperate PR.

oxinabox: But I guess we can do this one since you've written it now and bad things happen when you bring your AD too close to random numbers

src/rulesets/Base/nondiff.jl
willtebbutt: What do you want to do about this one? I'm happy to leave for future work for the sake of getting this in provided that we document what we _think_ is missing in an issue somewhere.

oxinabox:
Other zyogte ones:

- [ ] Base.gc_num
- [ ] Base.time_ns
- [ ] Channel
- [ ] schedule

src/rulesets/Random/random.jl
```diff
@@ -6,3 +6,51 @@ function rrule(::Type{MersenneTwister}, args...)
...
+@non_differentiable copy(::Random._GLOBAL_RNG)
+@non_differentiable copy(::MersenneTwister)
+@non_differentiable copy!(::MersenneTwister, ::MersenneTwister)
+@non_differentiable copy!(::MersenneTwister, ::Random._GLOBAL_RNG)
+@non_differentiable copy!(::Random._GLOBAL_RNG, ::MersenneTwister)
+
+@static if VERSION > v"1.3"
+    @non_differentiable Random.default_rng()
+    @non_differentiable Random.default_rng(::Int)
+  end
```
nickrobinson251, line 56:
```suggestion
end
```

willtebbutt (approved):
LGTM!

PR #255, opened 2020-08-29
add Base.sinc rule

stevengj:
I recently became aware that Julia `Base` defines a `cosc` function to be the derivative of `sinc`, so it seems worth adding this rule (especially since naive differentiation of `sinc` leads to an implementation that is numerically unstable: https://github.com/JuliaLang/julia/issues/37227).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.14"
+version = "0.7.15"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


nickrobinson251:
Thanks! please can you add a test (e.g. to https://github.com/JuliaDiff/ChainRules.jl/blob/bb08bd80b5cee92947002547de1f54556a5e3ee7/test/rulesets/Base/base.jl#L2) and bump the Project.toml version number :)

stevengj:
I added a test.  Since you are pre-1.0, how do you want to increment the version number.  Bump the patch version?

nickrobinson251:
> Bump the patch version?

yes :D Thanks!

stevengj:
Test failure seems unrelated?
```
tuple getindex: Error During Test at /home/runner/work/ChainRules.jl/ChainRules.jl/Zygote.jl/test/features.jl:318
  Test threw exception
  Expression: gradient((x->begin
                sum((size(x))[1:2:3])
            end), ones(2, 2, 2, 2)) == (nothing,)
  Mutating arrays is not supported
```

nickrobinson251:
> Test failure seems unrelated?

Agreed - seems like Zygote master is broken.

Thanks for this! LGTM.

nickrobinson251 (approved):

Issue #257, opened 2020-09-03
New Non-Diff Rules can cause Zygote to change keyword arguments


ChrisRackauckas:
MWE:

```julia
using Zygote

struct Type1{VJP}
  x::VJP
end

struct Type2{compile}
  Type2(compile=false) = new{compile}()
end

function loss_adjoint(θ)
  sum(f(sensealg=Type1(Type2(true))))
end
function f(;sensealg=nothing)
  @show "entry",sensealg
  return rand(100)
end

loss_adjoint([1.0])
Zygote.gradient(loss_adjoint,[1.0])
```

Before:

```
("entry", Type1{Type2{true}}(Type2{true}()))
("entry", Type1{Type2{true}}(Type2{true}()))
```

After:

```
("entry", Type1{Type2{true}}(Type2{true}()))
("entry", Base.OneTo(true))
```

I'm going to search for that rule.

oxinabox:
We know it is ChainRules  0.7.16 doing it because going back to 0.7.15 fixes it.
But its super weird since `@non_differentiable` doesn't even support kwargs.
(Probably something that should be changed, it should accept them as `kwargs...` and then pass them on)

oxinabox:
Underlying cause is:
https://github.com/JuliaDiff/ChainRulesCore.jl/issues/213

PR #258, opened 2020-09-03
Don't use non-differentiation  on any constructors

ChrisRackauckas:
I don't really know what to call this, but it fixes https://github.com/JuliaDiff/ChainRules.jl/issues/257 , whatever that is.


oxinabox (approved):

oxinabox:
Merged and tag when ever you are ready.
You have merge rights on JuliaDiff IIRC.

PR #260, opened 2020-09-03
delete non_diff Base.OneTo

### LINKED ISSUE #257: New Non-Diff Rules can cause Zygote to change keyword arguments

oxinabox:
@ChrisRackauckas  can you review, merge and tag when ready?
(It is late here and i am going to bed, so making this PR)

Fixes #257

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.16"
+version = "0.7.17"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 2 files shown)


oxinabox:
NEvermind https://github.com/JuliaDiff/ChainRules.jl/pull/258/
did get made and i just didn't see it

Issue #262, opened 2020-09-08
@non_differentiable rand


nmheim:
Really happy that ChainRules now has a `@non_differentiable` macro! Unfortunately it includes all `rand` methods, which is a bit too restrictive... there was a similar discussion about this in https://github.com/FluxML/Zygote.jl/pull/234 .

MWE:
```julia
using Random
using Zygote

struct Normal
    μ
    σ
end

Random.rand(d::Normal) = randn()*d.σ + d.μ
f(a,b) = rand(Normal(a,b))
gs = Zygote.gradient(f,2,2) # returns (nothing,nothing) with ChainRules > 0.7.15
```

willtebbutt:
Lol, yeah, we really shouldn't have done that 🤦 It was literally _the_ motivating for a more flexible `@non_differentiable` / `@nograd` macro.

A PR modifying the implementation would be very welcome.

nmheim:
Thanks for merging this so quickly!

PR #263, opened 2020-09-08
More restrictive @non_differentiable rand

nmheim:
Addresses #262. I am really not sure if this is a good approach because there are quite a few different `rand` methods...
This is what works with the use cases I have. What are the most common uses of `rand` that should be non-differentiable?


oxinabox:
Why not just add:
```
@non_differentiable rand()
@non_differentiable rand(::Any, ::Any, ::Any, ::Any, ::Any, ::Any)
```
?

willtebbutt:
@oxinabox see #262?

oxinabox:
Oh i was confused by the title.

src/rulesets/Random/random.jl
```diff
@@ -12,11 +12,19 @@ end
...
+@non_differentiable rand(::AbstractRNG, ::Int)
+@non_differentiable rand(::AbstractRNG, ::Int, ::Int)
+@non_differentiable rand(::AbstractRNG, ::Int, ::Int, ::Int)
+@non_differentiable rand(::AbstractRNG, ::Int, ::Int, ::Int, ::Int)
+@non_differentiable rand(::AbstractRNG, ::Int, ::Int, ::Int, ::Int, ::Int)
+@non_differentiable rand(::Int)
+@non_differentiable rand(::Int, ::Int)
+@non_differentiable rand(::Int, ::Int, ::Int)
+@non_differentiable rand(::Int, ::Int, ::Int, ::Int)
+@non_differentiable rand(::Int, ::Int, ::Int, ::Int, ::Int)
```
oxinabox, line 26: I think all the `Int` should be `Integer` 
to match:

```
[69] rand(X, d::Integer, dims::Integer...) in Random at /usr/local/src/julia/julia-1.5/usr/share/julia/stdlib/v1.5/Random/src/Random.jl:283
```

oxinabox:
Can we add a test that makes sure
`rrule` and `frule`  for something that we don't want to have mark as `@non_differentiable` returns `nothing` (i.e. no rule)

oxinabox:
I think we might also in particular want:
```julia
@non_differentiable rand(::AbstractRNG, ::Random.Sampler)
@non_differentiable rand(::AbstractRNG, ::Tuple)
```

nmheim:
Let me know what you think about my take on testing the new rules. I am really not sure if this is how it should be done. Especially because the `rrule`s and `frule`s return `Zero` or `DoesNotExist`...

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
+        for (args,xType) in non_differentiables
+            x, dΩ = frule((), rand, args...)
+            @test x isa xType
+            @test dΩ isa DoesNotExist
+
+            x, pb = rrule(rand, args...)
+            @test x isa xType
+            @test first(pb(10)) isa Zero
+        end
+
```
oxinabox, line 41:
```suggestion

        # Make sure that we do *not* have these set as non_differentiable. as they are differentiable
```

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
+        non_differentiables = [((), Float64),
+                               ((Random.default_rng(),), Float64),
+                               ((Random.default_rng(),2,2), Matrix{<:Float64}),
+                               ((Float32,), Float32),
+                               ((Float32,2,2), Matrix{<:Float32}),
+                               ((Float32,(2,2)), Matrix{<:Float32}),
+                               ((2,2), Matrix{<:Float64})]
+
+        for (args,xType) in non_differentiables
+            x, dΩ = frule((), rand, args...)
```
oxinabox, line 33: Lets make it a realistic call to `frule`
```suggestion
            x, dΩ = frule((Zero(), 1.2.*ones(args)...), rand, args...)
```

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
+            @test x isa xType
+            @test dΩ isa DoesNotExist
+
+            x, pb = rrule(rand, args...)
+            @test x isa xType
+            @test first(pb(10)) isa Zero
```
oxinabox, line 39: Isn't it important to test not the first partial, but rather all the others?
```suggestion
            dself, dargs = Iterators.peel(pb(10.0))
            @test dself isa Zero
            @test all(darg isa DoesNotExist for darg in dargs)
```

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
+        end
+
+        @test frule((), rand, ones(2,2)) === nothing
```
oxinabox, line 42: make it a realistic all to `frule`
```suggestion
        @test frule((Zero(), 1.3*ones(2,2)), rand, ones(2,2)) === nothing
```

oxinabox (approved):
Nice.
A few suggestions,
once those are addressed and the version number bumped we can merge this and tag a release

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
     end
+
+    @testset "rand" begin
+        non_differentiables = [((), Float64),
+                               ((Random.default_rng(),), Float64),
+                               ((Random.default_rng(),2,2), Matrix{<:Float64}),
+                               ((Float32,), Float32),
+                               ((Float32,2,2), Matrix{<:Float32}),
+                               ((Float32,(2,2)), Matrix{<:Float32}),
+                               ((2,2), Matrix{<:Float64})]
```
willtebbutt: Just a style thing -- not sure that BlueStyle addresses this particular construction explicitly, but below is "morally" the kind of thing that the style guide expects

```suggestion
        non_differentiables = [
            ((), Float64),
            ((Random.default_rng(),), Float64),
            ((Random.default_rng(),2,2), Matrix{<:Float64}),
            ((Float32,), Float32),
            ((Float32,2,2), Matrix{<:Float32}),
            ((Float32,(2,2)), Matrix{<:Float32}),
            ((2,2), Matrix{<:Float64}),
        ]
```

test/rulesets/Random/random.jl
```diff
@@ -19,4 +19,27 @@
...
+
+        for (args,xType) in non_differentiables
```
willtebbutt, line 32:
```suggestion
        for (args, xType) in non_differentiables
```

willtebbutt (approved):
Just a couple of style things.

Once @oxinabox 's suggestions and these are attended to, please bump the patch version and we'll merge when tests pass :)

Thanks for the contribution -- it's much appreciated!

nmheim:
Great, thanks for the suggestions! I think the tests are failing because of something unrelated to this PR though... they also fail in #261 because of something `getindex` related...

test/rulesets/Random/random.jl
nmheim: yes, makes sense! still learning how everything works here :)

willtebbutt:
> Great, thanks for the suggestions! I think the tests are failing because of something unrelated to this PR though... they also fail in #261 because of something getindex related...

Oh, yeah, don't worry about the Zygote integration tests -- Zygote master is currently broken.

nmheim:
Actually, I am not sure about the comment on the test of the "differentiable" cases.
I used `rand(ones(2,2))` as a test because we don't need an additional dependency like this, but I don't know why `rand` of an array works at all...
A better test case would be to use e.g. a distribution. Should I add that?

oxinabox:
> A better test case would be to use e.g. a distribution. Should I add that?

Yes

nmheim (commented):
I am sorry, but I am a bit clueless how to test the rules on fields of a struct if they don't appear in the `rand` function call... this is probably not how it should be done...

test/rulesets/Random/random.jl
```diff
@@ -1,3 +1,9 @@
+struct Normal
```
oxinabox, line 1: Maybe since this is not in a package called distbrutions we can be more explictit what it is?
```suggestion
# Simple Distributions like object for testing purposes
struct NormalDistribution
```

test/rulesets/Random/random.jl
```diff
@@ -1,3 +1,9 @@
...
+    μ
+    σ
+end
+Random.rand(d::Normal) = d.μ + d.σ*randn()
```
oxinabox, line 5:
```suggestion
Random.rand(d::NormalDistribution) = d.μ + d.σ*randn()
```

test/rulesets/Random/random.jl
```diff
@@ -44,11 +50,6 @@
         end
 
         # Make sure that we do *not* have these set as non_differentiable. as they are differentiable
-        struct Normal
-            μ
-            σ
-        end
-        Random.rand(d::Normal) = d.μ + d.σ*randn()
         @test frule((Zero(), Normal(0.5,2.0)), rand, Normal(0.1,1.5)) === nothing
```
oxinabox, line 53: The differentials for types tht don't define `+` and `zero` etc is `Composite`
```suggestion
        @test nothing === frule(
            (Zero(), Composite{NormalDistribution}(μ=0.5,σ=2.0),
            rand,
            NormalDistribution(0.1,1.5),
        )
```

test/rulesets/Random/random.jl
```diff
@@ -44,11 +50,6 @@
...
         @test rrule(rand, Normal(0.1,1.5)) === nothing
```
oxinabox, line 54:
```suggestion
        @test rrule(rand, NormalDistribution(0.1,1.5)) === nothing
```

oxinabox (approved):
I am going to add my suggestions and then if i didn't break CI merge this

test/rulesets/Random/random.jl
```diff
@@ -50,7 +51,11 @@ Random.rand(d::Normal) = d.μ + d.σ*randn()
         end
 
         # Make sure that we do *not* have these set as non_differentiable. as they are differentiable
-        @test frule((Zero(), Normal(0.5,2.0)), rand, Normal(0.1,1.5)) === nothing
-        @test rrule(rand, Normal(0.1,1.5)) === nothing
+        @test nothing === frule(
+            (Zero(), Composite{NormalDistribution}(μ=0.5,σ=2.0),
```
oxinabox, line 55:
```suggestion
            (Zero(), Composite{NormalDistribution}(μ=0.5,σ=2.0)),
```

nickrobinson251 (approved):

nmheim:
Awesome, thank you! :)

PR #266, opened 2020-09-15
Correct rrule for reshape

oxinabox:
Any time you test your rules without using finite differencing, you might as well not be testing your rules

### DIFF
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -3,15 +3,17 @@
 #####
 
 function rrule(::typeof(reshape), A::AbstractArray, dims::Tuple{Vararg{Int}})
+    A_dims = size(A)
     function reshape_pullback(Ȳ)
-        return (NO_FIELDS, reshape(Ȳ, dims), DoesNotExist())
+        return (NO_FIELDS, reshape(Ȳ, A_dims), DoesNotExist())
     end
     return reshape(A, dims), reshape_pullback
 end
 
 function rrule(::typeof(reshape), A::AbstractArray, dims::Int...)
+    A_dims = size(A)
     function reshape_pullback(Ȳ)
-        ∂A = reshape(Ȳ, dims)
+        ∂A = reshape(Ȳ, A_dims)
         return (NO_FIELDS, ∂A, fill(DoesNotExist(), length(dims))...)
     end
     return reshape(A, dims...), reshape_pullback
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.18"
+version = "0.7.19"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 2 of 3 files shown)


src/rulesets/Base/array.jl
```diff
@@ -3,15 +3,17 @@
...
+        return (NO_FIELDS, reshape(Ȳ, A_dims), DoesNotExist())
     end
     return reshape(A, dims), reshape_pullback
 end
 
 function rrule(::typeof(reshape), A::AbstractArray, dims::Int...)
+    A_dims = size(A)
     function reshape_pullback(Ȳ)
-        ∂A = reshape(Ȳ, dims)
+        ∂A = reshape(Ȳ, A_dims)
```
oxinabox, line 16: rather than close over all of `A` we just close over its size.
lets A be GCed

mzgubic: As far as I understand the issue is that the `reshape(Y, dims)` would not do anything because the output array `Y` is already shaped as `dims`, rather than the original shape of array `A`.

I don't understand your comment about closing and what GC means. Could you please explain or link to relevant wiki page?

oxinabox: > As far as I understand the issue is that the reshape(Y, dims) would not do anything because the output array Y is already shaped as dims, rather than the original shape of array A. (Essentially needs to reshape back to the original shape, but doesn't)

Correct.
I think by mistake the `frule` for rshape was written.

------

Closures construct callable structs  (functors),
with a field for each thing in partent scope they reference.
We call the things in parent scope that it references the _variables that it closes over_.
we say it is a _closures over those variables_.

So rather than closing over `A` we close over `A_dims`.
The fields of the closure are not copies, just references, so its not too bad.
But the garbage collector can't run until there are no more references to the thing.


Compare:
```julia
julia> function mkclosure()
       X = ones(5,5)
       pb() = (size(X) .+ 1)
       return pb
       end
mkclosure (generic function with 1 method)

julia> _pb = mkclosure()
(::var"#pb#13"{Array{Float64,2}}) (generic function with 1 method)

julia> _pb()
(6, 6)

julia> _pb.X
5×5 Array{Float64,2}:
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0  1.0

julia> fieldnames(typeof(_pb))
(:X,)

julia> Base.summarysize(_pb)
248
```

Vs:

```julia
julia> function mkclosure2()
       X = ones(5,5)
       X_dims = size(X)
       pb() = (X_dims .+ 1)
       return pb
       end
mkclosure2 (generic function with 1 method)

julia> _pb = mkclosure2()
(::var"#pb#15"{Tuple{Int64,Int64}}) (generic function with 1 method)

julia> _pb()
(6, 6)

julia> _pb.X_dims
(5, 5)

julia> fieldnames(typeof(_pb))
(:X_dims,)

julia> Base.summarysize(_pb)
16
```

mzgubic: Aaahh I see the comment is referring to why it isn't just `size(A)` inside the `pb`, rather than why the change is made at all.

mzgubic (approved):

src/rulesets/Base/array.jl
oxinabox: correct

PR #267, opened 2020-09-16
Make sum of bool arrays nondiffernetiable

oxinabox:
This matches to https://github.com/FluxML/Zygote.jl/pull/794

Fixing this in chainrules won't actually fix it in Zygote becuase 
Zygote defines `sum(::AbstractArray)` with a better definition than what ChainRules has right now, that will always take precidence.

### DIFF
--- a/test/rulesets/Base/nondiff.jl
+++ b/test/rulesets/Base/nondiff.jl
@@ -1,0 +1,7 @@
+@testset "Base/nondiff.jl" begin
+    @testset "Sum boolean arrays" begin
+        rrule_test(sum, 12, (falses(5, 3), nothing))
+        rrule_test(sum, 2, ([true, false, true], nothing))
+        rrule_test(sum, 2, ((true, false, true), nothing))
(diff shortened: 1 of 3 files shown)


test/rulesets/Base/nondiff.jl
```diff
@@ -0,0 +1,7 @@
+@testset "Base/nondiff.jl" begin
```
willtebbutt, line 1:
```suggestion
@testset "nondiff.jl" begin
```
testset nesting handles this for you.

test/rulesets/Base/nondiff.jl
```diff
@@ -0,0 +1,7 @@
...
+    @testset "Sum boolean arrays" begin
```
willtebbutt, line 2:
```suggestion
    @testset "Sum over boolean array" begin
```

willtebbutt (approved):
LGTM modulo style and patch bump

PR #268, opened 2020-09-17
Add Array * Scalar

oxinabox:
Nabla has this, we should have it too.

I am sure it can be extended to work for complex but i don't know how so am happy to make it a follow up PR

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -149,6 +149,21 @@ function rrule(::typeof(*), A::AbstractMatrix{<:Real}, B::AbstractMatrix{<:Real}
     return A * B, times_pullback
 end
 
+function rrule(::typeof(*), A::Real, B::AbstractArray{<:Real})
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(dot(Ȳ, B)), @thunk(A' * Ȳ))
+    end
+    return A * B, times_pullback
+end
+
+function rrule(::typeof(*), B::AbstractArray{<:Real}, A::Real)
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(A' * Ȳ), @thunk(dot(Ȳ, B)))
+    end
+    return A * B, times_pullback
+end
+
+
 #####
 ##### `pinv`
 #####
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -151,14 +151,14 @@ end
 
 function rrule(::typeof(*), A::Real, B::AbstractArray{<:Real})
     function times_pullback(Ȳ)
-        return (NO_FIELDS, @thunk(dot(Ȳ, B)), @thunk(A' * Ȳ))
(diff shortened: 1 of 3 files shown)


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -149,6 +149,21 @@ function rrule(::typeof(*), A::AbstractMatrix{<:Real}, B::AbstractMatrix{<:Real}
     return A * B, times_pullback
 end
 
+function rrule(::typeof(*), A::Real, B::AbstractArray{<:Real})
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(dot(Ȳ, B)), @thunk(A' * Ȳ))
```
mzgubic, line 154: Why `A'` rather than just `A`?

test/rulesets/LinearAlgebra/dense.jl
```diff
@@ -115,13 +115,21 @@
...
     end
     @testset "*" begin
-        dims = [3,4,5]
-        for n in dims, m in dims, p in dims
-            n > 3 && n == m == p && continue  # don't need to test square case multiple times
-            A = randn(m, n)
-            B = randn(n, p)
-            Ȳ = randn(m, p)
-            rrule_test(*, Ȳ, (A, randn(m, n)), (B, randn(n, p)))
+        @testset "Matrix-Matrix" begin
```
mzgubic, line 118: Out of scope perhaps but are there tests for Vector*Matrix anywhere? `dot` is tested but I don't think `*` is

mzgubic (approved):
LGTM

src/rulesets/LinearAlgebra/dense.jl
oxinabox: Yeah fair, with this restricted to `Real` it is the same.

```suggestion
        return (NO_FIELDS, @thunk(dot(Ȳ, B)), @thunk(A * Ȳ))
```

test/rulesets/LinearAlgebra/dense.jl
oxinabox: I am not sure we actually have a matric vector multiply `rrule` defined.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -149,6 +149,21 @@ function rrule(::typeof(*), A::AbstractMatrix{<:Real}, B::AbstractMatrix{<:Real}
...
+    end
+    return A * B, times_pullback
+end
+
+function rrule(::typeof(*), B::AbstractArray{<:Real}, A::Real)
+    function times_pullback(Ȳ)
+        return (NO_FIELDS, @thunk(A' * Ȳ), @thunk(dot(Ȳ, B)))
```
oxinabox, line 161:
```suggestion
        return (NO_FIELDS, @thunk(A * Ȳ), @thunk(dot(Ȳ, B)))
```

test/rulesets/LinearAlgebra/dense.jl
mzgubic: cool I would be keen to tackle this

PR #269, opened 2020-09-23
add `ignore` from Zygote to support Zygote-like AD's

### LINKED ISSUE #256: ChainRules.ignore
[Zygote.ignore](https://fluxml.ai/Zygote.jl/dev/utils/#Zygote.ignore) is a lovely piece of functionality that we aught to be able to provide at the `ChainRulesCore`-level. It would also be _incredibly_ straightforward to implement.

It'll only work properly with Zygote-like ADs ofc, but it's an improvement over it not working at all.

gxyd:
Fixes #256


willtebbutt:
Thanks for opening the PR.

To implement this, it shouldn't be necessary to utilise Zygote -- you should just have to define an appropriate `rrule` for `ignore` -- Zygote knows how to work with `rrule`s :)

gxyd:
So basically, Zygote when finding pullbacks as something like:
```jl
julia> x = π/4
julia> Zygote.pullback(x -> sin(x), π/4)
(0.7071067811865475, Zygote.var"#41#42"{typeof(∂(#3))}(∂(#3)))
```
Actually calls the `rrule` (probably apart from other performance stuff) of either ZygoteRules or ChainRules (in order), and tries to see if it can find the defined `rrule`, and if neither has it defined, then `nothing` is returned?

willtebbutt:
> Actually calls the rrule (probably apart from other performance stuff) of either ZygoteRules or ChainRules (in order), and tries to see if it can find the defined rrule, and if neither has it defined, then nothing is returned?

Exactly.

Rules created using `Zygote.@adjoint` take precedence over `ChainRules.rrule`s, but the goal is to move as many rules as possible from being `Zygote.@adjoint`s to `ChainRules.rrule`s.

(`Zygote.@adjoint` remains necessary for rules that call back into `Zygote` (e.g. `map`, `broadcast` etc) as we've not sorted out how higher-order rules work yet)

gxyd:
Is it possible to explicitly use ChainRules `frule` and `rrule` instead of using ZygoteRules, when things are called in from Zygote?

I was thinking of understanding the things a little better, I tried the below in Debugger:
```jl
julia> using Zygote
[ Info: Precompiling Zygote [e88e6eb3-aa80-5325-afca-941959d7151f]
^[^[
julia> using Debugger

julia> Debugger.@enter gradient(x -> sum([x x+1; x^2+x x]), 1)
In gradient(f, args) at /Users/gaurav/.julia/dev/Zygote/src/compiler/interface.jl:53
 52  function gradient(f, args...)
>53    y, back = pullback(f, args...)
 54    return back(sensitivity(y))
 55  end

About to run: (tuple)(var"#3#4"())
1|debug> s
In gradient(f, args) at /Users/gaurav/.julia/dev/Zygote/src/compiler/interface.jl:53
 52  function gradient(f, args...)
>53    y, back = pullback(f, args...)
 54    return back(sensitivity(y))
 55  end

About to run: (Core._apply_iterate)(iterate, ZygoteRules.pullback, (var"#3#4"(),), (1,))
```

My main objective is to see how the things are passed down to ChainRules from Zygote, but as we can see above it calls ZygoteRules (as expected), can I make it to explicitly pass it down to `ChainRules`.

oxinabox:
`ZygoteRules.pullback` eventually calls `ChainRules.rrule` (if one exists), unless it hits something defined by  `ZygoteRules.@adjoint` first.

gxyd:
> To implement this, it shouldn't be necessary to utilise Zygote -- you should just have to define an appropriate rrule for ignore -- Zygote knows how to work with rrules :)

But to invoke the correct `rrule` we would need to have `::typeof(ignore)` in function signature, and that would raise `UndefVarError` for `ignore` without `using Zygote`. How would we need to solve that?

willtebbutt:
> But to invoke the correct rrule we would need to have ::typeof(ignore) in function signature, and that would raise UndefVarError for ignore without using Zygote. How would we need to solve that?

The way to solve that is to define `ignore` inside `ChainRules`, remove `Zygote`s `ignore` function, and export `ChainRules.ignore` from `Zygote` to ensure that `Zygote`s external API remains unchanged :)

gxyd:
I still need to add test cases for it, but I couldn't find the definition of function `rrule_test` (or `frule_test`) in ChainRules. As I can see from other test cases [like this](https://github.com/JuliaDiff/ChainRules.jl/blob/master/test/rulesets/LinearAlgebra/blas.jl#L9), that I'll need to use `rrule_test`.

gxyd:
I think this is ready for review now, let me know the changes required. I'll try to fix them.

src/utils.jl
```diff
@@ -0,0 +1,46 @@
...
+julia> f(x) = x
+julia> _, v_pullback = ChainRules.rrule(ChainRules.ignore, f, (1,)...)
+julia> _, v̄ = v_pullback(1)
+julia> v̄
+nothing
+```
+"""
+
+function frule((_, ḟ), ::typeof(ignore), f, args...; kwargs...)
+    return f(args...; kwargs...), nothing
```
willtebbutt, line 31: This should return a `Zero` rather than `nothing` -- `Zygote` uses `nothing` to represent derivatives that are zero, while `ChainRules` has a collection of specific `AbstractZero` types.

src/utils.jl
```diff
@@ -0,0 +1,46 @@
...
+end
+
+function rrule(::typeof(ignore), f, args...; kwargs...)
+    y = f(args...; kwargs...)
+    function ignore_pullback(ȳ)
+        return (NO_FIELDS, nothing)
```
willtebbutt, line 37: Same here. Additionally, we need one `Zero` per-argument. @oxinabox I always forget what the most efficient way to generate a collection of `Zero`s of the correct size is likely to be -- could you adivse?

willtebbutt (commented):
This is definitely getting there! I think there's just some issues surrounding the conventions.

Also, it dawned on me that we probably want to include this functionality in `ChainRulesCore` rather than `ChainRules` because we're expecting users to want to use `ignore` and `@ignore`, and users should only have to load `ChainRulesCore`. Could you move this PR over to that repo?

src/utils.jl
gxyd: Would that be simply:
```jl
julia> zeros(2, 2)
2×2 Array{Float64,2}:
 0.0  0.0
 0.0  0.0
```
i.e. `zeros(length(args), length(args))`?

willtebbutt: Ah, wait, no. I misspoke.

Since the only method of `ignore` is unary, this `rrule` should also be unary, and doesn't need to have any `kwargs`. For example, you could simply write it as
```julia
function rrule(::typeof(ignore), f)
    function ignore_pullback(ȳ)
        return (NO_FIELDS, Zero())
    end
    return f(), ignore_pullback
end
```

Note the use of `Zero()` as the sensitivity of `f`, rather than `nothing`.

src/utils.jl
willtebbutt, line 31: Additionally, since `f` is unary, this `frule` doesn't need `args` and `kwargs`:
```suggestion
function frule((_, ḟ), ::typeof(ignore), f)
    return f(), nothing
```

gxyd:
Closing this in favor of https://github.com/JuliaDiff/ChainRulesCore.jl/pull/229

oxinabox:
Sorry for you to be caught in the cross-fire here, but I am in disagreement with @willtebbutt  on this belonging in ChainRulesCore.


> Also, it dawned on me that we probably want to include this functionality in ChainRulesCore rather than ChainRules because we're expecting users to want to use ignore and @ignore, and users should only have to load ChainRulesCore. Could you move this PR over to that repo?


I still think the same as i said in: https://github.com/JuliaDiff/ChainRules.jl/issues/256#issuecomment-685618537

> think I would want it in ChainRules.jl because its not needed for defining rules.
>
> Also it will break the on_new_rule functionality if we have any rules in ChainRulesCore other than the fallback

People defining rules, i.e. package authors, should only need to load ChainRulesCore.jl.
This keeps ChainRulesCore light, which is important for people defining rules that might not be used, e.g. if the end user isn't even doing AD.
People building AD systems should load ChainRules.jl.

End-users of AD systems, thus already have ChainRules.jl loaded.
End-users of AD systems shouldn't really have to directly depend on using ChainRules or ChainRulesCore;
they should just depend on their AD system.
The AD system probably should just be reexporting `ignore`

Also, it really will break ChainRulesCore if you define an `rrule` or a `frule` inside of it.

gxyd:
I'll make the necessary changes. 
> Sorry for you to be caught in the cross-fire here

Actually this might be a good thing in the hindsight, since you mentioned a few mentioned a points which clarify quite a few things for me :)

willtebbutt:
> I still think the same as i said in: #256 (comment)

Firstly, apologies @oxinabox -- I forgot we had discussed this already.

While I agree with your above points, they miss one category of developers: **package authors**. They generally have the following (fairly demanding) desiderata:
1. is compatible with any ChainRules-based AD
2. can use `ignore` (in some form or another) internally
3. users of the package that don't need AD should not have to load / depend on ChainRules (e.g. ones who are happy with gradient-free optimisers, which can definitely happen)
4. does not want to commit to a particular AD on behalf of their users

So my rationale for wanting `ignore`-like functionality in ChainRulesCore is so that the above can be satisfied.

One solution would be to only define an `@ignore` macro, and whenever it is called add a new `ignore_##XXXX` function to the method table and define an `rrule` and `frule` for it.

oxinabox:
Oh, I see where you are coming from now.
Something like a rule author doing:

```julia
function iterative_method(once_method, model, niters, args...)
    loss_record = []
    for ii 1:niters
        model, loss = once_method(model, args...)
        ignore() do
            push!(loss_record, loss)
            @info "Iterating" ii loss
        end
    end
    return model, loss_record
end
```

Has this come up in the wild?
If not, I am inclined to put this in ChainRules.jl
then move it to ChainRulesCore when and if it happens.

willtebbutt:
>  Something like a rule author doing:

This exactly.

> Has this come up in the wild?

Yup, came up during some of my recent work at Invenia -- there was some quite computationally demanding setup computation needed, but which we never need to interact with AD.

gxyd:
>  then move it to ChainRulesCore when and if it happens.

Shall I try and move this to `ChainRulesCore` then?

nickrobinson251:
from https://github.com/JuliaDiff/ChainRules.jl/issues/256

> it will break the `on_new_rule functionality` if we have any rules in ChainRulesCore other than the fallback

i think not breaking the `on_new_rule` functionality is crucial... so to put this in CRCore we'd need a separate PR that also makes the `on_new_rule functionality` robust to their being more than the fallback frule/rrule methods defined in CRCore. I imagine that's not a good-first-PR kinda task. 

If no one is available to make that PR, i wonder if we can get most of the benefit of `ignore` by leaving it defined in ChainRules for now? (I open an issue about the the `on_new_rule` functionality and moving `ignore` to CRC)

oxinabox:
Closing in favour of doing this in CRC

PR #270, opened 2020-09-28
Add scalar-array / and \

oxinabox:
Needed for https://github.com/invenia/Nabla.jl/pull/189

I worked this one out with a white-board.
I'm not good at math though, but tests pass.
So I guess i got it right

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -163,6 +163,25 @@ function rrule(::typeof(*), B::AbstractArray{<:Real}, A::Real)
     return A * B, times_pullback
 end
 
+#####
+##### `\`, `/` matrix-scalar_rule
+
+function rrule(::typeof(/), A::AbstractArray{<:Real}, b::Real)
+    Y = A/b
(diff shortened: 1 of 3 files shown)


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -163,6 +163,25 @@ function rrule(::typeof(*), B::AbstractArray{<:Real}, A::Real)
     return A * B, times_pullback
 end
 
+#####
+##### `\`, `/` matrix-scalar_rule
+
+function rrule(::typeof(/), A::AbstractArray{<:Real}, b::Real)
```
willtebbutt, line 169: Are these maybe a bit too abstractly typed?

oxinabox: I don' t think so in this case.
But I also don't think this is the PR to be having the discussion about it in.
This matchs to `*` defined above.
I think changing `*` and changing this can happen at same time

willtebbutt: Fair enough.

willtebbutt (approved):

PR #271, opened 2020-09-29
More Base's Array math into the Base ruleset folder

oxinabox:
This has been grating on me for a while.
The rules for functions defined in `Base` belong in the `Base` folder.
But these had been handing in the `LinearAlgebra` folder.

This PR only moves the code.
No changes.
I am not bumping version because this doesn't need a release, its is invisible to the user


nickrobinson251 (approved):

oxinabox:
Seems i have messed up something while moving these

PR #272, opened 2020-09-30
fix typo in 4 arg gemm rrule

oxinabox:
This is a simple typo.
Clearly we have no tests for this rule.

This is blocking for https://github.com/invenia/Nabla.jl/pull/189


willtebbutt (approved):
Lol, yeah, might as well unblock you.

You should probably bump the patch as well though.

PR #273, opened 2020-10-02
Efficient Inplace Accumulating Strided MatMul

oxinabox:
These come out from Nabla,
but i think are wrong as they don't currently work, and were not actually being hit.
See https://github.com/invenia/Nabla.jl/issues/192

I think we need to unwrap the `Transpose` and `Adjoint` objects before passing to GEMM
and/or maybe consider the double transpose case carefully.

I may split this PR in two and just get the simple case that has no `Transposes` or `Adjoints` in first, since Nabla actually uses that.


oxinabox:
I think this can all be cone much easier now with 5 arg `mul!`,
though that requires Julia 1.3.
and is not in Compat.jl yet https://github.com/JliaLang/Compat.jl/issues/681

I think we might even be able to just replace the generic code for the rrule/frule of matmul with that,
and not need this special casing at all

test/runtests.jl
```diff
@@ -36,6 +36,7 @@ println("Testing ChainRules.jl")
 
         @testset "LinearAlgebra" begin
             include(joinpath("rulesets", "LinearAlgebra", "dense.jl"))
+            include(joinpath("rulesets", "LinearAlgebra", "stided.jl"))
```
dkarrasch, line 39:
```suggestion
            include(joinpath("rulesets", "LinearAlgebra", "strided.jl"))
```

oxinabox:
`mul!` its all the cases except 1,
and I have a PR open for that one.
https://github.com/JuliaLang/julia/pull/37930
and its not one we are too worries about really

test/rulesets/Base/arraymath.jl
```diff
@@ -6,26 +6,54 @@
...
-                n > 3 && n == m == p && continue
-                A = randn(m, n)
-                B = randn(n, p)
-                Ȳ = randn(m, p)
-                rrule_test(*, Ȳ, (A, randn(m, n)), (B, randn(n, p)))
-            end
+    # TODO Test on Complex once we have Complex support for all of these.
+    @testset "*: $T" for T in (Float64,)
+        ⋆(a) = round.(5*randn(T, a))  # Helper to generate nice random values
+        ⋆(a, b) = ⋆((a, b))  # matrix
```
oxinabox, line 12: I wanted an inplace op to use to define random matrixes of a given size so that i could make sure i was getting the sizes right.

test/rulesets/Base/arraymath.jl
```diff
@@ -6,26 +6,54 @@
...
-                # don't need to test square case multiple times
-                n > 3 && n == m == p && continue
-                A = randn(m, n)
-                B = randn(n, p)
-                Ȳ = randn(m, p)
-                rrule_test(*, Ȳ, (A, randn(m, n)), (B, randn(n, p)))
-            end
+    # TODO Test on Complex once we have Complex support for all of these.
+    @testset "*: $T" for T in (Float64,)
+        ⋆(a) = round.(5*randn(T, a))  # Helper to generate nice random values
```
oxinabox, line 11: We should do this more for testing.
Errors are much easier to see if everything is whole numbers mostly between -10 and 10

test/rulesets/Base/arraymath.jl
```diff
@@ -6,26 +6,54 @@
...
+
+        @testset "AbstractMatrix-AbstractMatrix" begin
+            dims = [2, 4, 5, 10]  # small matrixes have some special cases
+            @testset "n=$n, m=$m, p=$p" for n in dims, m in dims, p in dims
+                @testset "Array" begin
+                    rrule_test(*, n⋆p, (n⋆₂m), (m⋆₂p))
+                end
+
+                #==  # This breaks FiniteDiff
+                # TODO uncomment after we have: https://github.com/JuliaDiff/FiniteDifferences.jl/pull/110
```
oxinabox, line 32: Needs https://github.com/JuliaDiff/FiniteDifferences.jl/pull/110

which needs

https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/63

oxinabox:
This is basically done, but can't be merged until
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/59

oxinabox:
Another advantage of this over the Nabla code is that we don't have to worry about the tests not hitting the right one, because of the fact that this *is* the most general one.

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.7.23"
```
oxinabox, line 4: Version number needs to be bumped

src/rulesets/Base/arraymath.jl
```diff
@@ -19,23 +19,47 @@ end
...
-function rrule(::typeof(*), A::Real, B::AbstractArray{<:Real})
+function rrule(::typeof(*), A::Number, B::AbstractArray{<:Number})
     function times_pullback(Ȳ)
-        return (NO_FIELDS, @thunk(dot(Ȳ, B)), @thunk(A * Ȳ))
+        return (
+            NO_FIELDS,
+            @thunk(dot(Ȳ, B)'),
+            InplaceableThunk(
+                @thunk(A' * Ȳ),
+                X̄ -> mul!(X̄, A, Ȳ, true, true)
```
oxinabox, line 46: I suspect some conjugation is missing here

src/rulesets/Base/arraymath.jl
```diff
@@ -19,23 +19,47 @@ end
 ##### `*`
 #####
 
-function rrule(::typeof(*), A::AbstractMatrix{<:Real}, B::AbstractMatrix{<:Real})
+function rrule(::typeof(*), A::AbstractMatrix{<:Number}, B::AbstractMatrix{<:Number})
```
willtebbutt, line 22: @sethaxen pointed out that we probably want to restrict these to `Real` and `Complex` numbers.

test/rulesets/Base/arraymath.jl
```diff
@@ -6,26 +6,50 @@
...
-                Ȳ = randn(m, p)
-                rrule_test(*, Ȳ, (A, randn(m, n)), (B, randn(n, p)))
-            end
+    # TODO Test on Complex once we have Complex support for all of these.
+    @testset "*: $T" for T in (Float64, ComplexF64)
+        ⋆(a) = round.(5*randn(T, a))  # Helper to generate nice random values
+        ⋆(a, b) = ⋆((a, b))  # matrix
+        ⋆() = only(⋆(()))  # scalar
+
+        ⋆₂(a) = (⋆(a), ⋆(a)) # Helper to generate random matrix and its cotangent
```
willtebbutt, line 15: What are these symbols? I don't think I've ever seen them before...

test/rulesets/Base/arraymath.jl
```diff
@@ -6,26 +6,50 @@
...
-            dims = [3,4,5]
-            for n in dims, m in dims, p in dims
-                # don't need to test square case multiple times
-                n > 3 && n == m == p && continue
-                A = randn(m, n)
-                B = randn(n, p)
-                Ȳ = randn(m, p)
-                rrule_test(*, Ȳ, (A, randn(m, n)), (B, randn(n, p)))
-            end
+    # TODO Test on Complex once we have Complex support for all of these.
```
willtebbutt, line 9: Can this go now?

willtebbutt (commented):
Looking good. Just a couple of comments.

test/rulesets/Base/arraymath.jl
oxinabox:
```suggestion
```

test/rulesets/Base/arraymath.jl
oxinabox: `\star` and `\star\_2`

willtebbutt: Fair enough. Thanks!

PR #277, opened 2020-10-09
add rules for round, floor, ceil

### LINKED ISSUE #100: Missing Scalar Rules
We have a lot of scalar rules 
A few I know are still missing

- [x] `ldexp` ([WolframAlpha partials](https://www.wolframalpha.com/input/?i=partial+derivatives+ldexp%28x%2C+y%29))
 - [x] `clamp` see https://discuss.pytorch.org/t/torch-round-gradient/28628/3
 - [x] `polygamma`

### `Zero` everywhere it is differentiable

Idk if we want to close over the point gsps. (We do that for `abs`)

 - [ ] `round` see https://discuss.pytorch.org/t/torch-round-gradient/28628/3
 - [ ] `floor`
 - [ ] `ceil`

Moelf:
close #100

tests don't fully pass because `fdm` doesn't work for all of them:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/85739e41653de0e0026776116b62f2bcf2a7ba82/src/testers.jl#L94
```
julia> _fdm = ChainRulesTestUtils._fdm;

julia> ChainRulesTestUtils.jvp(_fdm, xs->ceil(xs...), (0.1, 1.0))
5.833333333333332
```
should we "manually test" it and/or use `complex_jacobian_test` (like for `abs`)?

### DIFF
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -33,6 +33,11 @@ let
         @scalar_rule log10(x) inv(x) / log(oftype(x, 10))
         @scalar_rule log1p(x) inv(x + 1)
         @scalar_rule log2(x) inv(x) / log(oftype(x, 2))
+        
+        # rouding related
+        @scalar_rule round(x) zero(x)
+        @scalar_rule floor(x) zero(x)
+        @scalar_rule ceil(x) zero(x)
 
 
         # Unary complex functions
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -34,7 +34,8 @@ let
         @scalar_rule log1p(x) inv(x + 1)
         @scalar_rule log2(x) inv(x) / log(oftype(x, 2))
         
-        # rouding related
+        # rouding related, 
+        # we use `zero` rather than `Zero()` for scalar, and avoids issues with map etc
         @scalar_rule round(x) zero(x)
         @scalar_rule floor(x) zero(x)
         @scalar_rule ceil(x) zero(x)
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -90,6 +90,16 @@ const FASTABLE_AST = quote
         end
     end
 
+    @testset "rounding" begin
+        for x in (-0.1, 6.4, 0.5 + 0.25im)
+            test_scalar(round, x)
+            if x isa Real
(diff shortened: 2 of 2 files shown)


test/rulesets/Base/fastmath_able.jl
```diff
@@ -91,12 +91,9 @@ const FASTABLE_AST = quote
...
     @testset "rounding" begin
-        for x in (-0.1, 6.4, 0.5 + 0.25im)
-            test_scalar(round, x)
-            if x isa Real
-                test_scalar(floor, x)
-                test_scalar(ceil, x)
-            end
+        for x in (-0.6, -0.2, -0.1, 0.1, 0.2, 0.6, 6.4)
+            test_scalar(floor, x, fdm = backward_fdm(5, 1))
+            test_scalar(ceil, x, fdm = forward_fdm(5, 1))
```
oxinabox, line 96: keyword arguments in BlueStyle are sperated by `;` and don't use spaces around `=`
```suggestion
            test_scalar(floor, x; fdm=backward_fdm(5, 1))
            test_scalar(ceil, x; fdm=forward_fdm(5, 1))
```
https://github.com/invenia/BlueStyle#keyword-arguments

Also it it intentional that the tests on `round` were deleted?
Maybe there are to be moved down to there own section so we don't needthe `if` by you forgot to put them back?

Also do we need to test both `-0.1`, and `-0.2` (similar for positive).
My instinct is we want to test one point on each side of 0.5, and maybe again for negative

Moelf:
one caveat is that primal `round` works on `Complex` (acting on each part individually). But if backward and forward need to be mixed, it can't be tested properly.

oxinabox:
> one caveat is that primal round works on Complex (acting on each part individually). But if backward and forward need to be mixed (0.2 + 0.7im), it can't be tested properly.

I think that is fine not to test, we can assume the code is not actively trying to defeat the tests.

oxinabox (approved):
Thanks

src/rulesets/Base/fastmath_able.jl
```diff
@@ -33,6 +33,11 @@ let
         @scalar_rule log10(x) inv(x) / log(oftype(x, 10))
         @scalar_rule log1p(x) inv(x + 1)
         @scalar_rule log2(x) inv(x) / log(oftype(x, 2))
+        
+        # rouding related
```
oxinabox, line 37:
```suggestion
        # rouding related, 
        # we use `zero` rather than `Zero()` because cheaper for scalar, and avoids issues with map etc
```

PR #279, opened 2020-10-12
Correct inplace for complex BLAS and fix some accumulation against nondifferentails

### LINKED ISSUE #278: Inplace Accumulation for complex BLAS wrong?
Looking at the Integration CI failures on  https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/59
it seems like we might actually have some rules that are incorrect for inplace accumulation, if the inputs are Complex

oxinabox:
closes #278 

This PR is the follow up from the integration tests failing in
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/59

The core problem with the complex numbers was we were doing
`a = (b + a)' ` 
but we should have been doing
`a = b' + a`
but BLAS doesn;t have a shorthand for that
so what we actually do is:
`a = (b + a')'`

Other problem was us using some things were not differentials, and trying to acculate against them.

Requires
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/111
for tests to pass, as needs to `to_vec` support for `AbstractZeros`
because its used in how ChainRulesTestUtil checks equality for `Composites`
which doesn't seem great, and seems like  in longer term to destructure the eqaulity checks is better
like in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/57

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -43,7 +43,7 @@
             test_scalar(acscd, 1/x)
             test_scalar(acotd, 1/x)
         end
-        
+
         @testset "sinc" for x = (0.0, 0.434, Complex(0.434, 0.25))
             test_scalar(sinc, x)
         end
@@ -126,14 +126,14 @@
         frule_test(identity, (randn(T, 4), randn(T, 4)))
         frule_test(
             identity,
-            (Composite{Tuple}(randn(T, 3)...), Composite{Tuple}(randn(T, 3)...))
+            (Tuple(randn(T, 3)), Composite{Tuple{T, T, T}}(randn(T, 3)...))
         )
 
         rrule_test(identity, randn(T), (randn(T), randn(T)))
         rrule_test(identity, randn(T, 4), (randn(T, 4), randn(T, 4)))
         rrule_test(
-            identity, Tuple(randn(T, 3)),
-            (Composite{Tuple}(randn(T, 3)...), Composite{Tuple}(randn(T, 3)...))
+            identity, Composite{Tuple{T, T, T}}(randn(T, 3)...),
+            (Tuple(randn(T, 3)), Composite{Tuple{T, T, T}}(randn(T, 3)...))
         )
     end
 
--- a/test/rulesets/Base/evalpoly.jl
+++ b/test/rulesets/Base/evalpoly.jl
@@ -27,9 +27,8 @@
             Ω = evalpoly(x, p)
             Ω̄ = randn(T, size(Ω)...)
             frule_test(evalpoly, (x, ẋ), (p, ṗ))
-            frule_test(evalpoly, (x, ẋ), (Tuple(p), Tuple(ṗ)))
+            frule_test(evalpoly, (x, ẋ), (Tuple(p), rand_tangent(Tuple(p))))
             rrule_test(evalpoly, Ω̄, (x, x̄), (p, p̄))
-            rrule_test(evalpoly, Ω̄, (x, x̄), (Tuple(p), Tuple(p̄)))
+            rrule_test(evalpoly, Ω̄, (x, x̄), (Tuple(p), rand_tangent(Tuple(p))))
         end
     end
-
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -38,6 +38,8 @@ function complex_jacobian_test(f, z)
 end
 
 const FASTABLE_AST = quote
+    is_fastmath_mode = sin === Base.FastMath.sin_fast
+
     @testset "Trig" begin
         @testset "Basics" for x = (Float64(π)-0.01, Complex(π, π/2))
             test_scalar(sin, x)
@@ -66,7 +68,7 @@ const FASTABLE_AST = quote
     end
 
     @testset "exponents" begin
-        for x in (-0.1, 6.4, 0.5 + 0.25im)
+        for x in (-0.1, 7.9, 0.5 + 0.25im)
             test_scalar(inv, x)
 
             test_scalar(exp, x)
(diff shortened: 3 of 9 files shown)


test/rulesets/packages/SpecialFunctions.jl
```diff
@@ -50,7 +50,7 @@ end
             isreal(x) || continue
 
             Δx, x̄ = randn(2)
-            Δz = (randn(), randn())
+            Δz = Composite{Tuple{Float64, DoesNotExist}}(randn(), Zero())
```
sethaxen, line 53: Why this change? Is it related to the BLAS fixes?

test/rulesets/Base/evalpoly.jl
```diff
@@ -22,14 +22,14 @@
             sp = (np..., np...)
             x, ẋ, x̄ = randn(T, sx...), randn(T, sx...), randn(T, sx...)
             p = [randn(T, sp...) for _ in 1:m]
+            @show p
```
sethaxen, line 25: Did you mean to make this change?

test/rulesets/Base/evalpoly.jl
```diff
@@ -22,14 +22,14 @@
...
             ṗ = [randn(T, sp...) for _ in 1:m]
             p̄ = [randn(T, sp...) for _ in 1:m]
             Ω = evalpoly(x, p)
             Ω̄ = randn(T, size(Ω)...)
             frule_test(evalpoly, (x, ẋ), (p, ṗ))
-            frule_test(evalpoly, (x, ẋ), (Tuple(p), Tuple(ṗ)))
+            frule_test(evalpoly, (x, ẋ), (Tuple(p), rand_tangent(Tuple(p))))
```
sethaxen, line 31: Would it make sense to use `rand_tangent` for all tangents in this test set?

test/rulesets/Base/base.jl
```diff
@@ -126,14 +126,14 @@
         frule_test(identity, (randn(T, 4), randn(T, 4)))
         frule_test(
             identity,
-            (Composite{Tuple}(randn(T, 3)...), Composite{Tuple}(randn(T, 3)...))
+            (Tuple(randn(T, 3)), Composite{Tuple{T, T, T}}(randn(T, 3)...))
```
sethaxen, line 129: Could you use `rand_tangent` here as well?

test/rulesets/Base/base.jl
```diff
@@ -126,14 +126,14 @@
...
         )
 
         rrule_test(identity, randn(T), (randn(T), randn(T)))
         rrule_test(identity, randn(T, 4), (randn(T, 4), randn(T, 4)))
         rrule_test(
-            identity, Tuple(randn(T, 3)),
-            (Composite{Tuple}(randn(T, 3)...), Composite{Tuple}(randn(T, 3)...))
+            identity, Composite{Tuple{T, T, T}}(randn(T, 3)...),
```
sethaxen, line 135: Could you use `rand_tangent` here as well?

sethaxen (approved):
The BLAS fixes LGTM. I had a few minor questions about the changes to the tests.

Also, I wonder if it would make sense to also include in this PR some fixes to not allocate new conjugated matrices in the BLAS rules? e.g. in
https://github.com/JuliaDiff/ChainRules.jl/blob/c877550430a3cb0f657f43ccbf4e64b13b7177a7/src/rulesets/LinearAlgebra/blas.jl#L134-L137,
we allocate a new matrix `A`, which I think can be made faster by allocating a new vector instead and then conjugating a vector in-place:
```julia
             ∂x = InplaceableThunk( 
                 @thunk(gemv('N', α', conj(A), ȳ)), 
                 x̄ -> conj!(gemv!('N', α, A, conj(ȳ), one(T), conj!(x̄)))
             ) 
```

test/rulesets/Base/evalpoly.jl
oxinabox: I did not`

test/rulesets/packages/SpecialFunctions.jl
oxinabox: Not related to the problem with BLAS, but it does cause a problem for testing accumulation.
Since you can add to a tuple

test/rulesets/Base/evalpoly.jl
oxinabox:
```suggestion
```

test/rulesets/Base/evalpoly.jl
oxinabox: It would make sense to use `rand_tangent` in a whole ton of places.
Infact, we should use it by default everywhere.
See: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/67

I am not inclines to get through and change to using it manually in a bunch of places when what we should do is change to using it automatically everywhere.

I made the change here as it was shorted than expressly writing out the Composite
I could make it in more places here, but that would be longer than what we have.
so I figure the rest could wait for us to do tangent generation automatically

oxinabox:
> Also, I wonder if it would make sense to also include in this PR some fixes to not allocate new conjugated matrices in the BLAS rules? 

That is a good idea.
It would be worth benchmarking it.
I think not in this PR though, as this is a bug-fix, and better to get it in sooner, and have other PR for a feature enhancement.
I will open an issue.

test/rulesets/Base/evalpoly.jl
sethaxen: Okay, makes sense

test/rulesets/Base/fastmath_able.jl
```diff
@@ -52,7 +54,7 @@ const FASTABLE_AST = quote
         @testset "Inverses" for x = (0.5, Complex(0.5, 0.25))
             test_scalar(asin, x)
             test_scalar(acos, x)
-            test_scalar(atan, x)
+            test_scalar(atan, x; )
```
oxinabox, line 57:
```suggestion
            test_scalar(atan, x)
```

oxinabox:
This passes, once we have
https://github.com/JuliaDiff/ChainRules.jl/pull/286
to use FiniteDifferences v0.11

test/rulesets/Base/fastmath_able.jl
```diff
@@ -164,17 +167,17 @@ const FASTABLE_AST = quote
 
         @testset "^(x::$T, n::$T)" for T in (Float64, ComplexF64)
             # for real x and n, x must be >0
-            x = T <: Real ? 15rand() : 15randn(ComplexF64)
-            Δx, x̄ = 10rand(T, 2)
-            y, Δy, ȳ = rand(T, 3)
+            x, Δx, x̄ = rand(T, 2) .+ 3
+            y, Δy, ȳ = rand(T, 2) .+ 3
```
oxinabox, line 171:
```suggestion
            x, Δx, x̄ = rand(T, 3) .+ 3
            y, Δy, ȳ = rand(T, 3) .+ 3
```

PR #282, opened 2020-10-13
Adds some missing unthunks

oxinabox:
I am not 100% sure how i feel about this.
Maybe we should actually overload some more linear operators in ChainRulesCore.
Including in particular: the `Diagonal` constructor,
and `\` with a `Vector`

But for `svd_rev` because we use some of the inputs multiple times, it is actually faster to unthunk them at the start anyway.
I feel like we shouldn't actually have to use any input more than once, but that is another issue.

The testset in Nabla that revealed this was:
```julia
    @testset "Tape updating from multiple components" begin
        ∇f = ∇() do X
            U, S, V = svd(X)
            Y = U * Diagonal(S) * V'
            sum(Y)
        end
        X = [1.0 2.0; 3.0 4.0; 5.0 6.0]
        @test ∇f(X)[1] ≈ ones(3, 2) atol=1e-5
    end
```

I am not sure what is a good way to test this PR.
Do we want to add something to ChainRulesTestUtils that redoes all tests passing in Thunk's of the inputs in as well?
That wouldn't trivially catch Composites.

This is needed for https://github.com/invenia/Nabla.jl/pull/189

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -8,7 +8,7 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
 function rrule(::typeof(svd), X::AbstractMatrix{<:Real})
     F = svd(X)
     function svd_pullback(Ȳ::Composite)
-        # svd_rev does a lot of linear algebra, it it is efficient to unthunk before
+        # svd_rev does a lot of linear algebra, it is is efficient to unthunk before
         ∂X = svd_rev(F, unthunk(Ȳ.U), unthunk(Ȳ.S), unthunk(Ȳ.V))
         return (NO_FIELDS, ∂X)
     end
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -8,7 +8,8 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
 function rrule(::typeof(svd), X::AbstractMatrix{<:Real})
     F = svd(X)
     function svd_pullback(Ȳ::Composite)
-        ∂X = svd_rev(F, Ȳ.U, Ȳ.S, Ȳ.V)
+        # svd_rev does a lot of linear algebra, it it is efficient to unthunk before
+        ∂X = svd_rev(F, unthunk(Ȳ.U), unthunk(Ȳ.S), unthunk(Ȳ.V))
         return (NO_FIELDS, ∂X)
     end
     return F, svd_pullback
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -32,6 +32,9 @@ end
 #####
 
 function rrule(::Type{<:Diagonal}, d::AbstractVector)
+    function Diagonal_pullback(x::AbstractThunk)
+        return Diagonal_pullback(unthunk(x))
+    end
     function Diagonal_pullback(ȳ::AbstractMatrix)
         return (NO_FIELDS, diag(ȳ))
     end
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -32,9 +32,6 @@ end
 #####
 
 function rrule(::Type{<:Diagonal}, d::AbstractVector)
-    function Diagonal_pullback(x::AbstractThunk)
-        return Diagonal_pullback(unthunk(x))
-    end
     function Diagonal_pullback(ȳ::AbstractMatrix)
         return (NO_FIELDS, diag(ȳ))
     end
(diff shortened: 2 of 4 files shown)


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -8,7 +8,8 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
 function rrule(::typeof(svd), X::AbstractMatrix{<:Real})
     F = svd(X)
     function svd_pullback(Ȳ::Composite)
-        ∂X = svd_rev(F, Ȳ.U, Ȳ.S, Ȳ.V)
+        # svd_rev does a lot of linear algebra, it it is efficient to unthunk before
```
mzgubic, line 11:
```suggestion
        # svd_rev does a lot of linear algebra, it is efficient to unthunk before
```

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -32,6 +32,9 @@ end
 #####
 
 function rrule(::Type{<:Diagonal}, d::AbstractVector)
+    function Diagonal_pullback(x::AbstractThunk)
+        return Diagonal_pullback(unthunk(x))
+    end
     function Diagonal_pullback(ȳ::AbstractMatrix)
```
mzgubic, line 38: I'm just wondering how did that work before? I.e. would there not be a `MethodError` if an `AbstractThunk` is passed to this? Or is there some black magic in how thunks work?

mzgubic (approved):

mzgubic:
What actually happened in Nabla with that testset, i.e. why did it not pass? I thought the only difference between the current implementation and the new one is that the `unthunk(x)` happens only once and not many times, and the value is reused. Or are there some side-effects happening under the hood in `unthunk`?

oxinabox:
> What actually happened in Nabla with that testset, i.e. why did it not pass? 

It throws a MethodError,
both for the constructor, and then if that is fixed within `svd_rev` for `\`

oxinabox:
> Are such Composites a common case however? I.e. the rrule would have to return a pullback where Some_pulback(x::Composite{SpecificType}) while what would be passed is Composite{Thunk}, if I understand the issue correctly?

It isn't` Composite{Thunk}` it is `Composite{SpecificType}.X isa Thunk`
and it is very common.

mzgubic:
> It isn't` Composite{Thunk}` it is `Composite{SpecificType}.X isa Thunk`
> and it is very common.

How does this break things though? (Not by `MethodError`, right?)

oxinabox:
By `MethodError`

> > It isn't` Composite{Thunk}` it is `Composite{SpecificType}.X isa Thunk`
> > and it is very common.
> 
> How does this break things though? (Not by `MethodError`, right?)

By method error when you try and do `Diagonal(comp.F)`
or `comp.F / s`

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -13,12 +13,12 @@ using ChainRules: level2partition, level3partition, chol_blocked_rev, chol_unblo
...
                 @test dp === DoesNotExist()
 
-                ΔF = unthunk(dF)
-                dself2, dX = dX_pullback(ΔF)
+                dself2, dX = dX_pullback(dF)
                 @test dself2 === NO_FIELDS
                 X̄_ad = unthunk(dX)
                 X̄_fd = only(j′vp(central_fdm(5, 1), X->getproperty(svd(X), p), Ȳ, X))
                 @test all(isapprox.(X̄_ad, X̄_fd; rtol=1e-6, atol=1e-6))
+
```
mzgubic, line 21:
```suggestion
```

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -27,6 +27,26 @@ using ChainRules: level2partition, level3partition, chol_blocked_rev, chol_unblo
...
+                _, dF_unthunked, _ = dF_pullback(Ȳ)
+
+                @assert !(getproperty(dF_unthunked, p) isa AbstractThunk)
+                dF_thunked = map(f->Thunk(()->f), dF_unthunked)
+                @assert getproperty(dF_thunked, p) isa AbstractThunk
+
+                dself_thunked, dX_thunked = dX_pullback(dF_thunked)
+                dself_unthunked, dX_unthunked = dX_pullback(dF_unthunked)
+                @test dself_thunked == dself_unthunked
```
mzgubic, line 45: Is this added for completeness or is there a reason to think this test could fail and the other one pass? I guess we could also add a comment that we are fishing for `MethodError`s?

mzgubic (approved):

oxinabox:
The unthunk in Diagonal was a mistake in Nabla.
Nabla was supposed to be unthunking things before it input them into propagators, but it wasn't.
So we can remove the Diagonal one.
Probably doesn't help with the Composite one since nabla shouldn't be recursively unthunking.
https://github.com/JuliaDiff/ChainRulesCore.jl/pull/121 would solve it.
But in this case we still want to unthunk in advance since things are used multiple times in svd_rev.
They probably shouldn't be but they are.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -32,6 +32,9 @@ end
 #####
 
 function rrule(::Type{<:Diagonal}, d::AbstractVector)
+    function Diagonal_pullback(x::AbstractThunk)
+        return Diagonal_pullback(unthunk(x))
+    end
```
oxinabox, line 37:
```suggestion
```

Not needed see https://github.com/JuliaDiff/ChainRules.jl/pull/282#issuecomment-708452476

PR #283, opened 2020-10-13
det and logdet for structured matrixes

oxinabox:
This is based on https://github.com/invenia/Nabla.jl/blob/4cadc87677fb1187354999dcf93bd528f45f85d0/src/sensitivities/linalg/triangular.jl

Differences:
 - It generalizes it somewhat to apply `Diagonal` as well as `AbstractTriangular` matrixes
 - It doesn't currently implement the inplace rules, because I am not sure they are worth it. 
    - On the one hard it lets the `/` be fused with the accumulation, and avoids allocating the vector `s`
    - On the other hand it it means carrying a reference to `X` rather than just a vector `s` of the same size as the diagonal of `x`
   - @willtebbutt  what do you think? Should I add the inplace rules?

### DIFF
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -280,7 +280,7 @@ _diag_view(X::Diagonal) = parent(X)  #Diagonal wraps a Vector of just Diagonal e
 
 function rrule(::typeof(det), X::Union{Diagonal, AbstractTriangular})
     y = det(X)
-    s = y ./ _diag_view(X)
+    s = conj!(y ./ _diag_view(X))
     function det_pullback(ȳ)
         return (NO_FIELDS, Diagonal(ȳ .* s))
     end
@@ -289,7 +289,7 @@ end
 
 function rrule(::typeof(logdet), X::Union{Diagonal, AbstractTriangular})
     y = logdet(X)
-    s = one(eltype(X)) ./ _diag_view(X)
+    s = conj!(one(eltype(X)) ./ _diag_view(X))
     function logdet_pullback(ȳ)
         return (NO_FIELDS, Diagonal(ȳ .* s))
     end
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -274,3 +274,24 @@ function rrule(::typeof(tril), A::AbstractMatrix)
     end
     return tril(A), tril_pullback
 end
+
+_diag_view(X) = view(X, diagind(X))
+_diag_view(X::Diagonal) = parent(X)  #Diagonal wraps a Vector of just Diagonal elements
+
+function rrule(::typeof(det), X::Union{Diagonal, AbstractTriangular})
+    y = det(X)
+    s = y ./ _diag_view(X)
+    function det_pullback(ȳ)
+        return (NO_FIELDS, Diagonal(ȳ .* s))
+    end
+    return y, det_pullback
+end
+
+function rrule(::typeof(logdet), X::Union{Diagonal, AbstractTriangular})
+    y = logdet(X)
+    s = one(eltype(X)) ./ _diag_view(X)
+    function logdet_pullback(ȳ)
+        return (NO_FIELDS, Diagonal(ȳ .* s))
+    end
+    return y, logdet_pullback
+end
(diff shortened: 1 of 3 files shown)


test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -156,4 +156,20 @@
...
+        # rand (not randn) so det will be postive, so logdet will be defined
+        X = T(rand(n, n))
+        X̄_acc = Diagonal(rand(n, n))  # sensitivity is always a diagonal for these types
+        @testset "$op" for op in (det, logdet)
+            rrule_test(op, 2.7, (X, X̄_acc))
+
+            @testset "return type" begin
+                _, op_pullback = rrule(op, X)
+                X̄ = op_pullback(2.7)[2]
+                @test X̄ isa Diagonal
```
oxinabox, line 171: is there an issue about testing this using ChainRulesTestUtils?

test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -156,4 +156,20 @@
             rrule_test(Op, randn(n, n), (randn(n, n), randn(n, n)), (k, nothing))
         end
     end
+
+    @testset "det and logdet $T" for T in (Diagonal, UpperTriangular, LowerTriangular)
+        n = 5
+        # rand (not randn) so det will be postive, so logdet will be defined
+        X = T(rand(n, n))
```
willtebbutt, line 163: You might want to add a small +ve constant to these to ensure stability under changing seed -- finite differencing might make one of the values less than zero if one of the sampled values happens to be really close to zero.

willtebbutt (approved):
LGTM modulo slight testing modification.

willtebbutt:
> @willtebbutt what do you think? Should I add the inplace rules?

I really wouldn't bother for now -- I've certainly not come across a situation in which they're likely to be particuarly important.

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -274,3 +274,24 @@ function rrule(::typeof(tril), A::AbstractMatrix)
     end
     return tril(A), tril_pullback
 end
+
+_diag_view(X) = view(X, diagind(X))
+_diag_view(X::Diagonal) = parent(X)  #Diagonal wraps a Vector of just Diagonal elements
+
+function rrule(::typeof(det), X::Union{Diagonal, AbstractTriangular})
+    y = det(X)
+    s = y ./ _diag_view(X)
```
sethaxen, line 283: Why is this declared out here instead of in the pullback?

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -274,3 +274,24 @@ function rrule(::typeof(tril), A::AbstractMatrix)
...
+    function det_pullback(ȳ)
+        return (NO_FIELDS, Diagonal(ȳ .* s))
+    end
+    return y, det_pullback
+end
+
+function rrule(::typeof(logdet), X::Union{Diagonal, AbstractTriangular})
+    y = logdet(X)
+    s = one(eltype(X)) ./ _diag_view(X)
```
sethaxen, line 292: What's the benefit of doing this division here instead of just doing on division instead of multiplication in the pullback?

sethaxen (changes requested):
I don't think these rules currently conjugate correctly for complex inputs. Comparing with the dense rules should be enough.

src/rulesets/LinearAlgebra/structured.jl
oxinabox: To avoid closing over `X`
so that `X` can be garbage collected.
`s` is much smaller than `X`

src/rulesets/LinearAlgebra/structured.jl
oxinabox: To avoid closing over X
so that X can be garbage collected.
s is much smaller than X

src/rulesets/LinearAlgebra/structured.jl
oxinabox: I am hoping opaque closures will do a lot of this for us automatically.
Not sure though

oxinabox (commented):
.

test/rulesets/LinearAlgebra/structured.jl
oxinabox:
```suggestion
        X = T(3*rand(n, n) .+ 1)
```

Issue #284, opened 2020-10-13
using FiniteDifferences v0.11 required relaxing a bunch of tests


oxinabox:
This is the commit
https://github.com/JuliaDiff/ChainRules.jl/pull/279/commits/9fc8c57055b7e3315b747ea648336fb5f75bdca8#diff-18f60c608d1cadf3013a4e16d696c43ea9f488fa49caa48cb6b582481025bc0fR57

It might be that FiniteDifferences v0.11 need ssome changes to improve accuracy,
or it might be that we should be using a different set of settings in ChainRulesTestUtils.jl by default for `fdm`

oxinabox:
yes, I am pretty sure it is for Complex valued inputs

oxinabox:
@wesselb  and I had a meeting on this. here are our notes:


## **Observations _i.e. what we are seeing_**


*   Tests that passed with FiniteDifferences v0.10 don’t with v0.11
*   _Mostly Complex ones_
*   _Mostly FastMath_
*   _Float32 behaves badly, but that might not have changed_
*   _Fails for x=2.6, but passes for x=2.5 and x=2.3_


## **Diagnosis _ i.e. the challenge_**


*   Step size change means we are crossing over discontinoualities
*   Could be similar but for ill-conditioned.


## **Actions _ i.e. what we are going to do_**


*   Check for crossing Discontinuality
    *   Would be explained as we changed the step-size. Possibly old size just happened to work for size we were looking at
    *   Move away from them. regardless of the change we don't really want to be clsoe to them in the first place
    *   Use forward_fdm where necessary
*   Produce small test cases that fail
    *   Make those unit test for FiniteDifferences
*   Double check 1e-40

PR #286, opened 2020-10-14
Swap to FiniteDIfferences v0.11: relax a bunch of constraints in tests

oxinabox:
This is cherry-picked out of #279  
It is the subject of #284 
hopefully if we get to the bottom of that we can revert most of this PR


oxinabox:
Actually can't do this without the rest of #279 as other changes came in at the same time that require the fixes there

PR #304, opened 2020-11-10
Remove uses of similar to generate test inputs

oxinabox:
This is why we have been getting non-determinstic failures, even thouh the RNG seed was set.
`similar` just has whatever values was in RAM.
Which could well be NaN.

Turns out it was not FiniteDifferences.jl v0.11 at all.

### DIFF
--- a/test/rulesets/LinearAlgebra/dense.jl
+++ b/test/rulesets/LinearAlgebra/dense.jl
@@ -28,7 +28,7 @@
             M, N = 3, 4
             x, A, y = randn(T, M), randn(T, M, N), randn(T, N)
             ẋ, Adot, ẏ = randn(T, M), randn(T, M, N), randn(T, N)
-            x̄, Abar, ȳ = similar(x), similar(A), similar(y)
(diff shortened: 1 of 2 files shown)


glennmoy (approved):

nickrobinson251 (approved):

mcabbott:
My bad, I guess. But why is `rrule_test` reading such values, at all?

I mean if it wants empty space, it should ideally generate it, or zero it. If it wanted valid perturbations, wouldn't that be called forward mode?

simeonschaub:
Did you intentionally check in ChainRulesTestUtils as a submodule?

nickrobinson251:
> Did you intentionally check in ChainRulesTestUtils as a submodule?

😱 

OMG completely missed that in the review! (Feel like the one-line diff for addign a submodule hasn't helped us here *shakes fist at GitHub*) 

Pretty sure it wasn't intended (cc @oxinabox)

oxinabox:
It was very much not intended.

PR #306, opened 2020-11-15
Implement QR pullback

Kolaru:
Implement the pullback for the QR decomposition, following:

Walter and Lehmann, 2018, Algorithmic Differentiation of Linear Algebra Functions with Application in Optimum Experimental Design


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -261,3 +261,67 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+#####
+
+const QR_TYPE = Union{QR, QRPivoted, LinearAlgebra.QRCompactWY}
+
+function rrule(::typeof(qr), X::AbstractMatrix{<:Real})
+    F = qr(X)
+
+    if size(F.R, 2) != size(F.R, 1)
+        throw(ArgumentError("Pullback for QR decomposition is only supported for m × n matrices with m >= n"))
+    end
```
oxinabox, line 276: Why?

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -138,4 +138,40 @@ using ChainRules: level2partition, level3partition, chol_blocked_rev, chol_unblo
...
+                    @test dself1 === NO_FIELDS
+                    @test dp === DoesNotExist()
+
+                    dself2, dX = dX_pullback(dF)
+                    @test dself2 === NO_FIELDS
+                    X̄_ad = unthunk(dX)
+
+                    # Explicitely convert to Matrix since FiniteDifferences seem to
+                    # be broken for LinearAlgebra.QRCompactWYQ (infinite to_vec
+                    # recursion)
```
oxinabox, line 166: Can an issue openned for this an linked here?

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -138,4 +138,40 @@ using ChainRules: level2partition, level3partition, chol_blocked_rev, chol_unblo
...
+                X = randn(n, m)
+
+                if n < m
+                    @test_throws ArgumentError F, dX_pullback = rrule(qr, X)
+                    continue
+                end
+
+                F, dX_pullback = rrule(qr, X)
+                for p in [:Q, :R]
+                    Y, dF_pullback = rrule(getproperty, F, p)
```
oxinabox, line 153: does `rrule_test` not work on this?

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -138,4 +138,40 @@ using ChainRules: level2partition, level3partition, chol_blocked_rev, chol_unblo
             @test chol_unblocked_rev(B̄, B, true) ≈ chol_blocked_rev(B̄, B, 10, true)
         end
     end
+    @testset "qr" begin
+        @testset "the thing" begin
```
oxinabox, line 142: Please explain

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -261,3 +261,67 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+end
+
+function rrule(::typeof(getproperty), F::T, x::Symbol) where T <: QR_TYPE
+    function getproperty_qr_pullback(Ȳ)
+        C = Composite{T}
+        ∂F = if x === :Q
+            C(Q=Ȳ,)
+        elseif x === :R
+            C(R=Ȳ,)
+        end
```
oxinabox, line 292: Is it worth defining `C` ?
```suggestion
        ∂F = if x === :Q
            Composite{T}(Q=Ȳ,)
        elseif x === :R
            Composite{T}(R=Ȳ,)
        end
```

src/rulesets/LinearAlgebra/utils.jl
```diff
@@ -27,3 +27,17 @@ function _eyesubx!(X::AbstractMatrix)
 end
 
 _extract_imag(x) = complex(0, imag(x))
+
+# Lower triangle of X - X' overwrite X if possible
```
oxinabox, line 31: this seems to always overwrite?

src/rulesets/LinearAlgebra/utils.jl
```diff
@@ -27,3 +27,17 @@ function _eyesubx!(X::AbstractMatrix)
...
+function _sublowertri!(X::AbstractMatrix{T}) where T<:Real
```
oxinabox, line 32: Is this actually faster than `.-` and `LowerTriangular`?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -261,3 +261,67 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+        return NO_FIELDS, ∂F, DoesNotExist()
+    end
+    if x === :Q
+        # Return thing Q for consistency
+        n = size(F.R, 1)
+        return F.Q[:, 1:n], getproperty_qr_pullback
```
oxinabox, line 298: Is this on to do to the primal value?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -261,3 +261,67 @@ function chol_blocked_rev(Σ̄::AbstractMatrix, L::AbstractMatrix, nb::Integer,
...
+        C = Composite{T}
+        ∂F = if x === :Q
+            C(Q=Ȳ,)
+        elseif x === :R
+            C(R=Ȳ,)
+        end
+        return NO_FIELDS, ∂F, DoesNotExist()
+    end
+    if x === :Q
+        # Return thing Q for consistency
```
oxinabox, line 296: I was very confused
```suggestion
        # Return thin Q for consistency
```

src/rulesets/LinearAlgebra/factorization.jl
Kolaru: Because I have not found a reference for that case.

test/rulesets/LinearAlgebra/factorization.jl
Kolaru: Done here https://github.com/JuliaDiff/FiniteDifferences.jl/issues/114

test/rulesets/LinearAlgebra/factorization.jl
Kolaru: I just reused the same structure as the test for cholesky. I can give it a more explicit name.

src/rulesets/LinearAlgebra/factorization.jl
Kolaru: I did it analog to the `svd` case. I think I could go even simpler with

```julia
∂F = Composite{T}(; x => Ȳ)
```

Should I port that kind of cleanup to `svd` and `cholesky` as well?

test/rulesets/LinearAlgebra/factorization.jl
oxinabox: Please do.

src/rulesets/LinearAlgebra/utils.jl
Kolaru: Yes (I just quickly benchmarked it), because it does not compute anything for the upper part of the matrix.

test/rulesets/LinearAlgebra/factorization.jl
Kolaru: It does not work on `getproperty` because of the failure of `FiniteDifferences.jl` mentionned elsewhere on this PR.

It does not work directly on `qr` either, because the object returned by `qr` can not be `collect`ed. See https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/74

src/rulesets/LinearAlgebra/factorization.jl
Kolaru: This was because the anything else does not contribute, and the tests used the result of this to infer dimensions. However after pondering on it I figured out it is probably better to let the `getproperty` pullback do just the obvious and have the `qr` pullback make sure to only take in account what is relevant and make sure the dimensions match.

sethaxen:
I wonder if the QR rule implied by Seeger et al in https://arxiv.org/pdf/1710.08717.pdf is more performant than the one in Walter and Lehmann? (they actually define an LQ rule, but the same approach produces a QR rule). The below reimplementation of this PR's `qr_rev` using this rule seems to outperform `qr_rev` in a simple benchmark while yielding the same result.

```julia
function qr_rev2(QR_::ChainRules.QR_TYPE, Q̄, R̄)
    Q, R = QR_
    Q = Matrix(Q)
    Q̄ = Q̄ isa Zero ? Q̄ : @view Q̄[:, axes(Q, 2)]
    V = R̄*R' - Q'*Q̄
    Ā = (Q̄ + Q * Hermitian(V)) / R'
    return Ā
end

julia> A = randn(4, 4);

julia> F = qr(A);

julia> ΔF = Composite{typeof(F)}(Q = randn(eltype(F.Q), size(Matrix(F.Q))), R = randn(eltype(F.R), size(F.R)));

julia> @btime ChainRules.qr_rev($F, $(ΔF.Q), $(ΔF.R));
  16.374 μs (58 allocations: 7.11 KiB)

julia> @btime qr_rev2($F, $(ΔF.Q), $(ΔF.R));
  4.323 μs (12 allocations: 2.27 KiB)

julia> ChainRules.qr_rev(F, ΔF.Q, ΔF.R) ≈ qr_rev2(F, ΔF.Q, ΔF.R)
true

julia> A = randn(10, 4);

julia> F = qr(A);

julia> ΔF = Composite{typeof(F)}(Q = randn(eltype(F.Q), size(Matrix(F.Q))), R = randn(eltype(F.R), size(F.R)));

julia> @btime ChainRules.qr_rev($F, $(ΔF.Q), $(ΔF.R));
  56.299 μs (134 allocations: 20.86 KiB)

julia> @btime qr_rev2($F, $(ΔF.Q), $(ΔF.R));
  5.112 μs (12 allocations: 3.39 KiB)

julia> ChainRules.qr_rev(F, ΔF.Q, ΔF.R) ≈ qr_rev2(F, ΔF.Q, ΔF.R)
true
```
Or is the rule you have implemented expected to be more numerically stable?

src/rulesets/LinearAlgebra/factorization.jl
sethaxen: Section 3.2 of https://arxiv.org/pdf/2009.10071.pdf gives an `rrule` for `qr` for wide matrices (`m<n`). I haven't tested it, though.

Kolaru: I will look into adding that.

Kolaru:
> Or is the rule you have implemented expected to be more numerically stable?

I have no idea. The two references we are using do not directly compare each other, and I do not know how to determine this myself.

sethaxen:
> > Or is the rule you have implemented expected to be more numerically stable?
> 
> I have no idea. The two references we are using do not directly compare each other, and I do not know how to determine this myself.

The article I linked in https://github.com/JuliaDiff/ChainRules.jl/pull/306#discussion_r528169013 (https://arxiv.org/pdf/2009.10071.pdf), which covers wide and tall matrices as well, also uses the simpler rule from the Seeger et al paper. While I didn't implement their rules for wide and tall matrices, I ended up using a similar approach for the LU decomposition of wide and tall matrices in #354. For these reasons, I'm thinking the Seeger approach is preferable.

Kolaru:
I finally had time to come back to this, and it has been kind of a nightmare, because QR decompositions are represented in a weird way that do not play nicely with the tests and the comparison with FiniteDifferences.

After quite a lot of experimentations, I gave up on trying to make everything work with the default type returned by `qr`. Instead I define a new custom type `ExplicitQR` that stores the `Q` and `R` matrices explicitely. I use this struct for the tests, and compare the end result with the one from the `qr` method to ensure the latter is correct as well.

I hope this is sufficient. Otherwise I must admit I am out of idea about what should be done to test `qr` directly.

I am aware of #469 that has a somewhat different approach. I am not currently sure which is better.

Alos I implemented the algorithm suggested by @sethaxen.

So provided my way of testing is okay, this should be ready.

sethaxen:
Thanks, @Kolaru! I'll review #469 first, then this, then I'll recommend how to proceed.

Kolaru:
I tried the following

```julia
using ChainRules: rrule, ExplicitQR
using LinearAlgebra

A = randn(10, 5)
Q, _ = qr(A)
v = randn(5)
w = randn(10)

F, F_pullback = rrule(qr, A)
Q, Q_pullback = rrule(getproperty, F, :Q)
y1, y1_pullback = rrule(*, Q, v)
y2, y2_pullback = rrule(*, Q, w)

ȳ1 = rand(10)
_, Q̄1 = y1_pullback(ȳ1)
_, F̄1 = Q_pullback(Q̄1)
_, Ā1 = F_pullback(F̄1) 

ȳ2 = rand(10)
_, Q̄2 = y2_pullback(ȳ2)
_, F̄2 = Q_pullback(Q̄2)
_, Ā2 = F_pullback(F̄2)
```
and everything seems to be fine (i.e. nothing error, I haven't tested correctness). Is this the correct way to test your point? Whether this PR or #469 is used, adding proper test for quirks of `Q` seems unavoidable.

Note that the `ExplicitQR` object I define is only used in the test (and maybe its definition should be moved to the tests for clarity).

As far as I understand, `ChainRules` should be able to handle the QR objects properly. The problem I had was to create the tangent objects for the tests. The main issues is that we are taking derivative with respect to fields that are not stored in the object directly[*]. After some tries, I just gave up and sidestepped the whole testing issue, hoping this is still enough to ensure correctness.

[*] The most common error I had was:
```
ArgumentError: Tangent fields do not match primal fields.
Tangent fields: (:Q,). Primal (LinearAlgebra.QRCompactWY{Float64, Matrix{Float64}}) fields: (:factors, :T)
```
Using a custom struct with explicit fields `Q` and `R` naturally solved it.

PR #309, opened 2020-11-18
Add GitHub Actions CI workflow

sethaxen:
This PR adds a GitHub Actions for CI. The goal is ultimately to replace Travis, though I recommend keeping both in parallel for a little while until we are sure we are happy with this.

cc @simeonschaub

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,5 +38,5 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       # fixes for coveralls pushed since last release v1.1.1
-      - run: julia --color=yes -e 'using Pkg; Pkg.add("Coverage#master"); using Coverage; Coveralls.submit(process_folder())'
+      - run: julia --color=yes -e 'using Pkg; Pkg.add(name="Coverage", rev="master"); using Coverage; Coveralls.submit(process_folder())'
         shell: bash
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,8 +19,8 @@ jobs:
           - '1'
         os:
           - ubuntu-latest
+          - macOS-latest
           - windows-latest
-          - macos-latest
         arch:
           - x86
           - x64
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,3 +40,5 @@ jobs:
       # fixes for coveralls pushed since last release v1.1.1
       - run: julia --color=yes -e 'using Pkg; Pkg.add(PackageSpec(name="Coverage", rev="master")); using Coverage; Coveralls.submit(process_folder())'
         shell: bash
+        env:
+          COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -37,8 +37,7 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
-      # fixes for coveralls pushed since last release v1.1.1
-      - run: julia --color=yes -e 'using Pkg; Pkg.add(PackageSpec(name="Coverage", rev="master")); using Coverage; Coveralls.submit(process_folder())'
-        shell: bash
-        env:
-          COVERALLS_TOKEN: ${{ secrets.COVERALLS_TOKEN }}
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: lcov.info
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,40 @@
+name: CI
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.0'
+          - '1'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v1
+        env:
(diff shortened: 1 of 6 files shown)


sethaxen:
Once everything is working well (including code coverage) for release versions, I'll add another workflow that allows failure for nightly, as well as a badge.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,40 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.0'
+          - '1'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
```
simeonschaub, line 21: Since GH Actions can run a lot more jobs in parallel, I wonder whether it would make sense to test on more platforms. Especially 32-bit could be useful, because `Int`s not always being 64 can sometimes be a bit of a gotcha.

simeonschaub (commented):
Awesome, thanks for doing this! Looks great! What is your plan for nightly tests, do you plan on moving them eventually?

.github/workflows/CI.yml
sethaxen: Certainly, yeah, good idea. Do you think it's worth also testing on MacOS and Windows? Once I have everything working for ubuntu, I can add more to the grid.

simeonschaub: I don't think it would hurt. Since quite a bit of stuff depends on ChainRules nowadays, I think it would be nice to have broad test coverage. I could imagine that some IO functions we mark as `@nondifferentiable`, are defined a bit differently on Windows, so it wouldn't be impossible for there to be issues depending on the OS.

sethaxen:
Coveralls appears to be poorly supported by Coverage outside of Travis. Can we switch to codecov?

oxinabox:
> Coveralls appears to be poorly supported by Coverage outside of Travis. Can we switch to codecov?

Sure, i can never remember which is which.
I know i like one slightly better than the other, but I don't know what it is.

nickrobinson251:
Thanks for this!!

> I recommend keeping both in parallel for a little

👍  

Before we turn off Travis, we need to figure out how to have GA run CI nightly and notify us of failures

i.e. the GA version of the current Travis Cron job and
```
notifications:
  email:
    recipients:
      - nightly-rse@invenia.ca
    on_success: never
    on_failure: always
    if: type = cron
```

sethaxen:
> Before we turn off Travis, we need to figure out how to have GA run CI nightly and notify us of failures
> 
> i.e. the GA version of the current Travis Cron job and

That shouldn't be hard to do. Is the cron job only run on master or also for PRs? Where is the cron job specified?

> ```
> notifications:
>   email:
>     recipients:
>       - nightly-rse@invenia.ca
>     on_success: never
>     on_failure: always
>     if: type = cron
> ```

This is trickier, as GA's notification options are more limited. I'll see if I can figure it out with a 3rd party action.

sethaxen:
> This is trickier, as GA's notification options are more limited. I'll see if I can figure it out with a 3rd party action.

Something like this seems to be the recommended approach: https://github.community/t/no-email-notification-on-schedule-cron-job-failure/119638/2. i.e. create a new e-mail address specifically for e-mailing these notifications, and add its password to secrets. This is really inconvenient.

nickrobinson251:
> Something like this seems to be the recommended approach: ...

Ooof. Thanks for looking into it! Maybe let's leave it for now? ...I wonder if we can just keeping travis for the nightly CI run only? Don't think we need to solve it in this PR. :)

sethaxen:
Okay, this PR is ready for review. A few notes:
- All 32-bit builds fail. This is because the `rrule` test for `logdet` is too stringent. But this is the default for `rrule_test`, which sets `atol` and `rtol` to 1e-9. Should this be relaxed in ChainRulesTestUtils.jl to use something more like the `isapprox` defaults, which account for the precision of the number type?
- Julia nightly versions currently fail due to imprecision of `BLAS.nrm2`
- codecov and coveralls are both set up. But IMO codecov is nicer with github actions, since it automatically combines the code coverage into one metric instead of giving one for every job. Potentially we could completely drop coveralls for github actions, just using it for Travis.
- Julia nightly versions currently do not contribute to code coverage.

oxinabox:
@mattBrzezinski  can i leave it to you to see this merged?
You know GitHub Actions better

.github/workflows/JuliaNightly.yml
```diff
@@ -0,0 +1,48 @@
+# CI for Julia nightly, separate workflow to avoid failing CI badge on nightly fail
+name: JuliaNightly
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *"
```
mattBrzezinski, line 9: Only comment is maybe include what this cron translates to.

mattBrzezinski (approved):

mattBrzezinski:
The changes make sense, not sure how the owners want to handle the failures of the builds.

sethaxen:
There are two different failures:
- `logdet`/`logabsdet` tests: the result looks right, but perhaps it's hard for FiniteDifferences to get high precision consistently for those rules, so I propose to slightly increase the tolerance for those tests
- `logabsgamma` test: this failure seems to have been introduced in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/115. The offending line is https://github.com/JuliaDiff/ChainRules.jl/blob/4274352c53bd885d9c2d053223656d9286bec724/test/rulesets/packages/SpecialFunctions.jl#L53
```julia
julia> Δz = Composite{Tuple{Float64, DoesNotExist}}(1.0, Zero())
Composite{Tuple{Float64, DoesNotExist}}(1.0, Zero())
FiniteDifferences v0.11.3 changed how `to_vec` processes this differential, which causes the failure:
julia> to_vec(Δz)[1] # v0.11.2
2-element Vector{Float64}:
 -0.06132606674859018
  0.0

julia> to_vec(Δz)[1] # v0.11.3
1-element Vector{Float64}:
 -0.06132606674859018
```
@willtebbutt can you comment on whether this is an FD bug or whether we need to use a different differential for the test here?

oxinabox:
> logdet/logabsdet tests: the result looks right, but perhaps it's hard for FiniteDifferences to get high precision consistently for those rules, so I propose to slightly increase the tolerance for those tests

Possibly we need to use a matrix with larger values to avoid issues near zero?
But yes, fine to increase atol and rtol, there is an argument for it to pass in to the testers

willtebbutt:
> @willtebbutt can you comment on whether this is an FD bug or whether we need to use a different differential for the test here?

The reported new behaviour is correct -- the previous behaviour was incorrect (`to_vec(::Zero)` should be an empty vector).

Since we're still treating `Integer`s as differentiable things with `to_vec(1) = [1]`, you should replace `Zero()` with `randn()` I think.

PR #311, opened 2020-11-19
Revamp Cholesky implementation

willtebbutt:
The current Cholesky implementation isn't fantastic, in that we're choosing fairly arbitrary blocksizes based on code that I hacked together almost 4 years ago.

This is probably a better approach (due to Seeger). It basically copies what Zygote currently does, but with the important improvement that the type of the cotangent w.r.t. the matrix `A` whose Cholesky is taken is either a `Symmetric / Hermitian` or `StridedMatrix`, depending on whether `A` is Symmetric / Hermitian / a StridedMatrix. As shown in the additional tests, this is important to ensure that composition `cholesky(Symmetric(X::StridedMatrix))` gives the same answer as `cholesky(X::StridedMatrix)`.

I'll remove Zygote's implementation once this is in.


sethaxen (changes requested):
This rule looks more sane than the previous one. Had a few suggestions, the most substantial being tweaks to also support complex `Hermitian` matrices.

This PR does two different things: 1) implements an improved cholesky rule for a subset of matrices (also adding support for keyword arguments!) and 2) removes a generic rule that would be applied to all matrices. I wonder if it is better to split those two things into separate PRs. The 2nd one is non-breaking, but the 1st one could potentially be a breaking change if users decorated some custom matrix type with Symmetric.

It seems like it would be straightforward to add the `frule`s for `cholesky` too. Do you think that's worth adding to this PR, or do you think ADing directly through `cholesky` with forward mode would be more performant?

@nmheim it would be great if you could look at this too, since this solves some of the issues raised in https://github.com/TuringLang/DistributionsAD.jl/pull/95, and it's also removing the internal function that DistributionsAD.jl relies on (but shouldn't).

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -70,20 +70,69 @@ end
...
+    function cholesky_SymHerm_pullback(Δ::Composite)
+        Ā, U = _cholesky_pullback_shared_code(C, Δ)
+        Ā = BLAS.trsm!('R', 'U', 'T', 'N', one(eltype(Ā)) / 2, U.data, Ā)
+        return NO_FIELDS, _symhermtype(A)(Ā), DoesNotExist()
+    end
+    return C, cholesky_SymHerm_pullback
+end
+
+function rrule(
+    ::typeof(cholesky), A::StridedMatrix{<:Real}, ::Val{false}=Val(false); check::Bool=true,
```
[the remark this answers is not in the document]
willtebbutt: I've gone with `BlasFloat` instead.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -70,20 +70,69 @@ end
...
+        Ā[diagind(Ā)] ./= 2
+        return (NO_FIELDS, UpperTriangular(Ā), DoesNotExist())
+    end
+    return C, cholesky_StridedMatrix_pullback
+end
+
+function _cholesky_pullback_shared_code(C, Δ)
+    issuccess(C) || throw(PosDefException(C.info))
+    U = C.U
+    Ū = Δ.U
```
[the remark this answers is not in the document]
willtebbutt: This should be handled by `getproperty` I believe -- it flips stuff appropriately for the `uplo` property of the `cholesky`. Since the user-facing interface no longer provides an option to set `uplo`, I think we might be fine. Does this seem reasonable to you?

willtebbutt:
This PR now depends on [this FiniteDifferences PR](https://github.com/JuliaDiff/FiniteDifferences.jl/pull/115).

@sethaxen thanks for the careful review. I've adopted most of your recommendations, but I'm reluctant to extend the implementation to Complex-valued things in this PR, mainly because testing it will be a bit awkward and I don't want to get it wrong. Also I was planning to limit the scope of this PR to maintaining / correcting existing functionality present in `ChainRules` / `Zygote`, rather than extending it.

> It seems like it would be straightforward to add the frules for cholesky too. Do you think that's worth adding to this PR, or do you think ADing directly through cholesky with forward mode would be more performant?

This would be a good thing to do, but my preference would again be to leave it for later.

> This PR does two different things: 1) implements an improved cholesky rule for a subset of matrices (also adding support for keyword arguments!) and 2) removes a generic rule that would be applied to all matrices. I wonder if it is better to split those two things into separate PRs. The 2nd one is non-breaking, but the 1st one could potentially be a breaking change if users decorated some custom matrix type with Symmetric.

Hmmm good point. I wonder whether it would have worked for custom matrices before though? For example, I think that the previous implementation probably implicitly restricted users to `StridedMatrix`s and their `HermOrSym` counterparts anyway because of the BLAS calls. Does this address your concerns, or do you think there's a possibility that I'm missing?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -70,20 +70,73 @@ end
...
+        Ā = BLAS.trsm!('R', 'U', 'C', 'N', one(eltype(Ā)), U.data, Ā)
+        idx = diagind(Ā)
+        @views Ā[idx] .= real.(Ā[idx]) ./ 2
+        return (NO_FIELDS, UpperTriangular(Ā), DoesNotExist())
+    end
+    return C, cholesky_pullback
+end
+
+function _cholesky_pullback_shared_code(C, ΔC)
+    issuccess(C) || throw(PosDefException(C.info))
```
sethaxen, line 129: As with diagonal, could you move this out of the shared function and then only throw this error if the user did not specify `check = false`?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -70,20 +70,73 @@ end
...
+    U = C.U
+    Ū = ΔC.U
+    Ā = similar(U.data)
+    Ā = mul!(Ā, Ū, U')
+    Ā = LinearAlgebra.copytri!(Ā, 'U', true)
```
sethaxen, line 134: I know complex matrices aren't officially supported in this PR, but I tested locally that this last fix makes them work for me for complex Hermitian matrices, for when the type constraints are relaxed.
```suggestion
    Ā = LinearAlgebra.copytri!(Ā, 'U', true)
    idx = diagind(Ā)
    @views Ā[idx] .= real(Ā[idx])
```

sethaxen (changes requested):
Just a few more minor suggestions.

> I'm reluctant to extend the implementation to Complex-valued things in this PR, mainly because testing it will be a bit awkward and I don't want to get it wrong. Also I was planning to limit the scope of this PR to maintaining / correcting existing functionality present in `ChainRules` / `Zygote`, rather than extending it.

Technically Zygote's rules catch complex `StridedMatrix`es. Looks like Zygote's rule was never tested for complex numbers, and I can guarantee they are not right, so I agree it is fine to only support real in this PR.

> Hmmm good point. I wonder whether it would have worked for custom matrices before though? For example, I think that the previous implementation probably implicitly restricted users to `StridedMatrix`s and their `HermOrSym` counterparts anyway because of the BLAS calls. Does this address your concerns, or do you think there's a possibility that I'm missing?

It looks like the old rules supported generic matrices by converting them to `Matrix`es before passing to the function with the BLAS calls. So it looks like it would have worked for a custom matrix before, so long as the number type was a `BlasFloat`.

In the past, when restricting types of rules, have we marked those as breaking changes? If so, we should do the same here.

src/rulesets/LinearAlgebra/factorization.jl
sethaxen: Completely optional though, because some of the other `rrule`s wouldn't support complex either.

sethaxen (approved):
Feel free to merge whenever you're happy with version number and such. Thanks!

willtebbutt:
Thanks for the feedback @sethaxen . I'm going to stop development on this branch until CI is back up and running, then will process your feedback.

Issue #319, opened 2020-12-04
Drop SpecialFunctions dependency


stevengj:
The next release of SpecialFunctions will include ChainRulesCore rules: JuliaMath/SpecialFunctions.jl#238

Do we need to synchronize releases to prevent Julia from complaining about overwritten method definitions?  That is, once we update SpecialFunctions, do a new release of ChainRules shortly afterwards?

cc @devmotion

devmotion:
One could follow @oxinabox's suggestion in https://github.com/JuliaMath/SpecialFunctions.jl/pull/238#issuecomment-692900690 and introduce a check in ChainRules and releasing it before the next release of SpecialFunctions.

oxinabox:
We should do that then next breaking release of ChainRules.jl drop the dependency and the rules.

PR #320, opened 2020-12-04
Only define rules for SpecialFunctions if needed

devmotion:
This PR addresses https://github.com/JuliaDiff/ChainRules.jl/issues/319 by only defining rules for SpecialFunctions if SpecialFunctions.ChainRulesCore does not exist, as suggested by @oxinabox in https://github.com/JuliaMath/SpecialFunctions.jl/pull/238#issuecomment-692900690.

Should the same condition be added to the tests? And do you want me to bump the version?


oxinabox:
Yes let's add this to the tests and bump the version also

It will be nice for the tests and we won't see the deprecation warnings.

Can you also open an issue about dropping the SpecialFunctions dependency  and cross link that in the code?

stevengj:
Great!  Ideally, I'd like to delay tagging a new SpecialFunctions release until this is merged and tagged in ChainRules, so it would be good if you could do a 0.7.35 release with this in the near future.

oxinabox:
Will do tomorrow

oxinabox (approved):

PR #321, opened 2020-12-04
Add eigen and eigvals rules for StridedMatrix

sethaxen:
This PR adds rules for `eigen` and `eigvals`. It is very conservative and only applies for `StridedMatrix`es. Because the normalization and phase convention for `eigen` for these matrices is known, the rules can explicitly account for this, which allows us to stably test with FiniteDifferences.

A few notes:
- The rule doesn't handle degenerate matrices correctly. This can perhaos be handled in a future PR with the approach mentioned in #144 (perhaps in #179 or a follow-up). The methods that handle the normalization/phase contribution could be directly applied to that rule.
- The rule doesn't do the right thing for non-diagonal hermitian matrices, as those dispatch to `eigen(::Hermitian)`, which uses a different phase convention. That is, the rule will work in a program that is invariant to the phase convention but not one that isn't. This will be handled in a future PR for `eigen(::Hermitian)` (currently defined in #193 but will be split into its own PR)


willtebbutt:
This is great.

Would you mind adding some `@inferred` tests? I've recently come to realise that one source of performance issues in Zygote is type-instabilities in the primitives, so it would be good to start trying to clamp down on that.

sethaxen:
> Would you mind adding some `@inferred` tests? I've recently come to realise that one source of performance issues in Zygote is type-instabilities in the primitives, so it would be good to start trying to clamp down on that.

I can do that. Perhaps it would be good to add this to the testers in ChainRulesTestUtils with a configurable option like `check_inferred`, on by default.

willtebbutt:
> I can do that. Perhaps it would be good to add this to the testers in ChainRulesTestUtils with a configurable option like check_inferred, on by default.

Yes. This would be an excellent idea.

willtebbutt (approved):
I'm happy with this, modulo a patch bump.

sethaxen:
Thanks! All tests pass so I'll leave open for a few days to give others a chance to review and then merge on Tuesday.

sethaxen:
I changed the `frule`s to target the more basic mutating `eigen!` and `eigvals!`, so the test suite now needs https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/79 to pass, and this should be re-reviewed.

willtebbutt (approved):
This looks great, and should be merged with a patch bump.

PR #322, opened 2020-12-06
Move Symmetric/Hermitian rules and tests to own file

sethaxen:
As proposed in https://github.com/JuliaDiff/ChainRules.jl/pull/193#discussion_r426280828, this PR just moves the rules for functions of `Symmetric`/`Hermitian` to their own file and does the same for the tests. Not only does this mirror the structure of LinearAlgebra (where these methods are defined in `symmetric.jl`), but with #193 and as mentioned in #321, more specializations are coming for these matrices, and this will make `structured.jl` less cluttered.

This PR does not make any changes to functionality and only exists so that these future PRs are cleanly separated from the reorg. As such I don't think it needs a bump to version number.


willtebbutt:
Seems reasonable to me.

Regarding version number: to my mind it still makes sense to bump the patch version -- my understanding of our policy is that any changes to `src` or `test` get a patch bump. It doesn't really cost us anything, so might as well 🤷

willtebbutt (approved):

PR #323, opened 2020-12-06
Add rules for Symmetric/Hermitian eigen and svd

sethaxen:
This PR adds `frule`s for `eigen!` and `eigvals!` and `rrule`s for `eigen`, `eigvals`, `svd`, and `svdvals` for `Hermitian` or real `Symmetric` inputs. The rules for the `eigen` functions are similar to those in #321 but are more efficient due to known types and unitary eigenvectors. The `svd` functions are added because their primals in base call mutating functions, which will miss the `eigen` rules, hence we don't need to add corresponding `frule`s.

This PR will eventually also fix the normalization convention issue noted in #321, so it depends on that PR.


src/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -77,3 +72,201 @@ function _hermitian_back(ΔΩ::LinearAlgebra.AbstractTriangular, uplo)
...
+##### `svd`
+#####
+
+function rrule(::typeof(svd), A::LinearAlgebra.RealHermSymComplexHerm{<:BLAS.BlasReal,<:StridedMatrix})
+    F = svd(A)
+    function svd_pullback(ΔF::Composite{<:SVD})
+        U, V = F.U, F.V
+        c = _svd_eigvals_sign!(similar(F.S), U, V)
+        λ = F.S .* c
+        ∂λ = ΔF.S isa AbstractZero ? ΔF.S : ΔF.S .* c
```
sethaxen, line 209: For some reason this is type-unstable:
```julia
julia> using ChainRules, LinearAlgebra

julia> F = svd(randn(4, 4));

julia> ∂F = Composite{typeof(F)}(U = randn(4, 4), V = randn(4, 4)', Vt = randn(4, 4), S = randn(4));

julia> foo(ΔF) = ΔF.U, ΔF.S, ΔF.V, ΔF.Vt;

julia> @inferred foo(∂F)
ERROR: return type Tuple{Array{Float64,2},Array{Float64,1},Adjoint{Float64,Array{Float64,2}},Array{Float64,2}} does not match inferred return type Tuple{Array{Float64,2},Union{Zero, Array{Float64,1}},Union{Zero, Adjoint{Float64,Array{Float64,2}}},Union{Zero, Array{Float64,2}}}
```

willtebbutt:
```julia
julia> @code_warntype getproperty(ChainRulesCore.backing(∂F), :U)
Variables
  #self#::Core.Compiler.Const(getproperty, false)
  x::NamedTuple{(:U, :V, :Vt, :S),Tuple{Array{Float64,2},Adjoint{Float64,Array{Float64,2}},Array{Float64,2},Array{Float64,1}}}
  f::Symbol

Body::Union{Adjoint{Float64,Array{Float64,2}}, Array}
1 ─ %1 = Base.getfield(x, f)::Union{Adjoint{Float64,Array{Float64,2}}, Array}
└──      return %1
```
Looks like this call to `getfield` isn't inferring properly. Possibly something to do with constant-folding or something...

willtebbutt: Looking at this:
```julia
julia> @code_warntype getproperty(∂F, :U)
Variables
  #self#::Core.Compiler.Const(getproperty, false)
  comp::Composite{SVD{Float64,Float64,Array{Float64,2}},NamedTuple{(:U, :V, :Vt, :S),Tuple{Array{Float64,2},Adjoint{Float64,Array{Float64,2}},Array{Float64,2},Array{Float64,1}}}}
  idx::Symbol

Body::Union{Zero, Adjoint{Float64,Array{Float64,2}}, Array}
1 ─ %1 = (idx ∈ $(Expr(:static_parameter, 2)))::Bool
└──      goto #3 if not %1
2 ─      goto #4
3 ─ %4 = ChainRulesCore.Zero()::Core.Compiler.Const(Zero(), false)
└──      return %4
4 ─ %6 = ChainRulesCore.backing(comp)::NamedTuple{(:U, :V, :Vt, :S),Tuple{Array{Float64,2},Adjoint{Float64,Array{Float64,2}},Array{Float64,2},Array{Float64,1}}}
│   %7 = ChainRulesCore.getproperty(%6, idx)::Union{Adjoint{Float64,Array{Float64,2}}, Array}
│   %8 = ChainRulesCore.unthunk(%7)::Union{Adjoint{Float64,Array{Float64,2}}, Array}
└──      return %8
```
it looks like there's also an issue in that it's not able to realise that the `Zero` path won't get hit...

willtebbutt: @oxinabox what are we missing here? This feels like something basic...

sethaxen: Interestingly, the above example doesn't seem to be an issue on Julia nightly.

willtebbutt:
@sethaxen sorry, I dropped the ball on reviewing this. What's the status at the minute? It looks like CI is now passing.

willtebbutt:
Also, related reference (not sure if you've seen Seth): https://arxiv.org/abs/1909.02659

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -92,21 +96,36 @@ function frule((_, ΔA), ::typeof(eigen!), A::StridedMatrix{T}; kwargs...) where
 end
 
 function rrule(::typeof(eigen), A::StridedMatrix{T}; kwargs...) where {T<:Union{Real,Complex}}
-    F = eigen(A; kwargs...)
+    # NOTE: this check for hermitian-ness occurs in `eigen!`. We here do it in the rrule for
+    # `eigen` so that this works for non-mutating AD
+    isherm = ishermitian(A)
+    if isherm
+        hermA, back_Hermitian = rrule(Hermitian, A, :U)
```
sethaxen, line 103: The `rrule` is only defined for `Hermitian{<:BlasReal}`. Should we also restrict `T` here to `BlasReal`? Or in a future PR?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -92,21 +96,36 @@ function frule((_, ΔA), ::typeof(eigen!), A::StridedMatrix{T}; kwargs...) where
...
+        F = eigen(A; kwargs...)
+    end
     function eigen_pullback(ΔF::Composite{<:Eigen})
-        λ, V = F.values, F.vectors
-        Δλ, ΔV = ΔF.values, ΔF.vectors
-        if ΔV isa AbstractZero
-            Δλ isa AbstractZero && return (NO_FIELDS, Δλ + ΔV)
-            ∂K = Diagonal(Δλ)
-            ∂A = V' \ ∂K * V'
+        if isherm
```
sethaxen, line 109: Adding this check causes the rrule to become type-unstable. Not certain how to do this type-stably.

sethaxen (commented):
> @sethaxen sorry, I dropped the ball on reviewing this. What's the status at the minute? It looks like CI is now passing.

No worries! This PR is almost complete and is ready for review. Here's what's missing:
- It would be nice to know why we have the issue in https://github.com/JuliaDiff/ChainRules.jl/pull/323#discussion_r539259034, though resolving it isn't necessary for this PR
- I need to add a test for `eigen(A::StridedMatrix)` when `A` is hermitian
- ~resolve type-instability added by considering hermitian case of `eigen(A::StridedMatrix)` (why CI is now failing)~
- Add references (there are references for everything but handling of the phase. I'm actually working on a blog post that discusses that part I could reference when it's out).

> Also, related reference (not sure if you've seen Seth): https://arxiv.org/abs/1909.02659

Cool, yeah looks like they explicitly accounted for gauge, which is usually ignored for `eigen` and `svd`. Shouldn't be an issue in this case, because I account for the gauge in the rrule for `eigen`. The result should be more efficient than the rule in their reference, and the test suite confirms that it works for gauge-dependent functions.

src/rulesets/LinearAlgebra/factorization.jl
sethaxen: I think I figured it out

src/rulesets/LinearAlgebra/factorization.jl
sethaxen: Outdated since I now bypass the `rrule`

sethaxen:
In the interest of getting this merged, I have added I think sufficient comments and references. A later PR can always add more about phase/normalization conventions. @willtebbutt could you re-review?

sethaxen:
The CI errors are unrelated to this PR and due to https://github.com/JuliaDiff/FiniteDifferences.jl/issues/133.

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -41,4 +41,241 @@
...
+            asnt(E::Eigen) = (values=E.values, vectors=E.vectors)
+
+            function _eigen_stable(A)
+                F = eigen(A)
+                rmul!(F.vectors, _eigvecs_stabilize_mat(F.vectors, A.uplo))
+                return F
+            end
+
+            n = 10
+            @testset "eigen!(::$SymHerm{$T}) uplo=$uplo" for SymHerm in (Symmetric, Hermitian),
```
willtebbutt, line 68: Could you indicate that this only tests `frule`s, preferably in the testset label, but alternatively as a comment?

test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -41,4 +41,241 @@
...
+                @test ∂F_ad.values ≈ ∂F_fd.values
+
+                f_stable = x -> asnt(_eigen_stable(SymHerm(x, uplo)))
+                F_stable = f_stable(A)
+                ∂F_stable_fd = jvp(_fdm, f_stable, (A, ΔA))
+                C = _eigvecs_stabilize_mat(F.vectors, uplo)
+                @test ∂F_ad.vectors * C ≈ ∂F_stable_fd.vectors
+            end
+
+            @testset "eigen(::$SymHerm{$T}) uplo=$uplo" for SymHerm in (Symmetric, Hermitian),
```
willtebbutt, line 95: Same here that this just tests `rrule`s?

willtebbutt (approved):
I'm not going to pretend that I follow the details of the implementations, but I'm satisfied that what is implemented has been tested properly, so I'm happy to approve.

I've requested a couple of additional bits of documentation in the test sets because I had to squint to figure out what was being tested in each block and, since this is a comparatively complicated new bit of code, I feel that signposting what's going on is quite important. Would you mind making similar readability improvements to all of the testsets that you've added?

sethaxen:
Done! Thanks!

PR #324, opened 2020-12-06
Some Adjoint and Transpose stuff

willtebbutt:
This PR pretty much brings us up to feature-parity with Zygote on the rules for `Adjoint` and `Transpose`.

TODO:
- [x] Enable the use of `Composite`s in the reverse-pass
- [x] Make the various transpose-like functions play nicely with  complex numbers.
- [ ] ~~Make `parent` return a `Composite`, rather than the transposition of `ȳ`, since `ȳ` might itself be e.g. a `Composite`.~~
- [ ] ~~Test `parent`.~~

edit: the second and third points weren't achievable in this PR because our testing infrastructure isn't up to scratch at the minute. Opened [this issue](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/80) regarding the problem.

### DIFF
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -92,35 +92,31 @@ end
 
 # ✖️✖️✖️TODO: Deal with complex-valued arrays as well
 function rrule(::Type{<:Adjoint}, A::AbstractMatrix{<:Union{Real, Complex}})
-    function Adjoint_pullback(ȳ)
-        return (NO_FIELDS, adjoint(ȳ))
-    end
+    Adjoint_pullback(ȳ::Composite) = (NO_FIELDS, ȳ.parent)
+    Adjoint_pullback(ȳ::AbstractMatrix) = (NO_FIELDS, adjoint(ȳ))
     return Adjoint(A), Adjoint_pullback
 end
 
 function rrule(::Type{<:Adjoint}, A::AbstractVector{<:Union{Real, Complex}})
-    function Adjoint_pullback(ȳ)
-        return (NO_FIELDS, vec(adjoint(ȳ)))
-    end
+    Adjoint_pullback(ȳ::Composite) = (NO_FIELDS, vec(ȳ.parent))
+    Adjoint_pullback(ȳ::AbstractMatrix) = (NO_FIELDS, vec(adjoint(ȳ)))
     return Adjoint(A), Adjoint_pullback
 end
 
 function rrule(::typeof(adjoint), A::AbstractMatrix{<:Union{Real, Complex}})
-    function adjoint_pullback(ȳ)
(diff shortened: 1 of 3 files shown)


src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -91,63 +91,55 @@ end
 #####
 
 # ✖️✖️✖️TODO: Deal with complex-valued arrays as well
-function rrule(::Type{<:Adjoint}, A::AbstractMatrix{<:Real})
-    function Adjoint_pullback(ȳ)
-        return (NO_FIELDS, adjoint(ȳ))
-    end
+function rrule(::Type{<:Adjoint}, A::AbstractMatrix{<:Union{Real, Complex}})
```
nickrobinson251, line 94: does the comment above need adjusting given we know allow `Complex` eltype?

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -90,64 +90,55 @@ end
 ##### `Adjoint`
 #####
 
-# ✖️✖️✖️TODO: Deal with complex-valued arrays as well
-function rrule(::Type{<:Adjoint}, A::AbstractMatrix{<:Real})
-    function Adjoint_pullback(ȳ)
-        return (NO_FIELDS, adjoint(ȳ))
-    end
+function rrule(::Type{<:Adjoint}, A::AbstractMatrix{<:Union{Real, Complex}})
```
sethaxen, line 93: The `Adjoint`/`Transpose` functions in base work even with non-real/complex types, e.g. `Quaternion`s. Given that the pushforwards and pullbacks just call adjoint themselves and there, aren't these rules a good example of where it is safe to drop the type constraint?

sethaxen (approved):
Overall, this looks right. It looks like this is causing some relevant integration errors with Zygote. Once those are fixed (or if turns out to be a Zygote bug), feel free to bump the version number and merge.

src/rulesets/LinearAlgebra/structured.jl
willtebbutt: Turns out I had forgotten to account for `rrule(adjoint, ::Adjoint{T, Vector{T})` (and for `transpose`). Tests are now in, so hopefully this is now fixed 🤞

sethaxen: I don't follow. There is still from what I can tell not a rule for  `rrule(::Type{<:Adjoint}, A::AbstractMatrix{T}) where {T<:Number}`

willtebbutt: 🤦 I mean to reply to your comment about Zygote integration errors. I should probably go to bed...

You're correct, I've still not resolved this.

willtebbutt: I've pushed a change. Should be sorted.

PR #325, opened 2020-12-08
WIP matmul on arrays of arrays and fix Real*Array{<:Complex}`

oxinabox:
Closes https://github.com/FluxML/Zygote.jl/issues/837

But I ran into issues.
Seems like we were not testing `*(::Real, ::Array{Compplex})` which gives an incorrect anser for the first argument.
At least I sassume it is incorrect since it is a complex number.

Also needed to add a special case for not inplacing as one is broken
ttps://github.com/JuliaLang/julia/issues/38772

Also the CoVector one deems to not work.

It would be super useful if someone who really understand the complex number stuff (@sethaxen, @simeonschaub) are able to take over this PR, at least as far as getting the existing complex number tests passing.

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -19,10 +19,18 @@ end
 ##### `*`
 #####
 
+# Things act enough like numbers for matmal etc to apply to matrixes with this eltype
+# inparticular it also includes AbstractArrays, which can be a problem since operations on
+# AbstractArrays in general are often better left for the AD to decompose then to use a
+# High-level general-rule. f/rrules should only accept this with caution, but it is useful
+# for especially Zygote which stuggles to deal with generic code that might be mutating
+const MatMulField = Union{CommutativeMulNumber, AbstractArray{<:CommutativeMulNumber}}
+
+
 function rrule(
     ::typeof(*),
-    A::AbstractVecOrMat{<:CommutativeMulNumber},
-    B::AbstractVecOrMat{<:CommutativeMulNumber},
+    A::AbstractVecOrMat{<:MatMulField},
+    B::AbstractVecOrMat{<:MatMulField},
 )
     function times_pullback(Ȳ)
         return (
@@ -40,6 +48,9 @@ function rrule(
     return A * B, times_pullback
 end
 
+# Good inplacable version so longer as arrays of arrays are not involved
+# TODO: remove this after https://github.com/JuliaLang/julia/issues/38772
+# and make the above accept any MatMulField
 function rrule(
     ::typeof(*),
     A::AbstractVector{<:CommutativeMulNumber},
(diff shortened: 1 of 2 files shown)


sethaxen:
In general we really don't consistently enforce that cotangents of reals are real if they interact with a complex number, i.e. in most places we do the Zygote thing. There are a few places where I've explicitly added projections to real for new or improved rules, following https://github.com/JuliaDiff/ChainRulesCore.jl/issues/176, e.g. the rules for `abs` (or any scalar rule that calls `_realconjtimes`), the cotangents of the eigenvalues of `Hermitian` matrices, etc.

If we choose to go this way, which I think we should at least for real/complex, then it should probably be handled in its own PR. I haven't looked at this PR in detail yet, but I think the easy thing to do right now is just test all real and all complex.

oxinabox:
OK, I can drop that test, for now
Can you open an issue about that you said for enforcing?

src/rulesets/Base/arraymath.jl
```diff
@@ -62,13 +73,30 @@ function rrule(
...
+# Not inplace for if Arrays of Arrays are involved
+# TODO: remove this after https://github.com/JuliaLang/julia/issues/38772
+# and make the above accept any MatMulField
+function rrule(
+    ::typeof(*),
+    A::AbstractVector{<:MatMulField},
+    B::AbstractMatrix{<:MatMulField},
+)
+    function times_pullback(Ȳ)
+        @assert size(B, 1) === 1   # otherwise primal would have failed.
```
sethaxen, line 85: Is this necessary? Because the primal already did this check.

src/rulesets/Base/arraymath.jl
```diff
@@ -62,13 +73,30 @@ function rrule(
...
+        return NO_FIELDS, @thunk(Ȳ * vec(B')), @thunk(A' * Ȳ)
+    end
+    return A * B, times_pullback
+end
+
+
+
 function rrule(
-   ::typeof(*), A::CommutativeMulNumber, B::AbstractArray{<:CommutativeMulNumber}
+   ::typeof(*), A::CommutativeMulNumber, B::AbstractArray{<:MatMulField}
```
sethaxen, line 94: Would it be notationally clearer to make `A` lower case since it's a scalar?

src/rulesets/Base/arraymath.jl
```diff
@@ -62,13 +73,30 @@ function rrule(
...
 )
     function times_pullback(Ȳ)
         return (
             NO_FIELDS,
-            @thunk(dot(Ȳ, B)'),
+            (dot(Ȳ, B)'),
```
sethaxen, line 99: Swapping the order in `dot` conjugates the output, e.g.
```julia
julia> x, y = randn(ComplexF64, 3), randn(ComplexF64, 3);

julia> dot(x, y) == dot(y, x)'
true
```
Could make the same change in the below rule

```suggestion
            dot(B, Ȳ),
```

src/rulesets/Base/arraymath.jl
```diff
@@ -19,10 +19,18 @@ end
...
+# for especially Zygote which stuggles to deal with generic code that might be mutating
+const MatMulField = Union{CommutativeMulNumber, AbstractArray{<:CommutativeMulNumber}}
+
+
 function rrule(
     ::typeof(*),
-    A::AbstractVecOrMat{<:CommutativeMulNumber},
-    B::AbstractVecOrMat{<:CommutativeMulNumber},
+    A::AbstractVecOrMat{<:MatMulField},
+    B::AbstractVecOrMat{<:MatMulField},
```
sethaxen, line 33: Is there a test for this rule?

src/rulesets/Base/arraymath.jl
```diff
@@ -62,13 +73,30 @@ function rrule(
...
 end
 
+# Not inplace for if Arrays of Arrays are involved
+# TODO: remove this after https://github.com/JuliaLang/julia/issues/38772
+# and make the above accept any MatMulField
+function rrule(
+    ::typeof(*),
+    A::AbstractVector{<:MatMulField},
+    B::AbstractMatrix{<:MatMulField},
+)
```
sethaxen, line 83: Is there a test for this rule?

sethaxen (approved):
All conjugation looks correct. I made some minor suggestions, and it looks like a few tests are missing. Once tests pass and coverage is good, feel free to merge.

Issue #328, opened 2020-12-15
BLAS.gemm and BLAS.gemv pullbacks are type-unstable


sethaxen:
As identified by https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/78, the pullbacks for `BLAS.gemm` and `BLAS.gemv` are type-unstable because a different `InplaceableThunk` is returned based on the value of the input parameters. For the `norm` functions, I worked around this by pushing the condition check to within the thunk, which in this case looks like it will produce a lot of duplicate code. Do we expect that this type-instability to cause decreased efficiency?

PR #329, opened 2020-12-15
Fix various type-instabilities

sethaxen:
This PR fixes most of the type-instabilities identified by https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/78, the remainder being documented in #328 and https://github.com/JuliaDiff/ChainRulesCore.jl/issues/265 or for type-unstable primals. I'm not yet adding any new tests because that would be redundant with the above PR. Until it is merged, this PR just checks that our current functionality tests still pass.

### DIFF
--- a/src/rulesets/LinearAlgebra/blas.jl
+++ b/src/rulesets/LinearAlgebra/blas.jl
@@ -17,13 +17,15 @@ rrule(::typeof(BLAS.dot), x, y) = rrule(dot, x, y)
 
 function rrule(::typeof(BLAS.dot), n, X, incx, Y, incy)
     Ω = BLAS.dot(n, X, incx, Y, incy)
-    function blas_dot_pullback(::Zero)
-        return (NO_FIELDS, DoesNotExist(), Zero(), DoesNotExist(), Zero(), DoesNotExist())
-    end
     function blas_dot_pullback(ΔΩ)
-        ΔΩ = extern(ΔΩ)
-        ∂X = @thunk scal!(n, ΔΩ, blascopy!(n, Y, incy, _zeros(X), incx), incx)
-        ∂Y = @thunk scal!(n, ΔΩ, blascopy!(n, X, incx, _zeros(Y), incy), incy)
+        if ΔΩ isa Zero
+            ∂X = Zero()
+            ∂Y = Zero()
+        else
+            ΔΩ = extern(ΔΩ)
+            ∂X = @thunk scal!(n, ΔΩ, blascopy!(n, Y, incy, _zeros(X), incx), incx)
+            ∂Y = @thunk scal!(n, ΔΩ, blascopy!(n, X, incx, _zeros(Y), incy), incy)
+        end
         return (NO_FIELDS, DoesNotExist(), ∂X, DoesNotExist(), ∂Y, DoesNotExist())
     end
     return Ω, blas_dot_pullback
--- a/src/rulesets/LinearAlgebra/blas.jl
+++ b/src/rulesets/LinearAlgebra/blas.jl
@@ -17,15 +17,13 @@ rrule(::typeof(BLAS.dot), x, y) = rrule(dot, x, y)
 
 function rrule(::typeof(BLAS.dot), n, X, incx, Y, incy)
     Ω = BLAS.dot(n, X, incx, Y, incy)
+    function blas_dot_pullback(::Zero)
+        return (NO_FIELDS, DoesNotExist(), Zero(), DoesNotExist(), Zero(), DoesNotExist())
+    end
     function blas_dot_pullback(ΔΩ)
-        if ΔΩ isa Zero
-            ∂X = Zero()
-            ∂Y = Zero()
-        else
-            ΔΩ = extern(ΔΩ)
-            ∂X = @thunk scal!(n, ΔΩ, blascopy!(n, Y, incy, _zeros(X), incx), incx)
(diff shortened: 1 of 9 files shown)


src/rulesets/LinearAlgebra/blas.jl
```diff
@@ -17,15 +17,13 @@ rrule(::typeof(BLAS.dot), x, y) = rrule(dot, x, y)
...
     function blas_dot_pullback(ΔΩ)
-        if ΔΩ isa Zero
-            ∂X = Zero()
-            ∂Y = Zero()
-        else
-            ΔΩ = extern(ΔΩ)
-            ∂X = @thunk scal!(n, ΔΩ, blascopy!(n, Y, incy, _zeros(X), incx), incx)
-            ∂Y = @thunk scal!(n, ΔΩ, blascopy!(n, X, incx, _zeros(Y), incy), incy)
-        end
+        ΔΩ = extern(ΔΩ)
```
oxinabox, line 24: Its not part of this PR But
This line should not have been here.
```suggestion
```

src/rulesets/LinearAlgebra/blas.jl
```diff
@@ -17,15 +17,13 @@ rrule(::typeof(BLAS.dot), x, y) = rrule(dot, x, y)
 
 function rrule(::typeof(BLAS.dot), n, X, incx, Y, incy)
     Ω = BLAS.dot(n, X, incx, Y, incy)
+    function blas_dot_pullback(::Zero)
+        return (NO_FIELDS, DoesNotExist(), Zero(), DoesNotExist(), Zero(), DoesNotExist())
+    end
     function blas_dot_pullback(ΔΩ)
-        if ΔΩ isa Zero
```
oxinabox, line 21: why is this type unstable?
I mean we don't need this branch because its always applicable to do this.
So its not something to include in the rule.
but i want to understand.

src/rulesets/LinearAlgebra/utils.jl
```diff
@@ -27,3 +27,5 @@ function _eyesubx!(X::AbstractMatrix)
 end
 
 _extract_imag(x) = complex(0, imag(x))
+
+_unionallof(::Type{T}) where {T} = T.name.wrapper
```
oxinabox, line 31: This name is not obvious to me why we say that.
would
`_unionall_wrapper` be clearear?
Can we also add a docstring?

src/rulesets/LinearAlgebra/norm.jl
```diff
@@ -181,7 +182,11 @@ function rrule(
     return y, norm1_pullback
 end
 
-_norm1_back(x, y, Δy) = sign.(x) .* real(Δy)
+function _norm1_back(x, y, Δy)
+    ∂x = similar(x)
+    ∂x .= sign.(x) .* real(Δy)
+    return ∂x
+end
```
oxinabox, line 189: It feels bad that we have to do this.
Should we open an issue on JuliaLang/julia about it?

oxinabox (approved):
Awesome work.
A few points but once those a cleared up, merge when ready.
Thanks

src/rulesets/LinearAlgebra/norm.jl
sethaxen: No, I think this is necessarily type-unstable, e.g.
```julia
julia> x = Diagonal([1.0, 2.0, 3.0]);

julia> x .* 0
3×3 Diagonal{Float64,Array{Float64,1}}:
 0.0   ⋅    ⋅
  ⋅   0.0   ⋅
  ⋅    ⋅   0.0

julia> x .* 10
3×3 Diagonal{Float64,Array{Float64,1}}:
 10.0    ⋅     ⋅
   ⋅   20.0    ⋅
   ⋅     ⋅   30.0

julia> x .* Inf
3×3 Array{Float64,2}:
  Inf  NaN   NaN
 NaN    Inf  NaN
 NaN   NaN    Inf

julia> x .* NaN
3×3 Array{Float64,2}:
 NaN  NaN  NaN
 NaN  NaN  NaN
 NaN  NaN  NaN
```

src/rulesets/LinearAlgebra/utils.jl
sethaxen: > would
> `_unionall_wrapper` be clearear?

Yeah naming this is tricky. It's basically `Base.typename` but instead of returning a `TypeName` it returns a `UnionAll`. If `_unionall_wrapper` is clearer, I can do that.

src/rulesets/LinearAlgebra/blas.jl
sethaxen: Oh, actually, it looks like `extern` caused the type-instability in the type of `InplaceableThunk`. That seems like it should not happen? But I can revert my change and accept the one you provided.

oxinabox: Reassigning a variable to contain another type causes type instablilties.
I think it shouldn't (I think the compiler should declare a second slot if that can happen and swap to using that but i guess it is hard to workout in adance, possibly impossible).
This basically makes the slow dynamically types
Often the compilere does compile it away.
But not always.

src/rulesets/LinearAlgebra/norm.jl
mcabbott: Note that this breaks `gradient(norm, [1,2,3])`.

PR #330, opened 2020-12-18
Notificiations to Invenia slack on build fail

oxinabox:
These sets it up so an RSE at invenia will be notified if nightly build is broken.
So we w can fix it

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,7 +5,7 @@ on:
     tags: [v*]
   pull_request:
   schedule:
-    - cron: "0 0 * * *" # run on default branch every night at midnight UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -66,3 +66,19 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel-finished: true
+
+  slack:
+    name: Notify Slack Failure
+    needs: test
+    runs-on: ubuntu-latest
+    if: github.event == 'schedule'
```
oxinabox, line 74:
```suggestion
    if:  always() && github.event == 'schedule'
```

fchorney, line 74:
```suggestion
    if: always() && github.event_name == 'schedule'
```

fchorney (approved):

PR #333, opened 2020-12-23
remove FD and CRTestUtils from deps

### LINKED ISSUE #332: Remove ChainRulesTestUtils and FiniteDifferences from direct dependencies
Having these as package dependencies and not test dependencies makes CR a heavier dependency to have.
From what I understand they are in the deps only to be able to specify versions? If that's the case they could be replaced with a test/Project.toml

matbesancon:
Closes #332

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,9 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.9.21"
+ChainRulesTestUtils = "0.5.5"
 Compat = "3"
+FiniteDifferences = "0.11.4"
 Reexport = "0.2"
 Requires = "0.5.2, 1"
 julia = "1"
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -14,9 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.9.21"
-ChainRulesTestUtils = "0.5.5"
 Compat = "3"
-FiniteDifferences = "0.11.4"
```
oxinabox, line 19: these lines should not be removed.
They still function to bind the version of things listed in `[extras]`

matbesancon: >  They still function to bind the version of things listed in [extras]

TIL. Pkg documentation is really lacking some bits on test dependencies.

matbesancon: done

oxinabox (approved):

PR #340, opened 2021-01-07
Change eigen and svd pullbacks to accept Composite{Any}

sethaxen:
In https://github.com/FluxML/Zygote.jl/pull/875, I discovered that Zygote passes `Composite{Any}` types to our `rrule`s. #321 and #323 added rrules whose pullbacks were constrained to `Composite{<:Eigen}` and `Composite{<:SVD}`, which would cause Zygote to error. This PR changes those signatures to `::Composite`. It adds no new tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.44"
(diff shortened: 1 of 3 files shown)


willtebbutt (approved):
Ah, yeah, this is a thing at the minute. We should be able to constrain again in the future.

PR #341, opened 2021-01-11
Increase FiniteDifferences to v0.12

wesselb:
* Changes the compat entry of FiniteDifferences from `0.11.4` to `0.11.4, 0.12`.
* ~Changes the compat entry of ChainRulesTestUtils from `0.5.5` to `0.5.5, 0.6`.~

Currently, ChainRulesTestUtils prevents CI from actually running FiniteDifferences v0.12. ~If this is merged, compatibility with FiniteDifferences v0.12 should be tested in the ChainRules integration test in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/102. To allow this integration test to run, this PR also changes the compat entry of ChainRulesTestUtils, which is currently beyond v0.5.~


Project.toml
```diff
@@ -13,9 +13,9 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.9.21"
-ChainRulesTestUtils = "0.5.5"
```
oxinabox, line 16: we probably don't want to do this, as it blocks us being able to tag a release.
Letting the integration test fail is fine

wesselb: Okay, I've reverted the change. My only worry is that FiniteDifferences v0.12 is now allowed but in no CI tested for compatibility with ChainRules.

oxinabox: its true, so might just not do this til we do ChainRulesTestUtils too.
I suspect the PRs are open and waiting for review

wesselb: Let's wait with merging this then until ChainRules runs on some 0.6 of ChainRulesTestUtils.

oxinabox:
addressed already elsewhere

PR #342, opened 2021-01-12
Update to new version of ChainRulesTestUtils and FiniteDifferences

oxinabox:
Replaces #341 
This PR includes some changes i made when checking that everything was inferrable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docs/build
 docs/site
 .idea/*
 dev/*
+Project.toml
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ docs/build
 docs/site
 .idea/*
 dev/*
-Project.toml
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.9.25"
-ChainRulesTestUtils = "0.5, 0.6"
+ChainRulesTestUtils = "0.5, 0.6.1"
 Compat = "3"
 FiniteDifferences = "0.11, 0.12"
 Reexport = "0.2, 1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.45"
+version = "0.7.46"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -12,11 +12,11 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.9.21"
-ChainRulesTestUtils = "0.5.5"
+ChainRulesCore = "0.9.25"
+ChainRulesTestUtils = "0.5, 0.6"
 Compat = "3"
-FiniteDifferences = "0.11.4"
-Reexport = "0.2, 1.0"
+FiniteDifferences = "0.11, 0.12"
+Reexport = "0.2, 1"
 Requires = "0.5.2, 1"
 julia = "1"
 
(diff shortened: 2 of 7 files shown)


test/rulesets/Base/fastmath_able.jl
```diff
@@ -60,7 +60,7 @@ const FASTABLE_AST = quote
         @testset "Multivariate" begin
             @testset "sincos(x::$T)" for T in (Float64, ComplexF64)
                 x, Δx, x̄ = randn(T, 3)
-                Δz = (randn(T), randn(T))
+                Δz = Composite{Tuple{T,T}}(randn(T), randn(T))
```
oxinabox, line 63: This worked without this, because we ducktype on any iterable.
But `Composite{Tuple{T, T}}` is the right type

.gitignore
```diff
@@ -7,3 +7,4 @@ docs/build
 docs/site
 .idea/*
 dev/*
+Project.toml
```
mzgubic, line 10: that is not intended right?

oxinabox: Arg no lol that seems really bad
```suggestion
```

test/rulesets/Base/fastmath_able.jl
mzgubic: Just making sure I understand, [this is the line](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/master/src/rule_definition_tools.jl#L180) that ducktypes right?

mzgubic: How did you find this actually?

mzgubic (dismissed):

test/rulesets/Base/fastmath_able.jl
oxinabox: This is the line that ducktypes
https://github.com/JuliaDiff/ChainRulesCore.jl/blob/95b9fd72824945374bd1767a0fcc380265a7a038/src/rule_definition_tools.jl#L185
Because it basically generates `pullback((arg1, arg2))` rather than `pullback(x::Composite)` or `pullback((arg1, arg2)::Composite)`
It is just relying on destructuing, which just requires iteration.

> How did you find this actually?

When I was debugging the issue in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/276
I just noticed it in passing

oxinabox:
Urg what.
It all passes on 1.5 for me locally.

I can understand the CI for 1.0 failing, but not 1 (which should be 1.5).

For 1.0 we don't really care about inferability. If you are not using current Stable, you probably don't care about performance *that* much.
I wonder if we should introduce a global setting in ChainRulesTestUtils to let us disable inference checks.

oxinabox:
Ah halfway through me checking if everything inferred i mistakenly swapped to ChainRulesTestUtils v0.5.
Looks like there is still some work to do.
I will continue this tomorrow.

sethaxen:
>I can understand the CI for 1.0 failing, but not 1 (which should be 1.5).

I think these failures on 1 were all expected. BLAS failures were documented in #328. eigendecomposition is type-unstable, so we just need to set `check_inferred=false` there..

> I wonder if we should introduce a global setting in ChainRulesTestUtils to let us disable inference checks.
> Then we could conditionally set it in ChainRules.jl's tests depending on the VERSION.


That seems fine. I can open a PR.

oxinabox:
> 
> 
>     I wonder if we should introduce a global setting in ChainRulesTestUtils to let us disable inference checks.
>     Then we could conditionally set it in ChainRules.jl's tests depending on the VERSION.
> 
> That seems fine. I can open a PR.

Let's see if there are any that are left once I get 1.5 passing.

Project.toml
```diff
@@ -12,11 +12,11 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.9.21"
-ChainRulesTestUtils = "0.5.5"
+ChainRulesCore = "0.9.25"
+ChainRulesTestUtils = "0.5, 0.6"
```
sethaxen, line 16: Currently Julia v1.5 passes but only with v0.6.1
```suggestion
ChainRulesTestUtils = "0.5, 0.6.1"
```

sethaxen:
Looks like it's just a few tweaks we need to pass on Julia 1.0 now.

oxinabox:
nice, I will attend to them tomorrow

sethaxen (approved):
LGTM!

PR #346, opened 2021-01-13
Add Emoji to GHA names

oxinabox:
Trying this out to see if it makes it easier to distinguish the different jobs

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test:
-    name: ${{ matrix.version.icon }}Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    name: ${{ matrix.version.icon }} Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -68,7 +68,7 @@ jobs:
           parallel-finished: true
 
   slack:
-    name: 🔊Notify Slack Failure
+    name: 🔊 Notify Slack Failure
     needs: test
     runs-on: ubuntu-latest
     if: always() && github.event_name == 'schedule'
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,14 +9,14 @@ on:
 
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    name: ${{ matrix.version.icon }}Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1'
+          - {number: '1.0', icon: '🧓🏼'}
+          - {number: '1', icon: '🟣'}
         os:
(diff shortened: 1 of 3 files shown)


sethaxen:
TBH I don't particularly like the choice of emojis (is that an old woman? Is that a night-time cityscape?). We want some visual way to tell them apart, but isn't what we really want some way to tell the things that must pass apart from the things that don't need to? Perhaps then it would be better to put some green emoji in the name of the CI workflows and other colors next to other workflows? Then for a quick visual check you look for red next to green, and that's a bad thing.

oxinabox:
> Then for a quick visual check you look for red next to green, and that's a bad thing.

That's going to mess me up, I am going to spot any green and think that that line passes.

Is that an old woman?  The emoji name is old person, I am not sure if person always defaults to man or woman, or androgenous.
I wanted to indicate it is the oldest version of julia
I will try and find another one for that

Is that a night-time cityscape? For nightly.
I will try a sliver of a moon?

.github/workflows/CI.yml
```diff
@@ -9,14 +9,14 @@ on:
 
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    name: ${{ matrix.version.icon }}Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
```
oxinabox, line 12:
```suggestion
    name: ${{ matrix.version.icon }} Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
```

.github/workflows/CI.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           parallel-finished: true
 
   slack:
-    name: Notify Slack Failure
+    name: 🔊Notify Slack Failure
```
oxinabox, line 71:
```suggestion
    name: 🔊 Notify Slack Failure
```

.github/workflows/IntegrationTestZygote.yml
```diff
@@ -1,4 +1,4 @@
-name: IntegrationTestZygote 
+name: 🧩IntegrationTestZygote
```
oxinabox, line 1:
```suggestion
name: 🧩 IntegrationTestZygote 
```

.github/workflows/CI.yml
```diff
@@ -9,14 +9,14 @@ on:
...
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    name: ${{ matrix.version.icon }} Julia ${{ matrix.version.number }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1'
+          - {number: '1.0', icon: '🧓🏼'}
```
oxinabox, line 18:
```suggestion
          - {number: '1.0', icon: '🏛️'}
```

nickrobinson251:
i agree with Seth in that i find this more confusing.

can we just change to e.g. a star for things that are important (must pass), and no emoji for anything else?

oxinabox:
Problem is IntegrationTestZygote must pass, unless Zygote is broken or not compatible.

and in my last PR (#342) I really wanted to quickly see if it was Julia 1.0 failing or Julia 1.5.

I wonder if the answer is we should stop testing so many varients and operating systems.
because having all those different ones made identifying which was failing harder

PR #348, opened 2021-01-15
Stop doing CI on Windows and Mac

oxinabox:
An alternative/suppliment to #346 
The reason to add Emoji is to help navigate the over a dozen different varieties of things being tested.
What if we just tested less things?
Plus we have a lot of tests from our reverse dependency tests etc.

This code should be entirely OS independent.
We have never had a failure that occurs only one OS.
because we don't touch anything like binaries or IO where things could be weird.
It's just wasting time and CPU cycles testing on these extra OSes.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,17 +17,8 @@ jobs:
         version:
           - '1.0'
           - '1'
-        os:
-          - ubuntu-latest
-          - macOS-latest
-          - windows-latest
-        arch:
-          - x86
-          - x64
-        # 32-bit Julia binaries are not available on macOS
-        exclude:
-          - os: macOS-latest
-            arch: x86
+        os: [ubuntu-latest]
+        arch: [x86, x64]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -17,17 +17,8 @@ jobs:
...
-          - windows-latest
-        arch:
-          - x86
-          - x64
-        # 32-bit Julia binaries are not available on macOS
-        exclude:
-          - os: macOS-latest
-            arch: x86
+        os: [ubuntu-latest]
+        arch: [x86, x64]
```
oxinabox, line 21: Tempting to drop one of these.
Maybe from x64 since if it works on x86 it will work on x64.
and its likely that people have tested it on x64 locally.

mattBrzezinski (approved):

mattBrzezinski:
Changes LGTM, dropping `x64` also can make sense if you want to reduce the run time here.
Looks like it needs a rebase and conflicts resolved.

PR #349, opened 2021-01-15
use new Revere Dependency Test code and test ReversePropagation.jl

oxinabox:
cc @dpsanders  we are now doing reverse dependency tests on ReversePropagation.jl so we can be extra sure not to break it


oxinabox:
It passes all 4 tests?
```
 25l25hTest Summary:         | Pass  Total
ReversePropagation.jl |    4      4
    Testing ReversePropagation tests passed 
    ```

mattBrzezinski (approved):

PR #350, opened 2021-01-16
Install VersionVigilante

willtebbutt:
Installs [this](https://github.com/bcbi/VersionVigilante.jl) wonderful GHA that @oxinabox pointed me towards.

Downsides:
1. It insists on a version bump for _every_ PR, regardless which bit of the code the PR touches.

Upsides:
1. It should substantially reduce the chance of forgetting to bump the patch -- it's annoying when this happens.
2. Will make it clearer to new contributors that we expect the version number to change when they contribute, over and above what is already discussed in ColPrac.

On balance I think this is a win, hence the PR. If we achieve consensus around this, I'll open PRs to the other ChainRules-related packages.

### DIFF
--- a/.github/workflows/VersionVigilante_pull_request.yml
+++ b/.github/workflows/VersionVigilante_pull_request.yml
@@ -1,0 +1,31 @@
+name: VersionVigilante
+
+on: pull_request
+
+jobs:
+  VersionVigilante:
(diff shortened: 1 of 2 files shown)


oxinabox:
Let's try it out

oxinabox (approved):

PR #351, opened 2021-01-18
Add rules for dense matrix exponential

### LINKED ISSUE #331: Add rules for the matrix exponential
TL/DR a better approach for rules for matrix exponential of dense matrices.

We should add rules for `exp(::StridedMatrix)` would supersede Zygote's.

The *right* rule for `exp` would be none, i.e. to AD through `exp`, which uses the scaling and squaring algorithm, but this uses mutation, which Zygote doesn't support. it also is limited to `BlasFloat`, so ForwardDiff and I don't think any of the operator overloading AD's can handle it. All that to say, we should have the rules.

Zygote currently uses the power series pullback using eigendecomposition for the backward pass. The eigendecomposition is not an accurate way to compute the exponential in general (https://epubs.siam.org/doi/10.1137/S00361445024180) (it's fine for hermitian matrices though, hence the `exp(::Hermitian)` overload in LinearAlgebra). Zygote's adjoint uses `exp` for the primal, so it only introduces potential inaccuracy in the pullback though. However, it doesn't follow the same time complexity as the primal function, and it is quite wasteful.

EDIT: everything said below is still valid, but it is a general property of power series matrix functions with real coefficients that the pullback is the pushforward pre- and post- composed with `adjoint`, or, equivalently, the pushforward of the function applied to the adjoint of the primal. i.e. if `Y = f(A)`, then we have the equality
```julia
(f^*)_{Y} (ΔY) = (f_*)_{A'} (ΔY)
```
This applies to `exp`, `log`, and all trigonometric and hyperbolic functions.

~`Y = exp(A)` appears in the solution to an ODE. We can augment the ODE to get a new one whose solution also uses the matrix exponential and which gives us the pushforward (discussed in section 7 of https://www2.humusoft.cz/www/papers/tcp08/017_brancik.pdf, though the result is older and can be worked out from https://ieeexplore.ieee.org/document/1101743 or witty algebra).
In short, given `B = [A ΔA; zero(A) A]`, then `exp(B) = [Y ∂Y; zero(Y) Y]`~

~I didn't find a reference for this, but we can do the same thing for the pullback by constructing and solving the adjoint ODE
Given `B = [A ΔY'; zero(A) A]`, then `exp(B) = [Y ∂A'; zero(Y) Y]`
That is, the pullback `exp^*` is related to the pushforward `exp_{*}` by `exp^* = adjoint ∘ exp_{*} ∘ adjoint`.~ This is easy to verify:

```julia
julia> using LinearAlgebra, FiniteDifferences

julia> A, Δ = randn(ComplexF64, 30, 30), randn(ComplexF64, 30, 30);

julia> only(j′vp(central_fdm(5, 1), exp, Δ, A)) ≈ jvp(central_fdm(5, 1), exp, (A, Δ'))'
true
```

~The problem with the augmented matrix approach is that it is 8x the cost of the primal, when we should be able to get <5x. For small matrices (~<100x100) this is faster than the eigendecomposition approach, and it should be more accurate, but for large dense matrices, the eigendecomposition approach is faster.~

But the relationship between the pushforward and pullback motivates a solution. Namely, explicitly implement the pushforward of the scaling and squaring approach used by `LinearAlgebra.exp!`. Not only do we get the pushforward with the same time complexity as the primal, but we can then compute the pullback with the same time complexity of the primal without the need to checkpoint any of the intermediate matrices, and with mutation allowed.

I'm planning to tackle this after some of my other open PRs are wrapped up, but I wanted to get it in writing while it was fresh on my mind.

sethaxen:
Fixes #331

Because I plan to add a number of matrix functions, and they are long, I have added these rules to a new function `matfun.jl`.
These rules unfortunately require quite a bit of code duplication from LinearAlgebra, but following discussion in https://github.com/JuliaLang/julia/issues/5840, that's only unavoidable if we refactor these functions in LinearAlgebra to return their intermediates.


src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
+# matrix functions of dense matrices
+
+# NOTE: for matrix functions whose power series representation has real coefficients,
+# the pullback and pushforward are related by an adjoint.
+# Specifically, if the pushforward of f(A) is (f_*)_A(ΔA), then the pullback at Y=f(A) is
+# (f^*)_Y(ΔY) = (f_*)_{A'}(ΔY) = ((f_*)_A(ΔY'))'
```
sethaxen, line 6: Well that is hideous, but notation is hard, and harder in unicode.

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+
+Compute the Fréchet derivative of the matrix function `Y=f(A)`, where the Fréchet derivative
+of `A` is `ΔA`, and `intermediates` is the second argument returned by `_matfun`.
+"""
+_matfun_frechet
+
+"""
+    _matfun_frechet!(f, A, Y, ΔA, intermediates)
+
+Similar to `_matfun_frechet!`, but where `ΔA` may be overwritten.
```
sethaxen, line 37:
```suggestion
Similar to [`_matfun_frechet`](@ref), but where `ΔA` may be overwritten.
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+end
+
+## Destructive matrix exponential using algorithm from Higham, 2008,
+## "Functions of Matrices: Theory and Computation", SIAM
+## Adapted from LinearAlgebra.exp! with return of intermediates
+function _matfun!(::typeof(exp), A::StridedMatrix{T}) where T<:BlasFloat
+    n = LinearAlgebra.checksquare(A)
+    ilo, ihi, scale = LAPACK.gebal!('B', A)    # modifies A
+    nA   = opnorm(A, 1)
+    Inn    = Matrix{T}(I, n, n)
```
oxinabox, line 93:
```suggestion
    ilo, ihi, scale = LAPACK.gebal!('B', A)  # modifies A
    nA = opnorm(A, 1)
    Inn = Matrix{T}(I, n, n)
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        else
+            C = T[120.,60.,12.,1.]
+        end
+        si = 0
+    else
+        C = T[64764752532480000.,32382376266240000.,7771770303897600.,
+                1187353796428800.,  129060195264000.,  10559470521600.,
+                    670442572800.,      33522128640.,      1323241920.,
+                        40840800.,           960960.,           16380.,
+                             182.,                1.]
```
oxinabox, line 115: In generalk I am ok with this formatting, except should space after comma.
It did make me think this was a matrix rather than a vector initially.
But it's fine

```suggestion
            C = T[17643225600., 8821612800., 2075673600., 302702400.,
                     30270240.,    2162160.,     110880.,      3960.,
                           90.,          1.]
        elseif nA > 0.25
            C = T[17297280., 8648640., 1995840., 277200.,
                     25200.,    1512.,      56.,      1.]
        elseif nA > 0.015
            C = T[30240., 15120., 3360.,
                    420.,    30.,    1.]
        else
            C = T[120., 60., 12., 1.]
        end
        si = 0
    else
        C = T[64764752532480000., 32382376266240000., 7771770303897600.,
                1187353796428800.,   129060195264000.,   10559470521600.,
                    670442572800.,       33522128640.,       1323241920.,
                        40840800.,            960960.,            16380.,
                             182.,                 1.]
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+    if si > 0
+        A ./= convert(T,2^si)
+    end
+
+    A2 = A * A
+    P  = copy(Inn)
+    W  = C[2] * P
+    V  = C[1] * P
+    Apows = typeof(P)[]
+    for k in 1:(div(size(C, 1), 2) - 1)
```
oxinabox, line 129:
```suggestion
    for k in 1:(size(C, 1) ÷ 2 - 1)
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+                1187353796428800.,  129060195264000.,  10559470521600.,
+                    670442572800.,      33522128640.,      1323241920.,
+                        40840800.,           960960.,           16380.,
+                             182.,                1.]
+        s = log2(nA/5.4)               # power of 2 later reversed by squaring
+        si = ceil(Int,s)
+    end
+
+    if si > 0
+        A ./= convert(T,2^si)
```
oxinabox, line 121:
```suggestion
        A ./= convert(T, 2^si)
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        end
+        si = 0
+    else
+        C = T[64764752532480000.,32382376266240000.,7771770303897600.,
+                1187353796428800.,  129060195264000.,  10559470521600.,
+                    670442572800.,      33522128640.,      1323241920.,
+                        40840800.,           960960.,           16380.,
+                             182.,                1.]
+        s = log2(nA/5.4)               # power of 2 later reversed by squaring
+        si = ceil(Int,s)
```
oxinabox, line 117:
```suggestion
        si = ceil(Int, s)
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+    for k in 1:(div(size(C, 1), 2) - 1)
+        k2 = 2 * k
+        P *= A2
+        push!(Apows, P)
+        W += C[k2 + 2] * P
+        V += C[k2 + 1] * P
+    end
+    U = A * W
+    X = V + U
+    F = lu!(V-U) # NOTE: use lu! instead of LAPACK.gesv! so we can reuse factorization
```
oxinabox, line 138:
```suggestion
    F = lu!(V - U)  # NOTE: we use lu! instead of LAPACK.gesv! so we can reuse factorization
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+    ldiv!(F, X)
+    Xpows = typeof(X)[X]
+    if si > 0            # squaring to reverse dividing by power of 2
+        for t=1:si
```
oxinabox, line 142:
```suggestion
        for t in 1:si
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        push!(Apows, P)
+        W += C[k2 + 2] * P
+        V += C[k2 + 1] * P
+    end
+    U = A * W
+    X = V + U
+    F = lu!(V-U) # NOTE: use lu! instead of LAPACK.gesv! so we can reuse factorization
+    ldiv!(F, X)
+    Xpows = typeof(X)[X]
+    if si > 0            # squaring to reverse dividing by power of 2
```
oxinabox, line 141:
```suggestion
    if si > 0  # squaring to reverse dividing by power of 2
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        for t=1:si
+            X *= X
+            push!(Xpows, X)
+        end
+    end
+
+    # Undo the balancing
+    for j = ilo:ihi
```
oxinabox, line 149:
```suggestion
    for j in ilo:ihi
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        scj = scale[j]
+        for i = 1:n
```
oxinabox, line 151:
```suggestion
        for i in 1:n
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+            X[j,i] *= scj
```
oxinabox, line 152:
```suggestion
            X[j, i] *= scj
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+        end
+        for i = 1:n
```
oxinabox, line 154:
```suggestion
        for i in 1:n
```

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,243 @@
...
+            X[i,j] /= scj
```
oxinabox, line 155:
```suggestion
            X[i, j] /= scj
```

oxinabox (commented):
I started to comment on the formatting, but its a bit much.
I think just run https://github.com/domluna/JuliaFormatter.jl/ over it
`format_file("matfun.jl", BlueStyle())`
its pretty good.

I will review after that, since then I won't be spending time on the basic stuff.

src/rulesets/LinearAlgebra/matfun.jl
nickrobinson251: fwiw, JuliaFormatter BlueStyle tells me this "should" be
```suggestion
            C = T[
                17643225600.0,
                8821612800.0,
                2075673600.0,
                302702400.0,
                30270240.0,
                2162160.0,
                110880.0,
                3960.0,
                90.0,
                1.0,
            ]
        elseif nA > 0.25
            C = T[17297280.0, 8648640.0, 1995840.0, 277200.0, 25200.0, 1512.0, 56.0, 1.0]
        elseif nA > 0.015
            C = T[30240.0, 15120.0, 3360.0, 420.0, 30.0, 1.0]
        else
            C = T[120.0, 60.0, 12.0, 1.0]
        end
        si = 0
    else
        C = T[
            64764752532480000.0,
            32382376266240000.0,
            7771770303897600.0,
            1187353796428800.0,
            129060195264000.0,
            10559470521600.0,
            670442572800.0,
            33522128640.0,
            1323241920.0,
            40840800.0,
            960960.0,
            16380.0,
            182.0,
            1.0,
        ]
```

nickrobinson251: i think running JuliaFormatter on this file would be good :)

sethaxen:
> I started to comment on the formatting, but its a bit much.
Yeah, it's the formatting used by `exp!` in LinearAlgebra, which is not great.

> I think just run https://github.com/domluna/JuliaFormatter.jl/ over it
> `format_file("matfun.jl", BlueStyle())`
Done!

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,261 @@
...
+        scj = scale[j]
+        for i in 1:n
+            ∂X[j, i] *= scj
+        end
+        for i in 1:n
+            ∂X[i, j] /= scj
+        end
+    end
+
+    if ilo > 1  # apply lower permutations in reverse order
```
sethaxen, line 250: I haven't been able to find a matrix `A` that hits this branch. As far as I can tell, all `exp` tests in LinearAlgebra also do not hit the corresponding branch, but code coverage indicates it is covered, so it must be indirectly hit by one of the trigonometric/hyperbolic matrix function tests or something.

sethaxen: Nevermind, found one

oxinabox (approved):

src/rulesets/LinearAlgebra/matfun.jl
sethaxen: >What if we just made a section for this in the docs, (maybe as internal notes or something)
> and wrote the latex and then linked to that?

Hm, that's an idea. I'll consider it, potentially for a future PR.

> This bit makes sense:
> `(f^*)_Y(ΔY) = ((f_*)_A(ΔY'))'`
> so the pullback at `A`, i.e. the pullback from Y...is equal to the the adjoint of the pushing forward at `A`, the adjoint of of the output senstivity.

Ah yes your description is correct (although it's the adjoint of the pushing forward of the adjoint). I just checked Lee, and this should be the right notation:
```suggestion
# (f^*)_A(ΔY) = (f_*)_{A'}(ΔY) = ((f_*)_A(ΔY'))'
```

> (though that's not well defined since not all functions are monotonic?)

I'm not sure what you mean by this.

> Magical expodential symmetry? (I feel like i made the same suprised sounds for the same reason on your last PR)

It's still surprising to me. Although this property is general for all of the matrix functions defined in LinearAlgebra, not just `exp`. It doesn't follow for all matrix functions though, just those whose convergent power series have *real* coefficients.

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -0,0 +1,270 @@
...
+            ∂hermA = _matfun_frechet(exp, hermA, hermX, ΔX, hermX_intermediates)
+            ∂hermA isa LinearAlgebra.RealHermSymComplexHerm || return NO_FIELDS, ∂hermA
+            return NO_FIELDS, parent(∂hermA)
+        end
+        return LinearAlgebra.copytri!(parent(hermX), 'U', true), exp_pullback_hermitian
+    else
+        A = copy(A0)
+        X, intermediates = _matfun!(exp, A)
+        function exp_pullback(ΔX)
+            ΔX′ = copy(adjoint(ΔX))
```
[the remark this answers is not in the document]
sethaxen: Perhaps. It might make sense to explicitly introduce a `_matfun_frechet_adjoint` and in-place version that handle the adjoint logic, which will be a little different for every matrix type.

sethaxen:
@oxinabox I added a `_matfun_frechet_adjoint`, changed the signature to put the differential first, and substantially modified the docstrings and comments. Would you mind re-reviewing the comments and docstrings at the top of `matfun.jl` before I merge?

oxinabox (approved):
LGTM

PR #352, opened 2021-01-20
Fix CI cache bug

fchorney:


.github/workflows/CI.yml
```diff
@@ -15,17 +15,17 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1'
+          - '1.0'  # LTS
+          - '1'    # Latest Release
```
mattBrzezinski, line 19: Since we've started using 2x quotes everywhere else

```suggestion
          - "1.0"  # LTS
          - "1"    # Latest Release
```

mattBrzezinski (approved):

PR #354, opened 2021-02-04
Rules for LU decomposition of StridedMatrixes

sethaxen:
This PR adds rules for the LU decomposition of `StridedMatrix`es, as well as its use for computing the matrix inverse.
- [x] `frule` for `lu!(::StridedMatrix)`
- [x] `rrule` for `lu(::StridedMatrix)`
- [x] `rrule` for `getproperty(::LU, k)`
- [x] `frule` for `inv!(::LU)`
- [x] `rrule` for `inv(::LU)`

All rules are restricted to `StridedMatrix` inputs because these rules do not make efficient use of storage for inputs of type `SparseMatrixCSC` or `Tridiagonal`, each of which has their own specialized `lu` method.

EDIT: for a derivation of the rules, see https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -28,11 +28,9 @@ function frule(
         U = UpperTriangular(F.factors)
         rdiv!(∂factors, U)
         ldiv!(L, ∂factors)
-        ∂L = tril(∂factors, -1)
-        rmul!(UpperTriangular(∂factors), U)
-        lmul!(L, ∂L)
-        ∂factors .+= ∂L
-        ∂U = triu(∂factors)
+        ∂L = lmul!(L, tril(∂factors, -1))
+        ∂U = rmul!(triu(∂factors), U)
+        ∂factors .= ∂L .+ ∂U
     elseif m < n  # wide A, system is [P*A1 P*A2] = [L*U1 L*U2]
         L = UnitLowerTriangular(F.L)
         U = F.U
@@ -45,9 +43,9 @@ function frule(
         ldiv!(L, ∂factors)
         rdiv!(∂factors1, U1)
         ∂L = tril(∂factors1, -1)
-        rmul!(UpperTriangular(∂factors1), U1)
         mul!(∂factors2, ∂L, U2, -1, 1)
         lmul!(L, ∂L)
+        rmul!(triu!(∂factors1), U1)
         ∂factors1 .+= ∂L
         ∂U = triu(∂factors)
     else  # tall A, system is [P1*A; P2*A] = [L1*U; L2*U]
@@ -61,13 +59,12 @@ function frule(
         end
         rdiv!(∂factors, U)
         ldiv!(L1, ∂factors1)
-        mul!(∂factors2, L2, UpperTriangular(∂factors1), -1, 1)
-        ∂L = tril(∂factors, -1)
-        ∂L1 = @views LowerTriangular(∂L[1:q, :])
-        rmul!(UpperTriangular(∂factors1), U)
-        lmul!(L1, ∂L1)
-        ∂factors1 .+= ∂L1
         ∂U = triu(∂factors1)
+        mul!(∂factors2, L2, ∂U, -1, 1)
+        rmul!(∂U, U)
+        lmul!(L1, tril!(∂factors1, -1))
+        ∂factors1 .+= ∂U
+        ∂L = tril(∂factors, -1)
     end
     ∂F = Composite{typeof(F)}(; L=∂L, U=∂U, factors=∂factors)
     return F, ∂F
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -182,8 +182,8 @@ function frule((_, ΔF), ::typeof(LinearAlgebra.inv!), F::LU{<:Any,<:StridedMatr
     L = UnitLowerTriangular(F.factors)
     U = UpperTriangular(F.factors)
     # compute ∂Y = -(U \ (L \ ∂L + ∂U / U) / L) * P while minimizing allocations
-    ∂Y = ldiv!(L, ΔF.L)
-    ∂Y .+= rdiv!(UpperTriangular(ΔF.U), U)
+    ∂Y = ldiv!(L, tril!(ΔF.L, -1))
+    ∂Y .+= rdiv!(triu!(ΔF.U), U)
     ldiv!(U, ∂Y)
     rdiv!(∂Y, L)
     rmul!(∂Y, -1)
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -60,6 +60,7 @@ function frule(
             L1 = UnitLowerTriangular(L[1:q, :])
             L2 = L[(q + 1):end, :]
         end
+        # Here we manipulate ∂factors both directly and via views: ∂factors1 and  ∂factors2
         rdiv!(∂factors, U)
         ldiv!(L1, ∂factors1)
         ∂U = triu(∂factors1)
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -51,7 +51,6 @@ function frule(
         lmul!(L, ∂L)
         rmul!(triu!(∂factors1), U1)
         ∂factors1 .+= ∂L
-        ∂U = triu(∂factors)
     else  # tall A, system is [P1*A; P2*A] = [L1*U; L2*U]
         L = F.L
         U = UpperTriangular(F.U)
@@ -68,9 +67,8 @@ function frule(
         rmul!(∂U, U)
         lmul!(L1, tril!(∂factors1, -1))
         ∂factors1 .+= ∂U
-        ∂L = tril(∂factors, -1)
     end
-    ∂F = Composite{typeof(F)}(; L=∂L, U=∂U, factors=∂factors)
+    ∂F = Composite{typeof(F)}(; factors=∂factors)
     return F, ∂F
 end
 
@@ -79,16 +77,10 @@ function rrule(
 )
     F = lu(A, pivot; kwargs...)
     function lu_pullback(ΔF::Composite)
-        ΔL = ΔF.L
-        ΔU = ΔF.U
-        if ΔL isa AbstractZero && ΔU isa AbstractZero
-            return (NO_FIELDS, ΔL + ΔU, DoesNotExist())
-        end
+        Δfactors = ΔF.factors
+        Δfactors isa AbstractZero && return (NO_FIELDS, Δfactors, DoesNotExist())
         factors = F.factors
-        if eltype(A) <: Real
-            ∂L = real(ΔL)
-            ∂U = real(ΔU)
-        end
+        ∂factors = eltype(A) <: Real ? real(Δfactors) : Δfactors
         ∂A = similar(factors)
         m, n = size(A)
         q = min(m, n)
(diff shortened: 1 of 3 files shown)


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,218 @@
...
+
+# These rules are necessary because the primals call LAPACK functions
+
+# frule for square matrix was introduced in Eq. 3.6 of
+# de Hoog, F.R., Anderssen, R.S. and Lukas, M.A. (2011)
+# Differentiation of matrix functionals using triangular factorization.
+# Mathematics of Computation, 80 (275). p. 1585.
+# doi: http://doi.org/10.1090/S0025-5718-2011-02451-8
+# for derivations for wide and tall matrices, see
+# https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/
```
nickrobinson251, line 16: This is great!

Can we add it to the `More Examples` section of the `Deriving Array Rules` docs page?

E.g. some like
```md
For Differentiating the LU decomposition, see [this blog post by Seth Axen](https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/).
```

sethaxen: This is a good idea. I currently have an open PR to update that docs page, and I'll include this there: https://github.com/JuliaDiff/ChainRulesCore.jl/pull/290

oxinabox:
I had hoped to have time to review this today, but I have not.
I will make sure to do so tomorrow

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,218 @@
...
+        C = Composite{TF}
+        ∂F = if x === :L
+            C(; L=tril(ΔY, -1))
+        elseif x === :U
+            C(; U=triu(ΔY))
+        elseif x === :factors
+            m, n = size(F.factors)
+            q = min(m, n)
+            ∂L = tril(n == q ? ΔY : view(ΔY, :, 1:q), -1)
+            ∂U = triu(m == q ? ΔY : view(ΔY, 1:q, :))
```
oxinabox, line 171: Slightly more constant folding friendly
```suggestion
            ∂L = tril(n === q ? ΔY : view(ΔY, :, 1:q), -1)
            ∂U = triu(m === q ? ΔY : view(ΔY, 1:q, :))
```

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -12,6 +12,133 @@ function FiniteDifferences.to_vec(x::Val)
 end
 
 @testset "Factorizations" begin
+    @testset "lu decomposition" begin
+        # avoid implementing to_vec(::LU)
```
oxinabox, line 16: why are we avoiding this?

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -12,6 +12,133 @@ function FiniteDifferences.to_vec(x::Val)
...
+                F = lu!(copy(A), pivot)
+                ∂F_fd = jvp(_fdm, A -> asnt(lu!(copy(A), pivot)), (A, ΔA))
+                F_ad, ∂F_ad = @inferred frule(
+                    (Zero(), copy(ΔA), DoesNotExist()), lu!, copy(A), pivot
+                )
+                @test F_ad == F
+                @test ∂F_ad isa Composite{typeof(F)}
+                @test ∂F_ad.L ≈ ∂F_fd.L
+                @test ∂F_ad.U ≈ ∂F_fd.U
+                @test ∂F_ad.factors ≈ ∂F_fd.factors
```
oxinabox, line 37: what is preventing us using ChainRulesTestUtils here?
Can we open an issue and cross-reference it in the code?
(or cross reverence existing one)

I am trying to workout where the limitations in ChainRulesTestUtils are so we can fix them

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -12,6 +12,133 @@ function FiniteDifferences.to_vec(x::Val)
...
+                A = randn(T, m, n)
+                ΔA = rand_tangent(A)
+                F = lu!(copy(A), pivot)
+                ∂F_fd = jvp(_fdm, A -> asnt(lu!(copy(A), pivot)), (A, ΔA))
+                F_ad, ∂F_ad = @inferred frule(
+                    (Zero(), copy(ΔA), DoesNotExist()), lu!, copy(A), pivot
+                )
+                @test F_ad == F
+                @test ∂F_ad isa Composite{typeof(F)}
+                @test ∂F_ad.L ≈ ∂F_fd.L
```
oxinabox, line 35: These are good to use
```suggestion
                check_equal(∂F_ad.L, ∂F_fd.L)
```
which ChainRulesTestUtils exports

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -12,6 +12,133 @@ function FiniteDifferences.to_vec(x::Val)
...
+
+                A = randn(T, m, n)
+                F = lu(A, pivot)
+                F_fd = asnt2(lu(A, pivot))
+                ΔL = rand_tangent(F.L)
+                ΔU = rand_tangent(F.U)
+                ΔF = Composite{typeof(F)}(; U=ΔU, L=ΔL)
+                @testset "Composite with nonzero fields: $ks" for ks in ((), (:L, :U), (:L,), (:U,))
+                    ΔF_ad = Composite{typeof(F)}(; (k => getproperty(ΔF, k) for k in ks)...)
+                    # work around FD not accepting Zero arguments
```
oxinabox, line 62: can we cross reference an issue here?

oxinabox (approved):
Nice.
This is an excellent PR

A bunch of small things.
Merge when you are happy.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,218 @@
...
+        ldiv!(L, ∂factors)
+        ∂L = lmul!(L, tril(∂factors, -1))
+        ∂U = rmul!(triu(∂factors), U)
+        ∂factors .= ∂L .+ ∂U
+    elseif m < n  # wide A, system is [P*A1 P*A2] = [L*U1 L*U2]
+        L = UnitLowerTriangular(F.L)
+        U = F.U
+        @views begin
+            ∂factors1 = ∂factors[:, 1:q]
+            ∂factors2 = ∂factors[:, (q + 1):n]
```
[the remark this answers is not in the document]
sethaxen: When we know the end index, is it still better to use `end`? I wasn't certain.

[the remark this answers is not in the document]
oxinabox: idk,
I think it ls clearer, and I assume it will optimize out.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,218 @@
...
+end
+
+function rrule(
+    ::typeof(lu), A::StridedMatrix, pivot::Union{Val{false},Val{true}}; kwargs...
+)
+    F = lu(A, pivot; kwargs...)
+    function lu_pullback(ΔF::Composite)
+        ∂L = ΔF.L
+        ∂U = ΔF.U
+        if ∂L isa AbstractZero && ∂U isa AbstractZero
```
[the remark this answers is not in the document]
sethaxen: I like it. But I think this would break type-stability. e.g. someone passes in `Diagonal` `∂U` and `∂L` filled with zeros. This would result in the returned cotangent of `A` being `Diagonal`, whereas if any of the diagonal elements were non-zero, a `Matrix` would be returned.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,218 @@
...
+            return (NO_FIELDS, ∂L + ∂U, DoesNotExist())
```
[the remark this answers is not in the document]
sethaxen: Promote all the things to `DoesNotExist`

test/rulesets/LinearAlgebra/factorization.jl
sethaxen: I had thought we were avoiding type-piracy, but I see now we define `to_vec` for `Cholesky` in this file. I have added `to_vec` for `LU` now.

test/rulesets/LinearAlgebra/factorization.jl
sethaxen: It was the weirdness of us using the `L` and `U` fields of `Composite{LU}` instead of `factors`, when `L` and `U` are not fields of the primal. But I have now changed this and we can use ChainRulesTestUtils.

test/rulesets/LinearAlgebra/factorization.jl
sethaxen: No need now, as this has been removed.

sethaxen:
I re-factored (🙂) to use `factor` everywhere as the sole field in `Composite{LU}`. I had thought this would be more complicated and/or allocate more, but it actually simplified the code a bit and maybe even reduced allocations. It also means we can use ChainRulesTestUtils for all tests.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -1,6 +1,207 @@
...
+        elseif x === :U
+            m, n = size(F.factors)
+            S = eltype(ΔY)
+            triu!([ΔY; zeros(S, max(0, m - n), n)])
+        elseif x === :factors
+            Matrix(ΔY)
+        else
+            return (NO_FIELDS, DoesNotExist(), DoesNotExist())
+        end
+        ∂F = Composite{TF}(; factors=∂factors)
```
nickrobinson251, line 159: why are `:L` and `:U` cases inside `:factors`, rather than e.g.
```suggestion
        ∂F = Composite{TF}(; L=∂L, U=∂U, factors=∂factors)
```
(with each `∂` thunked) ?

sethaxen: We could do that, and originally I did something similar. Here are the pros and cons of this approach that I can think of:
Pros:
- don't do any work or allocate any matrices in the `rrule` for `getproperty`
- Accumulating derivatives is just simple addition of the `∂L`, `∂U`, and `∂factors`.

Cons:
- the `rrule` for `lu` now needs to do the work and encode the logic of combining `∂L`, `∂U`, and `∂factors`.
- it does not mirror the primal (`getproperty` generates `L` and `U`
- we can't test the rules using `ChainRulesTestUtils`

The last con ended up pushing me to reimplement everything to use `factors`. It's now simpler and more modular, plus easier to test. As an aside, there'd be no point in thunking here, since the cotangent `ΔY` should only be distributed to either `∂L`, `∂U`, or `∂factors`, and both the primal and pullback will by necessity be type-unstable.

nickrobinson251: > it does not mirror the primal

I don't think that matters, does it? Except insofar as it leads to 

> we can't test the rules using ChainRulesTestUtils

Is not mirroring the primal what leads to this? Do we have an issue on why?
I'm wondering because this will apply to lots of Factorizations (LinearAlgebra has a habit of overloading getproperty to do a lot of work). Is this why SVD/Eigen/Cholesky tests don't use ChainRulesTestUtils?

Should we have some docs on when/why it might be better to mirror the primal? or is it something we could extend ChainRulesTestUtils to handle?

(Just trying to understand what's going on. Don't think any changes are needed in this PR so don't let this hold it up)

sethaxen: > > it does not mirror the primal
> 
> I don't think that matters, does it? Except insofar as it leads to
> 
> > we can't test the rules using ChainRulesTestUtils

It doesn't matter much, except the work has to be done _some_time, which if not in the `rrule` for `getproperty` would be in the `rrule` for `lu`. But all things being equal, it then seems cleaner to mirror the actual primal.

> > we can't test the rules using ChainRulesTestUtils
> 
> Is not mirroring the primal what leads to this? Do we have an issue on why?
> I'm wondering because this will apply to lots of Factorizations (LinearAlgebra has a habit of overloading getproperty to do a lot of work). Is this why SVD/Eigen/Cholesky tests don't use ChainRulesTestUtils?
> 
> Should we have some docs on when/why it might be better to mirror the primal? or is it something we could extend ChainRulesTestUtils to handle?

I haven't thought about this enough to be very formal about it. `eigen(A)` for non-hermitian `StridedMatrix` `A` is blocked by instability of the phase convention for `eigen`, so it's a separate issue. For Hermitian `svd` and `eigen`, the problem is that FiniteDifferences and ChainRules only agree on `Hermitian` sensitvities when the function of interest is composed with a constructor, so it's also a separate issue. I'd need to look more closely at the `cholesky` and `svd` cases for `StridedMatrix`.

Issue #356, opened 2021-02-10
at-infererred failures in LinAlg/symmetric tests on Julia nightly


simeonschaub:
Not sure whether this is because some actual implementation in Base changed to make this type unstable, or if type inference actually got worse here (random guess in that case would be https://github.com/JuliaLang/julia/pull/36208).

```julia
Testing rulesets/LinearAlgebra/symmetric.jl:	back(::Matrix{T} where T): Error During Test at /home/simeon/Documents/Julia/JuliaDiff/ChainRules.jl/test/rulesets/LinearAlgebra/symmetric.jl:26
  Got exception outside of a @test
  return type Tuple{Zero, UpperTriangular{Float64, Matrix{Float64}}, DoesNotExist} does not match inferred return type Tuple{Zero, Union{LowerTriangular{Float64, Matrix{Float64}}, UpperTriangular{Float64, Matrix{Float64}}}, DoesNotExist}
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:33
    [2] _test_inferred(::Function, ::Matrix{Float64}, ::Vararg{Any}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ ChainRulesTestUtils ~/.julia/packages/ChainRulesTestUtils/2QRer/src/testers.jl:228
    [3] _test_inferred(::Function, ::Matrix{Float64}, ::Vararg{Any})
      @ ChainRulesTestUtils ~/.julia/packages/ChainRulesTestUtils/2QRer/src/testers.jl:227
    [4] test_rrule(::Type, ::ChainRulesTestUtils.PrimalAndTangent{Matrix{Float64}, Matrix{Float64}}, ::Vararg{Any}; output_tangent::Matrix{Float64}, fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}, check_inferred::Bool, fkwargs::NamedTuple{(), Tuple{}}, rtol::Float64, atol::Float64, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ ChainRulesTestUtils ~/.julia/packages/ChainRulesTestUtils/2QRer/src/testers.jl:172
[...]

PR #357, opened 2021-02-10
small printing fix when running test suite

simeonschaub:
Otherwise, deprecation warnings or stacktraces when running the test suite will show up a bit weird.


mzgubic (approved):

simeonschaub:
I am guessing, it's not worth bumping the version here, right?

mzgubic:
Correct, we only bump when code `src` changes.

simeonschaub:
JuliaNightly failures are #356.

PR #358, opened 2021-02-10
stop redefining rules for `redirect_stdio` on 1.6

simeonschaub:
Ref https://github.com/JuliaLang/julia/pull/39132. I have also widened the signature here, since those correspond to the ones in Base. Not sure how to test this, I don't think there's a nice way to test the absence of method redefinition warnings when loading a module. Still requires https://github.com/JuliaDiff/ChainRulesCore.jl/pull/298.


oxinabox (approved):
LGTM once CI passes.
Don't forget to bump version before merging

PR #359, opened 2021-02-10
[DO NOT MERGE] test Julia nightly on every PR

simeonschaub:
Currently still blocked by #356


simeonschaub:
Ah, nvm. I didn't realize we already ran those.

Issue #360, opened 2021-02-11
Update tests to use new easier testing mechanism


oxinabox:
We need to follow up on https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/116
and take advantage of the improvements.

Here is a list of  files.
Some of there are probably really easy since they only use `test_scalar` so nothing is needed.
Others we should take this chance to clean them up more broadly since they are also doing other things like retesting the primal output, even though `test_rrule` does that now; or not even using the testing functionality at all.

- Base
    - [x] array.jl
    - [x] arraymath.jl
    - [x] base.jl
    - [x] evalpoly.jl
    - [x] fastmath_able.jl
    - [x] indexing.jl
    - [x] mapreduce.jl
- LinearAlgebra
    - [x] blas.jl
    - [x] dense.jl
    - [ ] factorization.jl
    - [x] matfun.jl
    - [x] norm.jl
    - [x] structured.jl
    - [x] symmetric.jl
 - Random
    - [x] random.jl
 - Statistics
    - [x] statistics.jl
 - packages
    - [x] NaNMath.jl
    - [x] SpecialFunctions.jl

oxinabox:
For the SpecialFunctions.jl file we can just copy the tests that I already fixed over at 
https://github.com/JuliaMath/SpecialFunctions.jl/pull/300

oxinabox:
Our rules for NaNMath.jl have no tests.  😞 
Next breaking change we should delete those rules. Noone uses that package any more AFAIK

willtebbutt:
😬

oxinabox:
random.jl doesn't use `rule_test` because it is just checking things are all returning zeros.

oxinabox:
I think that is every PR openned

oxinabox:
across all the PR s the number of lines in the test directory went from `10,629`, down to `9,197`
so a net decrease of 1432 lines.
Plus probably improved clarity of another 1432 lines by decreasing number of arguments

PR #361, opened 2021-02-11
array.jl autotangent

oxinabox:
part of #360


mzgubic (approved):
I love the new interface

willtebbutt:
This is an example of a thing working well.

oxinabox:
I really like that you can take an example of an rrule being called
like `rrule(+, 1.2, 4.0)` and just prefix it with `test_` to get `test_rrule(+, 1.2, 4.0)`
which is the test that works most of the time.

PR #365, opened 2021-02-12
SpecialFunctions.jl autotangent

oxinabox:
part of https://github.com/JuliaDiff/ChainRules.jl/issues/360

Testign this is hard,
I had to manually add SpecialFunctions@0.10
because we need a version old enough to have `lgamma` and not the have `ChainRulesCore`.
CI will not do that and so the result will be meaningless

This code changes are pretty much just what is in 
https://github.com/JuliaMath/SpecialFunctions.jl/pull/300/files
but without `beta` stuff as  we don't define that apparently.

### DIFF
--- a/test/rulesets/packages/SpecialFunctions.jl
+++ b/test/rulesets/packages/SpecialFunctions.jl
@@ -46,12 +46,8 @@ end
 
         if isdefined(SpecialFunctions, :logabsgamma)
             isreal(x) || continue
-
-            Δx, x̄ = randn(2)
-            Δz = Composite{Tuple{Float64, DoesNotExist}}(randn(), randn())
-
-            frule_test(SpecialFunctions.logabsgamma, (x, Δx))
-            rrule_test(SpecialFunctions.logabsgamma, Δz, (x, x̄))
+            test_frule(logabsgamma, x)
+            test_rrule(logabsgamma, x; output_tangent=(randn(), randn()))
         end
     end
 end


mzgubic (approved):
Sorry could you explain the comment a bit? I don't think I follow. The code LGTM.

Also, what is the plan for dropping dependence on SpecialFunctions? I guess if it is all defined in the package we can just drop it and change a compat bound for SpecialFunctions? It doesn't seem breaking if we do it that way?

oxinabox:
>  Sorry could you explain the comment a bit? I don't think I follow. The code LGTM.

The test code won't trigger on current version of SpecialFunctions since the code is all there and there is a `if isdefined(SpecialFunctions, :ChainRulesCore)` guard around the whole testfile.

CI installs the current release of SpecialFunctions, so doesn't run this file

---

> Also, what is the plan for dropping dependence on SpecialFunctions? I guess if it is all defined in the package we can just drop it and change a compat bound for SpecialFunctions? It doesn't seem breaking if we do it that way?

Dropping it is breaking, because if someone is pinned to an old-version of SpecialFunction they should also be getting an old version of ChainRulesm but once we drop it the package manager won't know to check.
 (this is a reason why Type-Piracy is bad, its hard to leave.)
 
 It is not breaking if we add a lowered boundl; that is high enough to prevent it.
 We should do that. and delete these tests.
 The reason we didn't earlier is that Invenia has an internal package that was bound to an old version because (I believed) newer versions didn't support Julia 1.0 which we were using at the time.
 But we can update that now.
 
 Then after a while we drop that lower bound, entirely and tag as a breaking change.

PR #366, opened 2021-02-12
base.jl autotangent

mzgubic:
Helps with #360

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -68,72 +68,62 @@
     @testset "Complex" begin
         test_scalar(Complex, randn())
         test_scalar(Complex, randn(ComplexF64))
-        x, ẋ, x̄ = randn(3)
-        y, ẏ, ȳ = randn(3)
-        Δz = randn(ComplexF64)
-        frule_test(Complex, (x, ẋ), (y, ẏ))
-        rrule_test(Complex, Δz, (x, x̄), (y, ȳ))
+        x, y = randn(2)
+        test_frule(Complex, x, y)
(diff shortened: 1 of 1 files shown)


test/rulesets/Base/base.jl
```diff
@@ -68,72 +68,62 @@
...
     @testset "identity" for T in (Float64, ComplexF64)
-        frule_test(identity, (randn(T), randn(T)))
-        frule_test(identity, (randn(T, 4), randn(T, 4)))
-        frule_test(
+        test_frule(identity, randn(T))
+        test_frule(identity, randn(T, 4))
+        test_frule(
             identity,
-            (Tuple(randn(T, 3)), Composite{Tuple{T, T, T}}(randn(T, 3)...))
+            Tuple(randn(T, 3)) ⊢ Composite{Tuple{T, T, T}}(randn(T, 3)...)
```
oxinabox, line 118: Does Tuple not do that on it own?

test/rulesets/Base/base.jl
```diff
@@ -143,41 +133,35 @@
...
+        test_frule(fma, x, y, z)
+        test_rrule(fma, x, y, z)
     end
 
     @testset "clamp"  begin
-        x̄, ȳ, z̄    = randn(3)
-        Δx, Δy, Δz = randn(3)
-        Δk = randn()
-
         x, y, z = 1., 2., 3.  # to left
```
oxinabox, line 154: This test set would massively benifit from not using variables and just putting the values in

I think that is true across-the-board.
But here more so

oxinabox (approved):
Do look at my comments and then merge when ready

PR #368, opened 2021-02-12
fastmath_able.jl autotangent

mzgubic:
Helps with #360

### DIFF
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -59,11 +59,10 @@ const FASTABLE_AST = quote
         end
         @testset "Multivariate" begin
             @testset "sincos(x::$T)" for T in (Float64, ComplexF64)
-                x = randn(T)
                 Δz = Composite{Tuple{T,T}}(randn(T), randn(T))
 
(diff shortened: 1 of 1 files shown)


test/rulesets/Base/fastmath_able.jl
```diff
@@ -59,11 +59,11 @@ const FASTABLE_AST = quote
         end
         @testset "Multivariate" begin
             @testset "sincos(x::$T)" for T in (Float64, ComplexF64)
-                x, Δx, x̄ = randn(T, 3)
+                x = randn(T)
```
oxinabox, line 62: is there any reason to keep this as a seperate variable and not just put `randn()` below?

test/rulesets/Base/fastmath_able.jl
```diff
@@ -171,8 +169,8 @@ const FASTABLE_AST = quote
             y, Δy, ȳ = rand(T, 3) .+ 3
             Δz = rand(T)
```
oxinabox, line 170: We get to remove most of the above.

oxinabox (approved):
remove more variables then merge when you are ready

PR #370, opened 2021-02-12
fix tests for JuliaDiff/ChainRulesCore.jl#298

simeonschaub:


test/rulesets/Random/random.jl
```diff
@@ -46,7 +46,7 @@ Random.rand(d::NormalDistribution) = d.μ + d.σ*randn()
             x, pb = rrule(rand, args...)
             @test x isa xType
             dself, dargs = Iterators.peel(pb(10.0))
-            @test dself isa Zero
+            @test dself == DoesNotExist()
```
oxinabox, line 49: Maybe just
```suggestion
            @test iszero(dself)
```

oxinabox (approved):

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.50"
+version = "0.7.51"
```
oxinabox, line 3: this doesn't require a version bump as it isn't user visable.
But a bump is permitted

simeonschaub: Yeah, you are right, it isn't necessary here.
```suggestion
version = "0.7.50"
```

simeonschaub:
Nightly failures are just #356

PR #374, opened 2021-02-16
structured.jl autotangent

mzgubic:
Helps with #360

### DIFF
--- a/test/rulesets/LinearAlgebra/structured.jl
+++ b/test/rulesets/LinearAlgebra/structured.jl
@@ -2,29 +2,23 @@
     @testset "/ and \\ on Square Matrixes" begin
         @testset "//, $T on the RHS" for T in (Diagonal, UpperTriangular, LowerTriangular)
             RHS = T(randn(T == Diagonal ? 10 : (10, 10)))
-            Y = randn(5, 10)
-            Ȳ = randn(size(/(Y, RHS))...)
-            rrule_test(/, Ȳ, (Y, randn(size(Y))), (RHS, randn(size(RHS))))
+            test_rrule(/, randn(5, 10), RHS)
         end
 
         @testset "\\ $T on LHS" for T in (Diagonal, UpperTriangular, LowerTriangular)
             LHS = T(randn(T == Diagonal ? 10 : (10, 10)))
-            y = randn(10)
-            ȳ = randn(size(\(LHS, y))...)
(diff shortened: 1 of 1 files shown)


test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -2,29 +2,23 @@
...
-        D = Diagonal(randn( N))
-        rrule_test(Diagonal, D, (randn(N), randn(N)))
+        test_rrule(Diagonal, randn(N); output_tangent=randn(N, N))
+        D = Diagonal(randn(N))
+        test_rrule(Diagonal, randn(N); output_tangent=D)
         # Concrete type instead of UnionAll
-        rrule_test(typeof(D), D, (randn(N), randn(N)))
+        test_rrule(typeof(D), randn(N); output_tangent=D)
 
         # TODO: replace this with a `rrule_test` once we have that working
```
mzgubic, line 23: Needs this first
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/146

mzgubic: Actually, maybe better to just close this and treat the TODO separately

test/rulesets/LinearAlgebra/structured.jl
```diff
@@ -112,25 +96,23 @@
...
-            Ā = randn(T, n, m)
             Y = f(A)
             Ȳ_mat = randn(T, m, n)
             Ȳ_composite = Composite{typeof(Y)}(parent=collect(f(Ȳ_mat)))
 
-            rrule_test(f, Ȳ_mat, (A, Ā))
+            test_rrule(f, A; output_tangent=Ȳ_mat)
 
             _, pb = rrule(f, A)
             @test pb(Ȳ_mat) == pb(Ȳ_composite)
```
oxinabox, line 106: I wonder if i we can delete this and just do
```suggestion
test_rrule(f, A)
```
 since a composite is the default

oxinabox (approved):
I would be inclided to get rid of more variables like ybar
but your call

test/rulesets/LinearAlgebra/structured.jl
mzgubic: `test_rrule(f, A)` doesn't work, I've just spent an hour trying to figure out why but giving up now. Should I write an issue in ChainRulesTestUtils and move on with this?

oxinabox: yes

PR #375, opened 2021-02-16
factorizations.jl autotangent

oxinabox:
part of #360

### DIFF
--- a/test/rulesets/LinearAlgebra/factorization.jl
+++ b/test/rulesets/LinearAlgebra/factorization.jl
@@ -29,9 +29,7 @@ end
                 pivot in (Val(true), Val(false)),
                 m in (7, 10, 13)
 
-                A = randn(T, m, n)
-                ΔA = rand_tangent(A)
-                frule_test(lu!, (A, ΔA), (pivot, nothing))
+                test_frule(lu!, randn(T, m, n), pivot ⊢ nothing)
             end
             @testset "check=false passed to primal function" begin
                 Asingular = zeros(n, n)
@@ -40,6 +38,7 @@ end
                     (Zero(), copy(ΔAsingular)), lu!, copy(Asingular), Val(true)
                 )
                 frule((Zero(), ΔAsingular), lu!, Asingular, Val(true); check=false)
+                @test true  # above line would have errored if this was not working right
             end
         end
         @testset "lu rrule" begin
@@ -48,12 +47,7 @@ end
                 pivot in (Val(true), Val(false)),
                 m in (7, 10, 13)
 
-                A = randn(T, m, n)
-                ΔA = rand_tangent(A)
-                F = lu(A, pivot)
-                Δfactors = rand_tangent(F.factors)
-                ΔF = Composite{typeof(F)}(; factors=Δfactors)
-                rrule_test(lu, ΔF, (A, ΔA), (pivot, nothing))
+                test_rrule(lu, randn(T, m, n), pivot ⊢ nothing)
             end
             @testset "check=false passed to primal function" begin
                 Asingular = zeros(n, n)
@@ -62,6 +56,7 @@ end
                 @test_throws SingularException rrule(lu, Asingular, Val(true))
                 _, back = rrule(lu, Asingular, Val(true); check=false)
                 back(ΔF)
+                @test true  # above line would have errored if this was not working right
             end
         end
         @testset "LU" begin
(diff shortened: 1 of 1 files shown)


willtebbutt (approved):
Nice to see the `output_tangent` kwarg being put to good use. Another win!

oxinabox:
Looks like i messed some of these up. Will need some going over to fix

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -188,7 +165,9 @@ end
             n = 10
 
             @testset "eigen!(::Matrix{$T}) frule" for T in (Float64,ComplexF64)
-                X = randn(T, n, n)
+
+                # get a bit away from zero so don't have finite differencing woes
+                X = randn(T, n, n) .+ 5.0
```
oxinabox, line 170: I was getting these huge values like 10^5 in the output of FDM.
Fixing the seed fixes it, but this seems to fix it robustly.
I am guessing there are some values of X that fdm doesn't like,
and before we were lucky enough to not hit them.
but since we are using the RNG a different number of times now, we are not anymore.

Similar happened below

willtebbutt: Hmm interesting. @wesselb or @sethaxen might know what's going on? My suspicion would be that the eigenvalues are flipping around or something? If so, might help to find a way to space them out more or something.

sethaxen: > My suspicion would be that the eigenvalues are flipping around or something?

It could be, but the major source of instability I saw was in the phase of the eigenvectors. The convention LinearAlgebra uses (from LAPACK) is that the entry with the largest magnitude in each eigenvector is rotated to be positive real or negative real, whichever is nearest. So it could become unstable if another entry was close to that one in magnitude. So one way might be to do something like this:
```julia
V = hcat(Random.shuffle([normalize((x .+ 1e-2 .* randn.(ComplexF64)) * sign(randn(ComplexF64))) for x in eachcol(Matrix{Float64}(I, n, n))])...)
```
which creates an identity matrix, adds small random complex numbers to all entries, applies a random phase convention to each eigenvector, then normalizes them. The eigenvectors are then shuffled and made columns of a new matrix. This produces one where the element with the largest magnitude is well-separated from the rest. With a suitable set of eigenvalues, perhaps this would make a well-behaved primal.

oxinabox: Shall we open issue about this,  and merge this PR as is?

willtebbutt: I would be in favour of that. Let's get this in.

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -447,3 +421,4 @@ end
         end
     end
 end
+end
```
oxinabox, line 424:
```suggestion
```

test/rulesets/LinearAlgebra/factorization.jl
oxinabox: hmm still failing on some OS for some julia versions.
Which is annoying, something is funky happening with the RNG

```suggestion
                X = 10 .* (randn(T, n, n) .+ 5.0)
```

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -209,9 +188,11 @@ end
                 end
             end
 
-            @testset "eigen(::Matrix{$T}) rrule" for T in (Float64,ComplexF64)
+            @testset "eigen(::Matrix{$T}) rrule" for T in (Float64, ComplexF64)
                 # NOTE: eigen is not type-stable, so neither are is its rrule
-                X = randn(T, n, n)
+
+                # get a bit away from zero so don't have finite differencing woes
+                X = randn(T, n, n) .+ 5.0
```
oxinabox, line 195:
```suggestion
                X = 10 .* (rand(T, n, n) .+ 5.0)
```

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -209,9 +188,11 @@ end
                 end
             end
 
-            @testset "eigen(::Matrix{$T}) rrule" for T in (Float64,ComplexF64)
+            @testset "eigen(::Matrix{$T}) rrule" for T in (Float64, ComplexF64)
                 # NOTE: eigen is not type-stable, so neither are is its rrule
-                X = randn(T, n, n)
+
+                # get a bit away from zero so don't have finite differencing woes
```
oxinabox, line 194:
```suggestion
                # get a bit away from zero so don't have finite differencing woes
                Random.seed!(1)
```

oxinabox:
bother, I can't seem to fins a matrix that makes this consistently pass.

sethaxen:
I inspected one of the jobs, and at least the values shown in the failure looked really close (i.e. no blow-ups). Perhaps the precision of the test needs to be reduced slightly?

oxinabox:
yep that seems to have done it

PR #381, opened 2021-02-18
Ensure pullback of exp works for immutable arrays

### LINKED ISSUE #380: combining exp and tr cause setindex error from Zygote
this is because Zygote is using `FillArray` but the pullback implementation here assumes the gradient is mutable (so it has `setindex!`) , not sure why Zygote's `tr` rule is not dispatched to ChainRules.

https://github.com/JuliaDiff/ChainRules.jl/blob/master/src/rulesets/LinearAlgebra/matfun.jl#L236

---

I haven't followed the development of Julia AD in the last few months, and now I find these new packages are incredible! But I think this issue is related to something hasn't been well discussed yet: the convention of whether the gradient should be mutable or not, or both. Also sometimes, we want it to be mutable for various reasons,
there is no mechanism to explicitly guarantee this, or maybe should always use BangBang to mutate gradients in rrule? I feel there should be more clarification of the gradient
type interface in docs too, but this might need more discussion

---

XRef: #46

sethaxen:
Fixes #380 by ensuring that the cotangent used by the pullback of `exp` is mutable.

@Roger-luo can you confirm that this fixes the bug?

### DIFF
--- a/src/rulesets/LinearAlgebra/matfun.jl
+++ b/src/rulesets/LinearAlgebra/matfun.jl
@@ -129,7 +129,10 @@ function rrule(::typeof(exp), A0::StridedMatrix{<:BlasFloat})
         A = copy(A0)
         X, intermediates = _matfun!(exp, A)
         function exp_pullback(ΔX)
-            ∂A = _matfun_frechet_adjoint!(exp, ΔX, A, X, intermediates)
+            # Ensures ∂X is mutable. The outer `adjoint` is unwrapped without copy by
+            # the default _matfun_frechet_adjoint!
+            ∂X = Matrix(ΔX')'
+            ∂A = _matfun_frechet_adjoint!(exp, ∂X, A, X, intermediates)
             return NO_FIELDS, ∂A
         end
         return X, exp_pullback
--- a/src/rulesets/LinearAlgebra/matfun.jl
+++ b/src/rulesets/LinearAlgebra/matfun.jl
@@ -131,7 +131,7 @@ function rrule(::typeof(exp), A0::StridedMatrix{<:BlasFloat})
         function exp_pullback(ΔX)
             # Ensures ∂X is mutable. The outer `adjoint` is unwrapped without copy by
             # the default _matfun_frechet_adjoint!
-            ∂X = Matrix(ΔX')'
+            ∂X = ChainRulesCore.is_inplaceable_destination(ΔX) ? ΔX : convert(Matrix, ΔX')'
             ∂A = _matfun_frechet_adjoint!(exp, ∂X, A, X, intermediates)
             return NO_FIELDS, ∂A
         end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.52"
+version = "0.7.53"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


Roger-luo:
yes! thanks!

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -129,7 +129,10 @@ function rrule(::typeof(exp), A0::StridedMatrix{<:BlasFloat})
         A = copy(A0)
         X, intermediates = _matfun!(exp, A)
         function exp_pullback(ΔX)
-            ∂A = _matfun_frechet_adjoint!(exp, ΔX, A, X, intermediates)
+            # Ensures ∂X is mutable. The outer `adjoint` is unwrapped without copy by
+            # the default _matfun_frechet_adjoint!
+            ∂X = Matrix(ΔX')'
```
oxinabox, line 134: Weirdly `copy` promises to return something mutable and is probably cleaner than this?

Roger-luo: I guess `copy` would always allocate, where a type conversion won't? or should there be some dispatch using a mutable trait ?

oxinabox: On checking, I was apparently wrong about `copy`.
It doesn't nesc return something mutable.
I was confused by the fact that it makes some view-like things (including `Adjoint` and `SubArray`, but not including `Diagonal`) into an arrays.

`convert` vs   Constructors is a thing though.

https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Mutable-collections

> `convert(T, x)` is expected to return the original `x` if `x` is already of type `T`. In contrast, if `T` is a mutable collection type then `T(x)` should always make a new collection (copying elements from `x`).

So perhaps this should be a `convert` if we want to avoid allocating unnesc?

---

ChainRulesCore does actually have a trait that might be suitable for this. If we really wanted.
It's part of the mechanics for doing inplace gradient accumulation


>   `is_inplaceable_destination(x) -> Bool`
> 
>   Returns true if `x` is suitable for for storing inplace accumulation of gradients.
>   For arrays this boils down `x .= y` if will work to mutate `x`, if `y` is an appropriate differential.
>   Wrapper array types do not need to overload this if they overload `Base.parent`, and are `is_inplaceable_destination` if and only if their parent array is.
>   Other types should overload this, as it defaults to `false`.

sethaxen: I ended up going with a mixture of these both: Using the trait to decide if we should do anything and then using `convert`, which in this case will always allocate a copy. (`convert(Matrix, X')` will allocate unless `X` is an `Adjoint{T,Matrix{T}}`, but since that type is inplaceable, the trait will bypass the `convert` anyways).

sethaxen:
@roger-luo can you confirm that the latest version still works for you?

src/rulesets/LinearAlgebra/matfun.jl
```diff
@@ -129,7 +129,10 @@ function rrule(::typeof(exp), A0::StridedMatrix{<:BlasFloat})
         A = copy(A0)
         X, intermediates = _matfun!(exp, A)
         function exp_pullback(ΔX)
-            ∂A = _matfun_frechet_adjoint!(exp, ΔX, A, X, intermediates)
+            # Ensures ∂X is mutable. The outer `adjoint` is unwrapped without copy by
+            # the default _matfun_frechet_adjoint!
+            ∂X = ChainRulesCore.is_inplaceable_destination(ΔX) ? ΔX : convert(Matrix, ΔX')'
```
oxinabox, line 134: Why not just:
```suggestion
            ∂X = ChainRulesCore.is_inplaceable_destination(ΔX) ? ΔX : convert(Matrix, ΔX)
```

sethaxen: Because if `∂X'` is an `Adjoint`, then `_matfun_frechet_adjoint!` will `copy` it to make it non-`Adjoint`. This way, we do only one allocation instead of 2. An alternative is to bypass `_matfun_frechet_adjoint!` to call `_matfun_frechet!` directly, but to me this seems cleaner. What do you think?

oxinabox: But if it is an `Adjoint` then isn't it going to be mutable?
Or not becuase it might be an `Adjoint{FillArray}` ?

oxinabox: I guess this makes sense, leave a comment to that effect?

oxinabox (approved):

GiggleLiu:
Actually I and Roger discussed a lot about whether it is an issue of `tr` or `exp`. Are we fixing this issue for every function? or just fix `tr`?

src/rulesets/LinearAlgebra/matfun.jl
sethaxen: Can you suggest a change that would make this comment clearer?
https://github.com/JuliaDiff/ChainRules.jl/blob/608f3af380fe8d7a20085b0d57f2d9ea2e8a2e00/src/rulesets/LinearAlgebra/matfun.jl#L132-L133

sethaxen:
> Actually I and Roger discussed a lot about whether it is an issue of `tr` or `exp`. Are we fixing this issue for every function that calls `setindex!`? or just fix `tr` and `sum`?

I'm not sure I understand the question. This will fix for the case where a user passes an immutable array to the pullback for `exp`. That crops up with Zygote's rule for `tr` but is generic. I wouldn't be surprised if there are other `rrule`s in ChainRules that make incorrect assumptions about mutability, but perhaps that's best discussed in #380 or another issue.

mcabbott:
Does this mutate what's received by the pullback? Is that the cause of this:
```
julia> gradient(x -> 2sum(abs, exp(x)'), [1 2; 3 4.0])
([226.54923657197995 427.40643795569565; 309.849813509321 579.2191099506263],)

julia> gradient(x -> sum(abs, (exp(x) + exp(x))'), [1 2; 3 4.0])
([169.91192742898497 320.5548284667717; 232.38736013199076 434.4143324629697],)
```

PR #388, opened 2021-03-23
Canonicalize Composite{SVD}

mzgubic:
At the moment:
```julia
julia> x = Composite{SVD{Float64,Float64,Array{Float64,2}}}(V = [0.6662838804680036 -0.5008780874375076; 0.729969913633919 0.228331721577141],)

julia> ChainRules.canonicalize(x)
ERROR: ArgumentError: Composite fields do not match primal fields.
Composite fields: (:V,). Primal (SVD{Float64,Float64,Array{Float64,2}}) fields: (:U, :S, :Vt)
```
This seems to make sense because `V` is not a field of `SVD`. However, the `rrule` for `svd` expects the `Ȳ::Composite` to contain `U`, `S`, and `V` fields:
https://github.com/JuliaDiff/ChainRules.jl/blob/76ef95c326e773c6c7140fb56eb2fd16a2af468b/src/rulesets/LinearAlgebra/factorization.jl#L206-L214

I don't know much about svd factorisation so can't really tell why this is the case.

It kind of feels wrong to overload `canonicalize` in this way, but I can't quite tell why. Any thoughts? Also, which file to put this in, it's not really a rule so I don't think it belongs in the `rulesets` folder.

Came across this in https://github.com/FluxML/Zygote.jl/issues/922#issuecomment-805121978

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.54"
+version = "0.7.55"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/ChainRules.jl
+++ b/src/ChainRules.jl
@@ -13,8 +13,7 @@ using Statistics
 
 # Basically everything this package does is overloading these, so we make an exception
 # to the normal rule of only overload via `ChainRulesCore.rrule`.
-import ChainRulesCore: rrule, frule
-
+import ChainRulesCore: rrule, frule, canonicalize
 
 if VERSION < v"1.3.0-DEV.142"
     # In prior versions, the BLAS submodule also exported `dot`, which caused a conflict
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -203,6 +203,18 @@ end
 ##### `svd`
 #####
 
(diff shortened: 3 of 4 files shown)


sethaxen:
It seems like it would be more straightforward to modify the `rrule` for `getproperty(::SVD, ::Symbol)` to store the cotangent of `Vt` instead of `V`. After all `SVD` actually has `Vt` as a field, and `V` is just its `Adjoint`.

mzgubic:
> It seems like it would be more straightforward to modify the `rrule` for `getproperty(::SVD, ::Symbol)` to store the cotangent of `Vt` instead of `V`. After all `SVD` actually has `Vt` as a field, and `V` is just its `Adjoint`.

Thanks for taking a look, that makes sense to me. For some reason I assumed that there is a complicated reason behind using `V`, should have checked that.

For `getproperty` rrule:
https://github.com/JuliaDiff/ChainRules.jl/blob/76ef95c326e773c6c7140fb56eb2fd16a2af468b/src/rulesets/LinearAlgebra/factorization.jl#L217-L230
can we just do something like:
```julia
....
        elseif x === :V
            C(Vt=Ȳ',)
        elseif x === :Vt
            C(Vt=Ȳ,)
...
```
Or is there more to it?

sethaxen:
> can we just do something like:
> ```julia
> ....
>         elseif x === :V
>             C(Vt=Ȳ',)
>         elseif x === :Vt
>             C(Vt=Ȳ,)
> ...
> ```
> 
> Or is there more to it?

I think that's all there is to it. You could add `copy(Ȳ')` to make `Vt` not an `Adjoint`, but this is unnecessary, so long as the `rrule`s for `svd` still work when the cotangent for `Vt` is an `Adjoint`.

PR #391, opened 2021-03-24
[WIP] Complex valued SVD

GiggleLiu:
Sorry for the long delay (see https://github.com/GiggleLiu/BackwardsLinalg.jl/issues/17). I tried multiple times to add the backward rules to ChainRules, but feeling hard to write tests that fitting this framework. Can some one help me?

* added a new function for complex valued svd back-propagation. But feeling hard to test functions with gauge problem.
* added a `safe_inv` function to invert the singular matrix subtraction matrix safely. Otherwise, it can break up easily when there are degenerate spectrums.
* also, I am planning to fix the real valued SVD to reduce the number of matrix multiplication.

### Refs
*   [Automatic Differentiation for Complex Valued SVD](https://arxiv.org/abs/1909.02659)
*   https://j-towns.github.io/papers/svd-derivative.pdf
*   https://giggleliu.github.io/2019/04/02/einsumbp.html

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -203,7 +203,7 @@ end
 ##### `svd`
 #####
 
-function rrule(::typeof(svd), X::AbstractMatrix{<:Real})
+function rrule(::typeof(svd), X::AbstractMatrix{<:Union{Real, Complex}})
     F = svd(X)
     function svd_pullback(Ȳ::Composite)
         # `getproperty` on `Composite`s ensures we have no thunks.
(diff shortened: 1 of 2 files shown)


sethaxen:
Before I review, can you comment on how this implementation differs from the existing real one? That is, it looks like the current real implementation would also work on complex numbers if we released the real type constraint, so I'm wondering how this rule differs from that one, and can they be unified.

GiggleLiu:
> Before I review, can you comment on how this implementation differs from the existing real one? That is, it looks like the current real implementation would also work on complex numbers if we released the real type constraint, so I'm wondering how this rule differs from that one, and can they be unified.

Sure, if you open the third link, you will see the red term. That one is the missing term in the real version. For more detailed description, you need to check the first link to the original paper.

GiggleLiu:
Also, please note the `safe_inv` is important in some applications. e.g. In many applications in physics, there are degenerate singular values due to the high symmetry.
Normally, when the denominator `|s_i^2 - s_j^2|` is zero, the numerator should also be zero (the loss is gauge invariant). However, it assumption is vulnerable due to rounding errors, and this is why we need to handle this case manually to avoid fake singularity.

PR #393, opened 2021-03-29
force sym/herm back to return Matrix (1.6 inference fix)

oxinabox:
Fixed the type inference failures we have been seeing in 1.6.
They show up because:

On 1.5:
```julia
julia> L = LowerTriangular(rand(3,3));

julia> U = UpperTriangular(rand(3,3));

julia> typeof(L .+ U')
Array{Float64,2}
```

On 1.6:
```julia
julia> L = LowerTriangular(rand(3,3));

julia> U = UpperTriangular(rand(3,3));

julia> typeof(L .+ U')
LowerTriangular{Float64, Matrix{Float64}}
```

Which honestly is kinda cool, but not what we want if we are to be type stable
Possibly we should push `uplo` into the type domain with a value type,
but that is beyond the scope of this PR.
The other branchs force the result to be a `Matrix` so the last should too

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.55"
+version = "0.7.56"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/LinearAlgebra/symmetric.jl
+++ b/src/rulesets/LinearAlgebra/symmetric.jl
@@ -68,7 +68,7 @@ end
         end
     end
     L, U, D = LowerTriangular(ΔΩ), UpperTriangular(ΔΩ), Diagonal(ΔΩ)
-    return uplo === :U ? U .+ transpose(L) - D : L .+ transpose(U) - D
+    return Matrix(uplo === :U ? U .+ transpose(L) - D : L .+ transpose(U) - D)
 end
 
 @inline function _hermitian_back(ΔΩ, uplo::Symbol)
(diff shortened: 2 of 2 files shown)


willtebbutt (approved):
Patch bump please :)

mzgubic (approved):
Thanks for solving this

sethaxen:
Perhaps instead of converting to `Matrix`, it would be better to explicitly call `parent` if the produced cotangent is a `LowerTriangular` and then use `tril!`?

oxinabox:
> Perhaps instead of converting to Matrix, it would be better to explicitly call parent if the produced cotangent is a LowerTriangular and then use tril!?

We call `Matrix` in ther over branchs of this.

If there is a more efficient way should it be PRed to Base for `convert` ? (convert right now falls back to `Matrix` unless result is already a `Matrix`)
Possibly this code in all branchs should be rewritten to use convert to avoid copying in the case it is already a `Matrix`

sethaxen:
> We call `Matrix` in ther over branchs of this.

Oh you're right, then yes, this is fine. But a future PR should fix that.

> Possibly this code in all branchs should be rewritten to use convert to avoid copying in the case it is already a `Matrix`

My suggestion is to avoid converting to `Matrix` at all. Certain non-`Matrix` types have good support for being wrapped with `Symmetric`, e.g. `StaticMatrix`. Maybe sparse matrices. So in the spirit of the discussion in #232 it would be good to avoid making the cotangent a `Matrix`. The idea would be to let broadcast sort it out but handle this potential type-instability by calling `parent` on an `AbstractTriangular` matrix. But this can be a future PR.

> If there is a more efficient way should it be PRed to Base for `convert` ? (convert right now falls back to `Matrix` unless result is already a `Matrix`)

A more efficient way to do what?

sethaxen (approved):

Issue #392, opened 2021-03-29
Rules for `sort` and friends


mzgubic:
Add rules for

- [x] `sort`
- [x] `sortperm`
- [x] `sortperm!`
- [ ] `sortslices` (kind of tricky because there is no easy `sortslicesperm` function)
- [x] `insorted`
- [x] `partialsort`
- [x] `partialsortperm`
- [x] `partialsortperm!`

PR #394, opened 2021-03-31
Add Rules for sortperm and friends

andrewrosemberg:
Add rules for `sortperm`, `sortperm!`,`partialsortperm`, `partialsortperm!`

Starts https://github.com/JuliaDiff/ChainRules.jl/issues/392


mzgubic (commented):
Could you please bump the patch?

src/rulesets/Base/nondiff.jl
```diff
@@ -289,6 +289,7 @@ VERSION >= v"1.3" && @non_differentiable istaskfailed(::Task)
 VERSION >= v"1.4" && @non_differentiable only(::Char)
 @non_differentiable open(::Any)
 
+@non_differentiable partialsortperm(::AbstractArray{Any}, ::Union{Integer, OrdinalRange})
```
mzgubic, line 292:
```suggestion
@non_differentiable partialsortperm(::AbstractVector{Any}, ::Union{Integer, OrdinalRange})
```
I think this is how the `partialsortperm` is constrained, so we should probably do the same

src/rulesets/Base/nondiff.jl
```diff
@@ -350,6 +351,7 @@ end
 @non_differentiable size(::Any, ::Any)
 @non_differentiable sizeof(::Any)
 @non_differentiable sleep(::Any)
+@non_differentiable sortperm(::AbstractArray{Any})
```
mzgubic, line 354: similarly here
```suggestion
@non_differentiable sortperm(::AbstractVector{Any})
```

mzgubic (approved):
That's great, thanks!

If you could add the other friends (`sort`, etc) we can close https://github.com/JuliaDiff/ChainRules.jl/issues/392

andrewrosemberg:
Ready for final review :)

mzgubic:
Oh, of course, this is embarrassing.

oxinabox:
I think `sortperm` OTOH is nondifferentiable, in the sense that
it is `Zero`s most places and discontinous at the others.

One way to implement `sort` is as `sortperm` + `getindex`.
idk if that is the one Base uses or if it uses `copy` + `sort!`.
If Base does use that one  then ADs (including Zygote) should be able to handle `sort` as long as `sortperm` is marked nondifferentiable.
If it is the one that uses `sort!` + `copy` ADs tht can't handle mutation maybe can't.

If needed then this is a good place for a augmented primal.

I think it the rrule for sort is something like
```julia
function rrule(typeof(sort), xs::Vector{<:Real}; kwargs...)
    inds = sortperm(xs; kwargs
    ys = xs[inds]
    function sort_pullback(dys)
        # got to get everything back into the original order. Do this by assigning based on new order
        dxs = simular(dys)
        dxs[inds] .= dys
        return NO_FIELDS, dxs
    end
    return ys, sort_pullback
end
```

mzgubic:
Yeah the initial PR was just `sortperm` and then I asked Andrew to add the other ones that I didn't think about properly, my brain is slow today

We can just merge the ones he needs and I am happy to add the other ones later

andrewrosemberg:
PR updated, ready for review :)

PR #395, opened 2021-04-01
rrule for sort and partialsort

mzgubic:
After #394  `sortperm` and `partialsortperm` rrule are in and Zygote can compute `sort` on its own, but not `partialsort`.

no speed gain for the `sort` rule in Zygote as of now but supporting thunks will make it faster in the future, or for AD systems which natively support ChainRules types.

Helps with #392

### DIFF
--- a/src/rulesets/Base/sort.jl
+++ b/src/rulesets/Base/sort.jl
@@ -4,18 +4,15 @@ function rrule(::typeof(partialsort), xs::AbstractVector, k::Union{Integer,Ordin
 
     function partialsort_pullback(Δys)
         function partialsort_add!(Δxs)
-            for (Δy, i) in zip(Δys, inds)
-                Δxs[i] += Δy
-            end
+            Δxs[inds] += Δys
             return Δxs
         end
 
-        Δxs = InplaceableThunk(
-            @thunk(partialsort_add!(zero(xs))),
-            partialsort_add!
-        )
+        Δxs = InplaceableThunk(@thunk(partialsort_add!(zero(xs))), partialsort_add!)
+
         return NO_FIELDS, Δxs, DoesNotExist()
     end
+
     return ys, partialsort_pullback
 end
 
@@ -29,10 +26,7 @@ function rrule(::typeof(sort), xs::AbstractVector; kwargs...)
             return Δxs
         end
 
-        Δxs = InplaceableThunk(
-            @thunk(sort_add!(zero(xs))),
-            sort_add!
-        )
+        Δxs = InplaceableThunk(@thunk(sort_add!(zero(Δys))), sort_add!)
 
         return NO_FIELDS, Δxs
     end
--- a/src/rulesets/Base/sort.jl
+++ b/src/rulesets/Base/sort.jl
@@ -1,0 +1,40 @@
+function rrule(::typeof(partialsort), xs::AbstractVector, k::Union{Integer,OrdinalRange}; kwargs...)
+    inds = partialsortperm(xs, k; kwargs...)
+    ys = xs[inds]
+
+    function partialsort_pullback(Δys)
+        function partialsort_add!(Δxs)
+            for (Δy, i) in zip(Δys, inds)
+                Δxs[i] += Δy
+            end
+            return Δxs
+        end
+
+        Δxs = InplaceableThunk(
+            @thunk(partialsort_add!(zero(xs))),
+            partialsort_add!
+        )
+        return NO_FIELDS, Δxs, DoesNotExist()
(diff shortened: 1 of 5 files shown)


mzgubic:
CI is weird, only Mac seems to be failing, and on 1.0 only, and it doesn't look like it starts at all?

src/rulesets/Base/sort.jl
```diff
@@ -0,0 +1,40 @@
...
+    function partialsort_pullback(Δys)
+        function partialsort_add!(Δxs)
+            for (Δy, i) in zip(Δys, inds)
+                Δxs[i] += Δy
+            end
+            return Δxs
+        end
+
+        Δxs = InplaceableThunk(
+            @thunk(partialsort_add!(zero(xs))),
```
oxinabox, line 14:
```suggestion
            @thunk(partialsort_add!(zero(Δys))),
```

I think it would be better to do `Δys` than `xs` to get our zeros from.
that way we don't close over `xs`
also because `Δys` definately has the right type to hold these results.
Which could matter for things with dervivative that are not the same type as them.

src/rulesets/Base/sort.jl
```diff
@@ -0,0 +1,40 @@
...
+    ys = xs[inds]
+
+    function sort_pullback(Δys)
+        function sort_add!(Δxs)
+            Δxs[inds] += Δys
+            return Δxs
+        end
+
+        Δxs = InplaceableThunk(
+            @thunk(sort_add!(zero(xs))),
```
oxinabox, line 33:
```suggestion
            @thunk(sort_add!(zero(Δys))),
```

src/rulesets/Base/sort.jl
```diff
@@ -0,0 +1,40 @@
...
+            for (Δy, i) in zip(Δys, inds)
+                Δxs[i] += Δy
+            end
+            return Δxs
+        end
+
+        Δxs = InplaceableThunk(
+            @thunk(partialsort_add!(zero(xs))),
+            partialsort_add!
+        )
```
oxinabox, line 16: What do you think about writing short inplacable thunks like this on one line?

oxinabox (approved):
fix comments then should be good

src/rulesets/Base/sort.jl
```diff
@@ -0,0 +1,40 @@
+function rrule(::typeof(partialsort), xs::AbstractVector, k::Union{Integer,OrdinalRange}; kwargs...)
+    inds = partialsortperm(xs, k; kwargs...)
+    ys = xs[inds]
```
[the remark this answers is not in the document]
mzgubic: I tried doing
```julia
julia> y, pb = rrule(partialsort, a, 2); @btime pb(3.0)
```
with and without `@inbounds` but the results are pretty much the same

src/rulesets/Base/sort.jl
```diff
@@ -0,0 +1,40 @@
...
+
+    function partialsort_pullback(Δys)
+        function partialsort_add!(Δxs)
+            for (Δy, i) in zip(Δys, inds)
+                Δxs[i] += Δy
+            end
```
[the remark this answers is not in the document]
mzgubic: nice catch, done. without the dot though, because
```julia
julia> Δxs[4] .+= 2
ERROR: MethodError: no method matching copyto!(::Float64, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0}, Tuple{}, typeof(+), Tuple{Float64, Int64}})
Closest candidates are:
  copyto!(::Any, ::Base.Broadcast.Broadcasted{var"#s155", Axes, F, Args} where {var"#s155"<:StaticArrays.StaticArrayStyle, Axes, F, Args<:Tuple}) at /Users/mzgubic/.julia/packages/StaticArrays/rdb0l/src/broadcast.jl:29
  copyto!(::AbstractArray, ::Base.Broadcast.Broadcasted{var"#s832", Axes, F, Args} where {var"#s832"<:Base.Broadcast.AbstractArrayStyle{0}, Axes, F, Args<:Tuple}) at broadcast.jl:939
  copyto!(::AbstractArray, ::Base.Broadcast.Broadcasted) at broadcast.jl:936
  ...
```

src/rulesets/Base/sort.jl
mzgubic: Done this for `sort`.

However, in `partialsort` `Δys` can be a float (if `k` is an integer), so I think we need `zero(xs)`?

oxinabox: fair

PR #396, opened 2021-04-06
Update to using `DoesNotExist()` instead of `nothing` in tests

mzgubic:
Depends on https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/129

For integers and symbols the explicit tangent is removed as `rand_tangent` returns `DoesNotExist()`

### DIFF
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -1,6 +1,6 @@
 @testset "reshape" begin
-    test_rrule(reshape, rand(4, 5), (2, 10) ⊢ nothing)
-    test_rrule(reshape, rand(4, 5), 2 ⊢ nothing, 10 ⊢ nothing)
+    test_rrule(reshape, rand(4, 5), (2, 10) ⊢ DoesNotExist())
+    test_rrule(reshape, rand(4, 5), 2, 10)
 end
 
 @testset "hcat" begin
@@ -14,7 +14,7 @@ end
     A = randn(3, 2)
     B = randn(3, 1)
     C = randn(3, 3)
-    test_rrule(reduce, hcat ⊢ nothing, [A, B, C])
+    test_rrule(reduce, hcat ⊢ DoesNotExist(), [A, B, C])
 end
 
 @testset "vcat" begin
@@ -28,10 +28,10 @@ end
     A = randn(2, 4)
     B = randn(1, 4)
     C = randn(3, 4)
-    test_rrule(reduce, vcat ⊢ nothing, [A, B, C])
+    test_rrule(reduce, vcat ⊢ DoesNotExist(), [A, B, C])
 end
 
 @testset "fill" begin
-    test_rrule(fill, 44.0, 4 ⊢ nothing; check_inferred=false)
-    test_rrule(fill, 2.0, (3, 3, 3) ⊢ nothing)
+    test_rrule(fill, 44.0, 4; check_inferred=false)
+    test_rrule(fill, 2.0, (3, 3, 3) ⊢ DoesNotExist())
 end
--- a/test/rulesets/Base/indexing.jl
+++ b/test/rulesets/Base/indexing.jl
@@ -3,48 +3,48 @@
         x = [1.0 2.0 3.0; 10.0 20.0 30.0]
 
         @testset "single element" begin
-            test_rrule(getindex, x, 2 ⊢ nothing)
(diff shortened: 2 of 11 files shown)


test/rulesets/Base/array.jl
```diff
@@ -14,7 +14,7 @@ end
     A = randn(3, 2)
     B = randn(3, 1)
     C = randn(3, 3)
-    test_rrule(reduce, hcat ⊢ nothing, [A, B, C])
+    test_rrule(reduce, hcat ⊢ DoesNotExist(), [A, B, C])
```
oxinabox, line 17: `hcat ⊢ DoesNotExist()`
this makes me think that `NO_FIELDS` should be `DoesNotExist()` rather than `Zero()`

mzgubic: I was surprised that `NO_FIELDS` was `Zero()` rather than `DoesNotExist()` when I first saw it to be honest.

oxinabox: It is a discussion we have had a few times and flipped back and forth.
It's why it is a constant rather than being hard-coded

test/rulesets/Base/indexing.jl
```diff
@@ -3,48 +3,48 @@
...
-            test_rrule(getindex, x, [2,3] ⊢ nothing)
+            test_rrule(getindex, x, 2:3 ⊢ DoesNotExist())
+            test_rrule(getindex, x, 3:-1:2 ⊢ DoesNotExist())
+            test_rrule(getindex, x, [3,2] ⊢ DoesNotExist())
+            test_rrule(getindex, x, [2,3] ⊢ DoesNotExist())
 
-            test_rrule(getindex, x, 1:2 ⊢ nothing, 2:3 ⊢ nothing)
-            test_rrule(getindex, x, (:) ⊢ nothing, 2:3 ⊢ nothing)
+            test_rrule(getindex, x, 1:2 ⊢ DoesNotExist(), 2:3 ⊢ DoesNotExist())
+            test_rrule(getindex, x, (:) ⊢ DoesNotExist(), 2:3 ⊢ DoesNotExist())
```
oxinabox, line 20: `:` should have this as the default.

test/rulesets/Base/indexing.jl
```diff
@@ -3,48 +3,48 @@
...
 
-            test_rrule(getindex, x, 1:2 ⊢ nothing, 1 ⊢ nothing)
-            test_rrule(getindex, x, 1 ⊢ nothing, 1:2 ⊢ nothing)
+            test_rrule(getindex, x, 1:2 ⊢ DoesNotExist(), 1)
```
oxinabox, line 22: should we have 
```
function rand_tangent(x::AbstractArray)
    dx = map(rand_tangent, x)
    return all(dxi -> dxi isa DoesNotExist, dx) ? DoesNotExist() : dx
end
```
or something else like that to condense multiple representations of zero

test/rulesets/Base/indexing.jl
mzgubic: some discussion here https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/133

oxinabox (approved):

test/rulesets/Base/indexing.jl
nickrobinson251: that issue doesn't meantion `Colon()` which i think is what @oxinabox meant?
i.e. that we should be able to have:
```suggestion
            test_rrule(getindex, x, (:), 2:3 ⊢ DoesNotExist())
```

mzgubic:
I've made issues for all the discussion points, they are not blocking for this PR right @oxinabox ?

oxinabox:
Correct

oxinabox (approved):

test/rulesets/LinearAlgebra/blas.jl
```diff
@@ -58,7 +58,7 @@
             @testset "Array{$T,$N}" for N in eachindex(dims), T in (Float64, ComplexF64)
                 s = (dims[1] * incx, dims[2:N]...)
                 n = div(prod(s), incx)
-                test_rrule( BLAS.asum, n ⊢ nothing, randn(T, s), incx ⊢ nothing)
+                test_rrule( BLAS.asum, n, randn(T, s), incx)
```
mzgubic, line 61:
```suggestion
                test_rrule(BLAS.asum, n, randn(T, s), incx)
```

PR #398, opened 2021-04-17
non_diff `typejoin` and `Base.depwarn`

DhairyaLGandhi:
cc @willtebbutt

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.59"
+version = "0.7.60"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


willtebbutt:
LGTM - just needs a patch bump.

willtebbutt:
And a test.

willtebbutt (approved):
LGTM. Will merge when tests pass.

DhairyaLGandhi:
https://github.com/JuliaDiff/ChainRules.jl/runs/2370533443?check_suite_focus=true#step:6:30 suggests there is a missing symbol.

DhairyaLGandhi:
Does `test_rrule` respect types as input? Should I use a different function?

DhairyaLGandhi:
`rand_tangent` sure doesn't like being given types, since it would annotate types to go to `DataType`

https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/8b5e9067cec52cf97da5cceec9bc040460c5cccd/src/generate_tangent.jl#L70

DhairyaLGandhi:
LMK what the correct way to write a test for types etc is

willtebbutt:
Yeah, I think we're just missing a `rand_tangent` for it -- we should probably write a mini test-suite to check things that have been `@non_differentiable`d.

For now, could you just check that the rules aren't nothing?
Something like
```julia
@test frule((NO_FIELDS, DoesNotExist(), DoesNotExist()), typejoin, type1, type2) !== nothing
@test rrule(typejoin, type1, type2) !== nothing
```

willtebbutt (approved):
Failures on nightly, but otherwise seems fine, so I'll merge.

Issue #383, opened 2021-02-21
rules for `repeat` are missing


CarloLucibello:
Some of them can be ported over from Zygote. 

cf. https://github.com/FluxML/Zygote.jl/issues/906
https://github.com/FluxML/Zygote.jl/blob/956cbcf3c572c0eb09c146189bb38b1b434634ff/src/lib/array.jl#L130

Dsantra92:
Can I take over this issue?

willtebbutt:
Please do :) Happy to review a PR resolving this!

mzgubic:
Yes, sure, you are very welcome to

willtebbutt:
It would be great if you could do it in two parts:
1. port across the current `repeat` rules from `Zygote`.
2. add the missing rules.

Doing it in two steps will make it really easy for us to review :)

cossio:
Bump.

Please note that the rules for Zygote are incomplete: https://github.com/FluxML/Zygote.jl/issues/950

nickrobinson251:
PRs welcome! The docs should help introduce the idea of `rrule`s and how to write them, for anyone who want to contribute :) (https://juliadiff.org/ChainRulesCore.jl/dev/#frule-and-rrule)

mzgubic:
I've recommended this issue to https://github.com/Wimmerer (can't seem to @ him) whom I will (likely) co-mentor for his `GraphBLAS x ChainRules` GSoC project

mcabbott:
Zygote has several different implementations for the repeat gradient, https://github.com/FluxML/Zygote.jl/blob/2468e2f1f787e42e9c3d0aa93cec58a0488f8573/src/lib/array.jl#L149-L173 , and the one you're timing seems to be 300x slower than the other one:
```julia
julia> xs = collect(1:100)

julia> y_in7, back_in7 = Zygote.pullback(xs) do x  # same as above
           repeat(x, inner=(7, ))
       end;

julia> @btime back_in7($(ones(size(y_in7))));
  144.166 μs (4202 allocations: 132.14 KiB)

julia> y_out7, back_out7 = Zygote.pullback(xs) do x  # also uses slow implementation
           repeat(x, outer=(7, ))
       end;

julia> @btime back_out7($(ones(size(y_out7))));
  144.167 μs (4202 allocations: 132.14 KiB)

julia> y_vec7, back_vec7 = Zygote.pullback(xs) do x
           repeat(x, 7)
       end;

[...]

PR #405, opened 2021-04-30
[WIP] Add Zygote repeat rrules

Wimmerer:
These are as direct of translations as you can get from Zygote. The last rule, however, is not working because of kwargs shenanigans. Not sure how this should work

Edit: Works towards #383


Wimmerer:
Will add missing rrules and frules in another PR

mzgubic:
It looks like you are passing optional arguments rather than kwargs actually, kwargs need `;`

Wimmerer:
`Error During Test at /home/will/.julia/packages/ChainRulesTestUtils/bDd51/src/check_result.jl:19
  Test threw exception
  Expression: isapprox(actual, expected; kwargs...)
  MethodError: no method matching isapprox(::Array{Float64, 3}, ::Array{Float64, 3}; rtol=1.0e-9, atol=1.0e-9, inner=(2, 2, 1), outer=(1, 1, 3))`

src/rulesets/Base/array.jl
```diff
@@ -103,3 +103,36 @@ function rrule(::typeof(fill), value::Any, dims::Int...)
...
+    function repeat_pullback(Ȳ)
+        Ȳ′ = zero(xs)
+        S = size(xs)
+        for (dest_idx, val) ∈ pairs(IndexCartesian(), Ȳ)
+            src_idx = [mod1(div(dest_idx[dim] - 1, inner[dim]) + 1, S[dim]) for dim ∈ 1:length(S)]
+            Ȳ′[src_idx...] += val
+        end
+        return (NO_FIELDS, Ȳ′)
+    end
+    return repeat(xs; inner = inner, outer = outer), repeat_pullback
```
mzgubic, line 137:
```suggestion
    return repeat(xs; inner=inner, outer=outer), repeat_pullback
```

test/rulesets/Base/array.jl
```diff
@@ -35,3 +35,9 @@ end
     test_rrule(fill, 44.0, 4; check_inferred=false)
     test_rrule(fill, 2.0, (3, 3, 3) ⊢ DoesNotExist())
 end
+
+@testset "repeat" begin
+    test_rrule(repeat, randn(5), 3)
+    test_rrule(repeat, randn(3,3), 2)
+    test_rrule(repeat, randn(5,4,3); fkwargs=(inner=(2,2,1), outer=(1,1,3)))
```
mzgubic, line 42: Should we add a few more tests? E.g. both matrix and vectors for both 1 and two indices. And similarly a few edge cases for the inner outer one?

Perhaps we could also test for `Float64[]` or `fill(4.0)`, both zero dimensional arrays

mzgubic (commented):
Thanks, looks good overall! I've added a few style changes and a couple of minor suggestions.

test/rulesets/Base/array.jl
Wimmerer: What do you want the result to be for the empty array? Repeat just returns a zero-dim array for that.

src/rulesets/Base/array.jl
```diff
@@ -103,3 +103,52 @@ function rrule(::typeof(fill), value::Any, dims::Int...)
...
+            src_idx = [mod1(div(dest_idx[dim] - 1, inner[dim]) + 1, S[dim]) for dim ∈ 1:length(S)]
+            Ȳ′[src_idx...] += val
+        end
+        return (NO_FIELDS, Ȳ′)
+    end
+    return repeat(xs; inner=inner, outer=outer), repeat_pullback
+end
+
+function rrule(::typeof(repeat), x::AbstractArray{<:Real, 0}, m::Integer)
+    repeat_pullback(Ȳ) = (NO_FIELDS, similar(x, eltype(Ȳ)) .= sum(Ȳ), DoesNotExist())
```
mzgubic, line 140:
```suggestion
```

test/rulesets/Base/array.jl
```diff
@@ -35,3 +35,18 @@ end
...
+    test_rrule(repeat, randn(3, 3), 2)
+    test_rrule(repeat, randn(5, 5), 2,5)
+    test_rrule(repeat, randn(5, 4, 3); fkwargs=(inner=(2, 2, 1), outer=(1, 1, 3)))
+    test_rrule(repeat, fill(4.0), 3)
+    test_frule(repeat, randn(5), 3)
+    test_frule(repeat, randn(5), 3,3)
+    test_frule(repeat, randn(3, 3), 2)
+    test_frule(repeat, randn(3, 3), 2,5)
+    test_frule(repeat, randn(5, 4, 3); fkwargs=(inner=(2, 2, 1), outer=(1, 1, 3)))
+    test_frule(repeat, fill(4.0), 3)
```
mzgubic, line 51:
```suggestion
    test_rrule(repeat, randn(5), 3)
    test_rrule(repeat, randn(5), 3, 3)
    test_rrule(repeat, randn(3, 3), 2)
    test_rrule(repeat, randn(5, 5), 2,5)
    test_rrule(repeat, randn(5, 4, 3); fkwargs=(inner=(2, 2, 1), outer=(1, 1, 3)))
    test_rrule(repeat, fill(4.0), 3)
    
    @testset "frule" begin
        test_frule(repeat, fill(4.0), 3)
        
        test_frule(repeat, randn(5), 3)
        test_frule(repeat, randn(5), 3, 3)
        
        test_frule(repeat, randn(3, 3), 2)
        test_frule(repeat, randn(3, 3), 2, 5)
        
        test_frule(repeat, randn(5, 4, 3); fkwargs=(inner=(2, 2, 1), outer=(1, 1, 3)))
    end
```

and similarly for the rrule to keep things tidy

src/rulesets/Base/array.jl
```diff
@@ -103,3 +103,40 @@ function rrule(::typeof(fill), value::Any, dims::Int...)
...
+#####
+
+function rrule(::typeof(repeat), x::AbstractVector, m::Integer)
+    function repeat_pullback(Ȳ)
+        return (NO_FIELDS, dropdims(sum(reshape(Ȳ, length(x), :); dims=2); dims=2), DoesNotExist())
+    end
+    return repeat(x, m), repeat_pullback
+end
+
+function rrule(::typeof(repeat), x::AbstractVecOrMat, m::Integer, n::Integer=1)
```
mcabbott, line 118: Since things like `repeat(rand(Int8,2,2,2),1,1,2,2)` are allowed, it might be nice to treat the general case? Untested, but my attempt is: 
```
function rrule(::typeof(repeat), x::AbstractArray, scales::Integer...)
    function repeat_pullback_1(dy)
        size2ndims = ntuple(d -> isodd(d) ? size(x,1+d÷2) : get(scales,d÷2,1), 2*ndims(dy))
        sumdy = sum(reshape(dy, size2ndims); dims = ntuple(d -> 2d, ndims(dy)))
        return (NO_FIELDS, reshape(sumdy, size(x)), map(_->DoesNotExist(), scales)...)
    end
    return repeat(x, scales...), repeat_pullback_1
end
```
What this won't handle is `repeat(1:2, 3,2,1,0)` but perhaps nobody will write that!

src/rulesets/Base/array.jl
```diff
@@ -103,3 +103,40 @@ function rrule(::typeof(fill), value::Any, dims::Int...)
...
     return fill(value, dims), fill_pullback
 end
+
+#####
+##### `repeat`
+#####
+
+function rrule(::typeof(repeat), x::AbstractVector, m::Integer)
+    function repeat_pullback(Ȳ)
+        return (NO_FIELDS, dropdims(sum(reshape(Ȳ, length(x), :); dims=2); dims=2), DoesNotExist())
```
nickrobinson251, line 113: should this close over `length(x)` rather than `x` (to reduce memory requirements)?
```suggestion
    length_x = length(x)
    function repeat_pullback(Ȳ)
        return (NO_FIELDS, dropdims(sum(reshape(Ȳ, length_x, :); dims=2); dims=2), DoesNotExist())
```
And likewise with the `size` calls in the rules below. 

This is a genuine question, i'm not sure what best practice here is (cc @oxinabox).

I believe this would be manually doing one thing we want an `OpaqueClosure` to be able to do automatically. 
Also maybe this should be mentioned on the "writing good rules docs"?

oxinabox: Yeah, I am cautious about doing that since it makes code uglier.
So I wouldn't block a PR over it (it can always be optimized later)
But it does save memory.
We should discuss this with pros and cons in "Writing God Rules"

mcabbott:
Closed by #460

PR #407, opened 2021-05-07
Update rules for SpecialFunctions

devmotion:
This updates the rules for SpecialFunctions, in the same way as https://github.com/JuliaMath/SpecialFunctions.jl/pull/308.

I think it might be necessary to fix SpecialFunctions to an older version without ChainRulesCore dependency if we want the rules in ChainRules to be tested properly. However, this might cause problems with other test dependencies (older versions etc.), so maybe it should be done in a separate environment?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.61"
+version = "0.7.62"
 
 [deps]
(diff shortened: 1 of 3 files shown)


oxinabox:
>  if we want the rules in ChainRules to be tested properly. 

TBH i don't care about them being tested properly.
I think noone is using them.

oxinabox (approved):

devmotion:
Test failures with Julia nightly seem to be unrelated and exist on master as well.

PR #413, opened 2021-05-18
rename differentials

mzgubic:
follow up to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/353

### DIFF
--- a/src/ChainRules.jl
+++ b/src/ChainRules.jl
@@ -2,6 +2,7 @@ module ChainRules
 
 using Reexport
 @reexport using ChainRulesCore
+export Zero, DoesNotExist, Composite, AbstractDifferential
 
 using Base.Broadcast: materialize, materialize!, broadcasted, Broadcasted, broadcastable
 using Compat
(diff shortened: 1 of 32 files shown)


mzgubic:
`ReversePropagation.jl` integration test is broken, possibly because of `@reexport` of ChainRulesCore types?

mzgubic:
There's a PR to fix the ReversePropagation integration test: https://github.com/dpsanders/ReversePropagation.jl/pull/38.

I think the issue is that `@reexport` macro does not reexport the deprecations, see https://github.com/simonster/Reexport.jl/issues/28

So this should be good to go.

oxinabox:
> I think the issue is that @reexport macro does not reexport the deprecations, see simonster/Reexport.jl#28

We should just reexport them manually then.

src/ChainRules.jl
```diff
@@ -2,6 +2,7 @@ module ChainRules
 
 using Reexport
 @reexport using ChainRulesCore
```
oxinabox, line 4: Why not go all the way and reexport everything by hand and then we can drop the dependency on Reexport.jl?

oxinabox (approved):

src/ChainRules.jl
mzgubic: Because if we add an export to ChainRulesCore we also have to modify ChainRules. And Reexport is a one-file package with no dependencies.

PR #416, opened 2021-05-27
Use true instead of One()

mzgubic:
Going towards https://github.com/JuliaDiff/ChainRulesCore.jl/issues/354


oxinabox (approved):
Can a few of these be  benchmarked to make sure it it doesn't introduce performance regressions?

Also it looks a bit funny to use `true`, should we add an explict note to `@scalar_rule`'s docstring about this?

mzgubic:
> Can a few of these be benchmarked to make sure it it doesn't introduce performance regressions?

for reference https://github.com/JuliaDiff/ChainRulesCore.jl/pull/357#discussion_r640863115

Issue #415, opened 2021-05-25
Should we be reexporting ChainRulesCore.jl?


oxinabox:
Right now we reexport ChainRulesCore.jl
this was on the basis that AD package should depend only on ChainRules.jl not on ChainRulesCore.
But it feels more natural to me for AD packages to depend on both.
and it gives a bit more control on compat

willtebbutt:
What sort of thing would package maintainers be able to do that they can't do now if we don't re-export?

oxinabox:
It wouldn't change much at all.
Except it would mean a breaking change to ChainRulesCore would not nesc be a breaking change to ChainRules.jl
though I don't know how that is possible.

willtebbutt:
Ahh I see. Okay, well I don't have an especially strong opinion on the matter -- seems to me that the current approach is probably fine though.

mzgubic:
One practical issue with this came up when renaming ChainRulesCore differential types, i.e. the `@reexport` macro does not reexport deprecated bindings, see https://github.com/simonster/Reexport.jl/issues/28.

PR #417, opened 2021-05-27
Replace `NO_FIELDS` by `NoTangent()`

mzgubic:
Follow up to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/358

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.7.65"
+version = "0.7.68"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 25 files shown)


oxinabox (approved):

mzgubic:
Are we ok with this as a non-breaking change?

oxinabox:
The problem with #415  is that it makes any change to CHainRulesCore also breaking for us. so I think we have to say this is breaking.
While we do so maybe we want to also address #415 + delete SpecialFunctions.jl code + delete NaNMath code (noone uses it AFAIK)

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.8.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
```
oxinabox, line 7: seems to have been added by mistake as a diriect dependency

```suggestion
```

mzgubic:
I think we can ignore the SpecialFunctions failure since we will delete those anyway

PR #421, opened 2021-05-28
fix `redirect_stdio` rules on nightly again

simeonschaub:
These were broken by JuliaLang/julia#37978 again.


oxinabox (approved):

simeonschaub:
I opened #422 for tracking the remaining failures.

oxinabox:
Merge when you are happy.
Might want to bump the version and relase

simeonschaub:
Ah, yes. Should probably tag a release since otherwise this always causes precompile to fail in 1.7.

PR #423, opened 2021-06-01
Delete rules and tests for `NaNMath` and `SpecialFunctions`

### LINKED ISSUE #142: Add tests for new special functions and nan math
New rules are introduced in https://github.com/JuliaDiff/ChainRules.jl/pull/80

### LINKED ISSUE #206: Support for svdvals
We already have support for this via `svd(X).S`, but support for `svdvals` directly would be nice. It'd also be great to support GenericLinearAlgebra, which I don't think we do at the moment.

### LINKED ISSUE #280: Glue packages with working Pkg compat versioning
This is the outcome of a discussion I had with @KristofferC  about how to make glue packages work.

We should put it into practice for SpecialFunctions.jl and NaNMath.jl

We make a new package called:
**ChainRules_SpecialFunctions_Glue.jl**
which has the released with compat on **SpecialFunctions.jl** declared
then in **ChainRules.jl** we would declare a dependency NOT on SpecialFunctions
but on **ChainRules_SpecialFunctions_Glue.jl**
and then in the source code of **ChainRules.jl** would do:

```julia
    @require SpecialFunctions="276daf66-3868-5448-9aa4-cd146d93841b" begin
        using ChainRules_SpecialFunctions_Glue  # maybe use `Base.require` instead
    end
```

So **ChainRules.jl** never has a direct dependency on **SpecialFunctions.jl**
only on **ChainRules_SpecialFunctions_Glue.jl**.
But ChainRules.jl *does*  use Require to conditionally load  **ChainRules_SpecialFunctions_Glue.jl** in response to **SpecialFunction.jl** being loaded.

and  **ChainRules_SpecialFunctions_Glue.jl** *does* have a direct dependency on **SpecialFunctions.jl** (and on **ChainRulesCore.jl**), and it declares constraints on it
so the pakcage manager can workout which should be loaded.

In pacticular,  **ChainRules_SpecialFunctions_Glue.jl** would not support both **SpecialFunctions.jl** v0.7 and v0.8 at the same time.
Rather **ChainRules_SpecialFunctions_Glue.jl** v1.0.0 would support **SpecialFunctions.jl** v0.7 
and **ChainRules_SpecialFunctions_Glue.jl** v1.1.0 would support **SpecialFunctions.jl** v0.8

And **ChainRules.jl** would support **ChainRules_SpecialFunctions_Glue.jl** `1.x.y`

TL; DR:
you use Requires on the Other package, to load the Glue
and you put the Glue package in your dependencies

### LINKED ISSUE #292: add derivatives for exponential integrals
Just added to SpecialFunctions.jl: https://github.com/JuliaMath/SpecialFunctions.jl/pull/236

The derivative with respect to `x` is a simple recurrence: https://en.wikipedia.org/wiki/Exponential_integral#Derivatives

### LINKED ISSUE #319: Drop SpecialFunctions dependency

### LINKED ISSUE #402: Numerical precision of primal not respected for SpecialFunctions.jl
Several methods from SpecialFunctions.jl are not respecting the numerical precision of the primal:
```julia
julia> using ChainRules, SpecialFunctions

julia> y, ȳ = ChainRules.rrule(SpecialFunctions.erf, 1.0f0)
(0.8427008f0, SpecialFunctions.var"#erf_pullback#12"{Float32}(1.0f0))

julia> ȳ(1.0f0)[2]
0.41510750774498784
```

I'm guessing this is because the rules were originally copy-pasted from DiffRules.jl which suffers from the same issue: https://github.com/JuliaDiff/DiffRules.jl/pull/55.

I think I'm currently of the opinion that the best way to fix this is to copy-paste the constants from e.g. StatsFuns.jl to remove most of the promotions, and then I guess we'll have to use `oftype` for the remainders.

Is there a better solution?

Btw, I know this related to other issues, e.g. #232, but to me it seems like this particular case is non-ambiguous and can be easily fixed so might as well do it asap:)

mzgubic:
Closes #319, closes #142 , closes #280, closes #402, relevant for  #206 and #292 

To be merged after https://github.com/JuliaDiff/ChainRules.jl/pull/417 in the same breaking release. (this PR branches off those changes by mistake)


oxinabox:
Can you rebase this to not include those other changes?
(if you need help doing so, ping me on slack)

PR #427, opened 2021-06-01
Cancel previous running workflows

mzgubic:
I don't know anything about CI, so this might be very wrong - just testing it out

### DIFF
--- a/.github/workflows/Cancel.yml
+++ b/.github/workflows/Cancel.yml
@@ -9,5 +9,5 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.9.0
         with:
-          workflow_id: "CI", "IntegrationTest"
+          workflow_id: 'CI.yml', 'IntegrationTest.yml'
           access_token: ${{ github.token }}
--- a/.github/workflows/Cancel.yml
+++ b/.github/workflows/Cancel.yml
@@ -1,5 +1,6 @@
 name: Cancel
 on: [push]
+
 jobs:
   cancel:
     name: 'Cancel Previous Runs'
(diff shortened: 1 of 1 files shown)


devmotion:
I haven't used it anywhere (just opened a PR to KernelFunctions where we had an open issue about canceling workflows) but I think it could be helpful to configuration in https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks such that it also works for PRs from forks.

mzgubic:
Could we disable email notification when the workflow is cancelled rather than when it fails?

.github/workflows/Cancel.yml
```diff
@@ -0,0 +1,17 @@
...
+       - "JuliaNightly"
+     types:
+       - requested
+ jobs:
+   cancel:
+     runs-on: ubuntu-latest
+     steps:
+     - uses: styfle/cancel-workflow-action@0.9.0
+       with:
+         all_but_latest: true
```
mzgubic: By default, only the workflows started before this action are cancelled. This flag cancels all but the latest workflow, which is helpful if the pipeline is saturated

oxinabox:
Hmm this seems to have resulted in everything being cancelled because of VersionVigilante failing.
We don't want everything cancelled when anything fails, since Nightly fails all the time,

mzgubic:
Oh no I cancelled manually because I didn't want to block other pipelines, sorry!

oxinabox:
OK, lets give this a try

oxinabox (approved):

mzgubic:
Did not seem to have worked, i.e. this PR still ran two full pipelines:
https://github.com/JuliaDiff/ChainRules.jl/pull/433

PR #429, opened 2021-06-02
stop CI testing every  OS

oxinabox:
This package is pretty much OS agnostic.
I say we disable them until someone opens an issue saying something doesn't work on Windows etc.

A possibly alternative to this PR.
Would be to mix our matrix over the "diagonal".
So e.g.
 - Julia 1.0, Mac, 64 bit
 - Julia 1.0, Ubuntu, 32 bit
 - Julia 1.x, Windows, 32 bit
 gives use coverage of all OS and all julia version and all archetectures.
 
 And maybe for bit more completeness:
 - Julia 1.x, Ununtu, 64 bit
 
But writing GHA for that is a bit annoying, so could be a follow up PR

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,15 +19,9 @@ jobs:
           - "1"    # Latest Release
         os:
           - ubuntu-latest
-          - macOS-latest
-          - windows-latest
         arch:
           - x86
           - x64
-        exclude:
-          # 32-bit Julia binaries are not available on macOS
-          - os: macOS-latest
-            arch: x86
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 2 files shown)


simeonschaub:
We do have some nondiff rules for IO stuff, which sometimes differ in signatures between different platforms. Although I don't think we actually test them currently, but it would catch precompile failures at least.

oxinabox:
> We do have some nondiff rules for IO stuff, which sometimes differ in signatures between different platforms. Although I don't think we actually test them currently, but it would catch precompile failures at least.

Yeah, I did consider those.
But I don't think they are worth having it.
I am happy to be proven wrong, but until I am proven wrong i want to disable the CI

simeonschaub:
Yeah, until we have proper tests for those, it's probably ok to remove the jobs.

simeonschaub:
The JuliaNightly jobs should probably be adapted as well

nickrobinson251 (approved):
let's try it!

mzgubic (approved):
sounds good

PR #432, opened 2021-06-02
RFC: use multiple threads for tests

simeonschaub:
Not sure this is the right way to do this and how much this will end up
screwing with the test output, but let's just see for now whether this
actually makes CI finish faster.


oxinabox:
Last i checked (which was quite a while ago) the Test standard library isn't threadsafe, or even async-safe.
because testset interact with there parent testset, when they start and finish.
(I can't remember if it was in task-local or outright global last i checked. I know now it is in task local, which might have fixed things)

PR #433, opened 2021-06-02
remove `Requires` dependency and some `NaNMath`/`SpecialFunctions` artifacts

mzgubic:
Clean up after https://github.com/JuliaDiff/ChainRules.jl/pull/425, not sure what went wrong


oxinabox (approved):
my bad for missing these before

PR #434, opened 2021-06-02
fix lu deprecation warnings on Julia nightly

simeonschaub:
Ref https://github.com/JuliaLang/julia/pull/40623

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -15,14 +15,14 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
 # for derivations for wide and tall matrices, see
 # https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/
 
-const _RowMaximum = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum : Val{true}
-const _NoPivot = VERSION >= v"1.7.0-DEV.1188" ? NoPivot : Val{false}
+const LU_RowMaximum = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum : Val{true}
+const LU_NoPivot = VERSION >= v"1.7.0-DEV.1188" ? NoPivot : Val{false}
 
 function frule(
-    (_, ΔA), ::typeof(lu!), A::StridedMatrix, pivot::Union{_RowMaximum,_NoPivot}; kwargs...
+    (_, ΔA), ::typeof(lu!), A::StridedMatrix, pivot::Union{LU_RowMaximum,LU_NoPivot}; kwargs...
 )
     F = lu!(A, pivot; kwargs...)
-    ∂factors = pivot === _RowMaximum() ? ΔA[F.p, :] : ΔA
+    ∂factors = pivot === LU_RowMaximum() ? ΔA[F.p, :] : ΔA
     m, n = size(∂factors)
     q = min(m, n)
     if m == n  # square A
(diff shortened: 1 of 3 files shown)


oxinabox:
Should we put these in Compat.jl instead?

simeonschaub:
I thought about that, but wasn't sure defining `Compat.RowMaximum` and `Compat.NoPivot` as just `Val{true}`/`Val{false}` was the right thing to do for `Compat.jl`, or whether it should define them as separate singletons and overload `lu`.

oxinabox:
I don't think we are in any rush to fix this.
So lets open a PR on Compat.jl and see what the Compat.jl maintainers think?

simeonschaub:
Ah, looking at the Julia PR again it seems like `Val{true}` only means `RowMaximum` for `lu`, while for `qr` `Val{true}` means `ColumnNorm`. This likely means we don't want to implement them as aliases for `Val{true}` in Compat, because `lu(x, ColumnNorm())` would return `lu(x, RowMaximum())` without any warning, which seems bad. I can open a PR defining these as singletons in Compat instead, but that would be independent of this PR, since we do actually need to overload `lu(x, ::Val)` on earlier Julia versions.

oxinabox:
> I can open a PR defining these as singletons in Compat instead, but that would be independent of this PR, since we do actually need to overload lu(x, ::Val) on earlier Julia versions.

let's do this

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -15,11 +15,14 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
 # for derivations for wide and tall matrices, see
 # https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/
 
+const _RowMaximum = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum : Val{true}
+const _NoPivot = VERSION >= v"1.7.0-DEV.1188" ? NoPivot : Val{false}
```
oxinabox, line 19: Do we need to call this LU_NoPivot
since we will want this to differ for QR?

simeonschaub: Yeah, makes sense

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -15,11 +15,14 @@ using LinearAlgebra.BLAS: gemv, gemv!, gemm!, trsm!, axpy!, ger!
...
+const LU_RowMaximum = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum : Val{true}
+const LU_NoPivot = VERSION >= v"1.7.0-DEV.1188" ? NoPivot : Val{false}
+
 function frule(
-    (_, ΔA), ::typeof(lu!), A::StridedMatrix, pivot::Union{Val{false},Val{true}}; kwargs...
+    (_, ΔA), ::typeof(lu!), A::StridedMatrix, pivot::Union{LU_RowMaximum,LU_NoPivot}; kwargs...
 )
     F = lu!(A, pivot; kwargs...)
-    ∂factors = pivot === Val(true) ? ΔA[F.p, :] : ΔA
+    ∂factors = pivot === LU_RowMaximum() ? ΔA[F.p, :] : ΔA
```
oxinabox, line 25:
```suggestion
    ∂factors = pivot isa LU_RowMaximum ? ΔA[F.p, :] : ΔA
```

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -124,7 +127,7 @@ function rrule(
             ldiv!(L1', ∂A1)
             rdiv!(∂A, U')
         end
-        if pivot === Val(true)
+        if pivot === LU_RowMaximum()
```
oxinabox, line 130:
```suggestion
        if pivot isa LU_RowMaximum
```

oxinabox (approved):
merge when you are happy

PR #436, opened 2021-06-04
Return the correct number of cotangents for functions with optional arguments

mzgubic:
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/168 (merged, and then reverted) introduced checks that the number of cotangents from the rrule is correct.

Possible fix for https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/168#issuecomment-854731827

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -196,8 +196,19 @@ end
 
 function rrule(
     ::typeof(pinv),
+    x::Union{AbstractVector{T}, LinearAlgebra.AdjOrTransAbsVec{T}},
+) where {T<:Union{Real,Complex}}
+    y, full_pb = rrule(pinv, x, 0)
+    function pinv_pullback(Δy)
+        return pull_pb(Δy)[1:2]
+    end
+    return y, pinv_pullback
+end
+
+function rrule(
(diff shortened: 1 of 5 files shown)


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -194,10 +194,21 @@ function frule((_, ΔA), ::typeof(pinv), A::AbstractMatrix{T}; kwargs...) where
...
 end
 
+function rrule(
+    ::typeof(pinv),
+    x::Union{AbstractVector{T}, LinearAlgebra.AdjOrTransAbsVec{T}},
+) where {T<:Union{Real,Complex}}
+    y, full_pb = rrule(pinv, x, 0)
+    function pinv_pullback(Δy)
+        return full_pb(Δy)[1:2]
+    end
```
oxinabox, line 204: I am trying to remeber if there is a reason we don't do this:
```suggestion
    pinv_pullback(Δy) = full_pb(Δy)[1:2]
```

oxinabox (approved):

mzgubic:
It looks like inference works on Julia 1.6 and 1.5, but does not work on 1.3 (I didn't check 1.4).

Are we ok with that?

PR #439, opened 2021-06-08
fix some numerical instabilities in Base tests

simeonschaub:
These were failing on nightly because of RNG changes. Take greater care
not to sample too close to singularities in functions like `mod`.

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -1,3 +1,9 @@
+# be careful not to sample near singularities for jumping functions like `mod`
+function rand_mod_pair()
+    base = rand() + 1
+    return (rand(0:10) + .6rand() + .2) * base, base
+end
+
 @testset "base" begin
     @testset "Trig" begin
         @testset "Basics" for x = (Float64(π)-0.01, Complex(π, π/2))
@@ -100,8 +106,8 @@
     end
 
     @testset "mod" begin
-        test_frule(mod, 10rand(), rand())
-        test_rrule(mod, 10rand(), rand())
+        test_frule(mod, rand_mod_pair()...)
+        test_rrule(mod, rand_mod_pair()...)
     end
 
     @testset "identity" for T in (Float64, ComplexF64)
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -1,9 +1,3 @@
-# be careful not to sample near singularities for jumping functions like `mod`
-function rand_mod_pair()
-    base = rand() + 1
-    return (rand(0:10) + .6rand() + .2) * base, base
-end
-
 @testset "base" begin
     @testset "Trig" begin
         @testset "Basics" for x = (Float64(π)-0.01, Complex(π, π/2))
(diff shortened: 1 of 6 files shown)


test/rulesets/Base/fastmath_able.jl
```diff
@@ -125,8 +125,8 @@ const FASTABLE_AST = quote
 
     @testset "binary functions" begin
         @testset "$f(x, y)" for f in (atan, rem, max, min)
-            test_frule(f, 100rand(), 10rand())
-            test_rrule(f, 100rand(), 10rand())
+            test_frule(mod, rand_mod_pair()...)
+            test_rrule(mod, rand_mod_pair()...)
         end
```
nickrobinson251, line 130: Typo: `f` not `mod` 

(Not sure why GitHub won't let me add suggestion)

test/rulesets/Base/fastmath_able.jl
```diff
@@ -125,8 +125,8 @@ const FASTABLE_AST = quote
 
     @testset "binary functions" begin
         @testset "$f(x, y)" for f in (atan, rem, max, min)
-            test_frule(f, 100rand(), 10rand())
-            test_rrule(f, 100rand(), 10rand())
+            test_frule(mod, rand_mod_pair()...)
+            test_rrule(mod, rand_mod_pair()...)
```
simeonschaub, line 129:
```suggestion
            test_frule(f, rand_mod_pair()...)
            test_rrule(f, rand_mod_pair()...)
```

test/rulesets/Base/fastmath_able.jl
simeonschaub: Oops, bad copy-paste!

simeonschaub:
Can I get a review? Nightly still fails of course, but the tests in questions seem to have been fixed at least.

test/rulesets/Base/base.jl
```diff
@@ -100,8 +106,8 @@
     end
 
     @testset "mod" begin
-        test_frule(mod, 10rand(), rand())
-        test_rrule(mod, 10rand(), rand())
+        test_frule(mod, rand_mod_pair()...)
+        test_rrule(mod, rand_mod_pair()...)
```
oxinabox, line 110: I would rather hard code this with literals.
Tests are clearer when we use literals.
A lot of are scalar tests have already been rewritten to use literals.

and in this case the mode code adds much complexity for little gain.

test/rulesets/Base/fastmath_able.jl
```diff
@@ -125,8 +125,8 @@ const FASTABLE_AST = quote
 
     @testset "binary functions" begin
         @testset "$f(x, y)" for f in (atan, rem, max, min)
-            test_frule(f, 100rand(), 10rand())
-            test_rrule(f, 100rand(), 10rand())
+            test_frule(f, rand_mod_pair()...)
+            test_rrule(f, rand_mod_pair()...)
```
oxinabox, line 129: Similar here
I probably should have done that, rather than adding the 100, 10 thing

test/rulesets/Base/base.jl
simeonschaub: Ok, I can change that

oxinabox (approved):
merge when happy

PR #441, opened 2021-06-08
Rule for sum(f, xs)

oxinabox:
This is the real use of https://github.com/JuliaDiff/ChainRulesCore.jl/pull/363
needs tht to be merged first.
Soon I will make a Zygote PR that will to hit it.

Need to workout testing this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docs/build
 docs/site
 .idea/*
 dev/*
+.vscode/settings.json
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.10.4"
-ChainRulesTestUtils = "0.7.6"
+ChainRulesTestUtils = "0.7.8"
 Compat = "3.30"
 FiniteDifferences = "0.12.8"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 ChainRulesCore = "0.10.4"
-ChainRulesTestUtils = "0.7.8"
+ChainRulesTestUtils = "0.7.9"
 Compat = "3.30"
 FiniteDifferences = "0.12.8"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -14,13 +14,15 @@ ChainRulesCore = "0.10.4"
 ChainRulesTestUtils = "0.7.9"
 Compat = "3.30"
 FiniteDifferences = "0.12.8"
+StaticArrays = "1.2"
 julia = "1"
 
 [extras]
 ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["ChainRulesTestUtils", "FiniteDifferences", "Random", "Test"]
+test = ["ChainRulesTestUtils", "FiniteDifferences", "Random", "StaticArrays", "Test"]
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.10"
+ChainRulesCore = "0.10.4"
 ChainRulesTestUtils = "0.7.6"
 Compat = "3.30"
 FiniteDifferences = "0.12.8"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.11"
+version = "0.8.12"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -19,6 +19,27 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
     return y, sum_pullback
 end
 
+# Can't map over Adjoint/Transpose Vector
+function rrule(
+    config::RuleConfig{>:HasReverseMode},
+    ::typeof(sum),
+    f,
+    xs::Union{Adjoint{<:Number,<:AbstractVector},Transpose{<:Number,<:AbstractVector}};
+    kwargs...
+)
+    op = xs isa Adjoint ? adjoint : transpose
+    # since summing a vector we don't need to worry about dims which simplifies adjointing
+    vector = parent(xs)
+    y, vector_sum_pb = rrule(config, sum, f, vector; kwargs...)
+    function covector_sum_pb(ȳ)
+        s̄um, f̄, v̄ = vector_sum_pb(ȳ)
+        return s̄um, f̄, op(v̄)
+    end
+
+    return y, covector_sum_pb
+end
+
+
 function rrule(
     config::RuleConfig{>:HasReverseMode}, ::typeof(sum), f, xs::AbstractArray; dims=:
 )
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -19,6 +19,26 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
     return y, sum_pullback
 end
 
+function rrule(
+    config::RuleConfig{>:CanReverseMode}, ::typeof(sum), f, xs::AbstractArray{T}; dims=:
+) where {T<:Number}
+    fx_and_pullbacks = map(x->rrule_via_ad(config, f, x), xs)
+    y = sum(first, fx_and_pullbacks; dims=dims)
+
+    pullbacks = last.(fx_and_pullbacks)
+    function sum_pullback(ȳ)
+        f̄_and_x̄s = [pullback(ȳ) for pullback in pullbacks]
+        # no point thunking as most of work is in f̄_and_x̄s which we need to compute for both
+        f̄ = if fieldcount(typeof(f)) === 0 # Then don't need to worry about derivative wrt f
+            NoTangent()
+        else
+            sum(first, f̄_and_x̄s)
+        end
+        x̄s = sum(last, f̄_and_x̄s)
+        return NoTangent(), f̄, x̄s
+    end
+end
+
 function frule(
     (_, _, ẋ),
     ::typeof(sum),
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -20,23 +20,24 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
 end
 
 function rrule(
-    config::RuleConfig{>:CanReverseMode}, ::typeof(sum), f, xs::AbstractArray{T}; dims=:
+    config::RuleConfig{>:HasReverseMode}, ::typeof(sum), f, xs::AbstractArray{T}; dims=:
 ) where {T<:Number}
     fx_and_pullbacks = map(x->rrule_via_ad(config, f, x), xs)
     y = sum(first, fx_and_pullbacks; dims=dims)
 
     pullbacks = last.(fx_and_pullbacks)
     function sum_pullback(ȳ)
-        f̄_and_x̄s = [pullback(ȳ) for pullback in pullbacks]
+        @show f̄_and_x̄s = [pullback(ȳ) for pullback in pullbacks]
         # no point thunking as most of work is in f̄_and_x̄s which we need to compute for both
         f̄ = if fieldcount(typeof(f)) === 0 # Then don't need to worry about derivative wrt f
             NoTangent()
         else
             sum(first, f̄_and_x̄s)
         end
-        x̄s = sum(last, f̄_and_x̄s)
+        x̄s = map(last, f̄_and_x̄s)
         return NoTangent(), f̄, x̄s
     end
+    return y, sum_pullback
 end
 
 function frule(
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -20,8 +20,8 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
 end
 
 function rrule(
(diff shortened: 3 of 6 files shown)


mzgubic:
Is there an issue with defining `frule_via_ad(args...; kwargs...) = frule(args...; kwargs)` for testing?

This needs defining the `f/rrule`s for the `f` being passed to `sum`, but that's not too bad, right?

test/rulesets/Base/mapreduce.jl
```diff
@@ -1,3 +1,30 @@
+"For testing this config re-dispatches Xrule_via_ad to Xrule without config argument"
```
mzgubic, line 1: Why? We want to preserve the config, right? And let it hit the fallback in CRC?

oxinabox: fair, yes, that would be better

src/rulesets/Base/mapreduce.jl
```diff
@@ -19,6 +19,27 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
     return y, sum_pullback
 end
 
+function rrule(
+    config::RuleConfig{>:HasReverseMode}, ::typeof(sum), f, xs::AbstractArray{T}; dims=:
```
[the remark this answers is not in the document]
oxinabox: idk, I also want to define thios on iterators more generally even.

oxinabox:
This is ready to go, once it's downstream
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/166
and
https://github.com/JuliaDiff/ChainRulesCore.jl/pull/363

are done

src/rulesets/Base/mapreduce.jl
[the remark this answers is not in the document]
oxinabox: Also I want to delete the old ones from Zygote, and Zygote is always super-general

test/runtests.jl
```diff
@@ -22,6 +22,8 @@ end
 
 println("Testing ChainRules.jl")
 @testset "ChainRules" begin
+    include_test("test_helpers.jl")
+    println()
```
mzgubic, line 26:
```suggestion
```

mzgubic (approved):
Needs compat change

test/runtests.jl
oxinabox: This wa intentional. See the other uses of `println()` below

mzgubic: yeah do we want to rethink that since we add `println` in `include_test`? It doesn't really matter

oxinabox: out of scope for this PR

Project.toml
```diff
@@ -10,7 +10,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.10"
+ChainRulesCore = "0.10.4"
 ChainRulesTestUtils = "0.7.6"
```
oxinabox, line 14:
```suggestion
ChainRulesTestUtils = "0.7.8"
```

mzgubic:
This is going to need CRTU 0.7.9, https://github.com/JuliaRegistries/General/pull/38672

Project.toml
```diff
@@ -10,8 +10,8 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "0.10"
-ChainRulesTestUtils = "0.7.6"
+ChainRulesCore = "0.10.4"
+ChainRulesTestUtils = "0.7.8"
```
mzgubic, line 14:
```suggestion
ChainRulesTestUtils = "0.7.9"
```

src/rulesets/Base/mapreduce.jl
```diff
@@ -19,6 +19,27 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
...
 
+function rrule(
+    config::RuleConfig{>:HasReverseMode}, ::typeof(sum), f, xs::AbstractArray; dims=:
+)
+    fx_and_pullbacks = map(x->rrule_via_ad(config, f, x), xs)
+    y = sum(first, fx_and_pullbacks; dims=dims)
+
+    pullbacks = last.(fx_and_pullbacks)
+    function sum_pullback(ȳ)
+        f̄_and_x̄s = [pullback(ȳ) for pullback in pullbacks]
```
mcabbott, line 30: Have you investigated the performance of this? Storing an array of the need not be slow, but I wonder how well it works in practice?

I also wonder whether you could avoid making an array of tuples,  before `sum(first, fx_and_pullbacks)` and `last.(fx_and_pullbacks)` separates them. 

Finally, have you given thought to weird arrays, like `SMatrix`, or `CuArray`s? For the former at least, perhaps using `map` instead of a generator for `f̄_and_x̄s` may better preserve the structure.

oxinabox: > Have you investigated the performance of this? Storing an array of the need not be slow, but I wonder how well it works in practice?

I have not. This is a first PR to get this feature out the door.
I am only really interested in it working, and having something to test AD's against.

For interest, Zygote's approach is
https://github.com/FluxML/Zygote.jl/blob/12f5c1d75eeaa8c7a818f2db7f8d082956c00cac/src/lib/array.jl#L296
which changes the primal to a `sum(f.x)` then AD's that expression.
Which hits https://github.com/FluxML/Zygote.jl/blob/12f5c1d75eeaa8c7a818f2db7f8d082956c00cac/src/lib/broadcast.jl#L172-L182
which is strictly worse than this, since it makes a temporary array for all the `y`s (which we just sum out), and for all the pullbacks

I will time them soon and post back, it will be interesting.

> I also wonder whether you could avoid making an array of tuples, before `sum(first, fx_and_pullbacks)` and `last.(fx_and_pullbacks)` separates them.

Yeah, I would like to look into that.
It's why i am doing `sum` rather than for `map` because I think for `map` that would be even more important.
It's a fiddly little pattern to write to make sure everything gets the right types, so I didn't want to do it for the first PR.

> Finally, have you given thought to weird arrays, like SMatrix, or CuArrays? For the former at least, perhaps using map instead of a generator for f̄_and_x̄s may better preserve the structure.

Good point, that's worth testing.

mcabbott: Without AD, the reason to write `sum(f, x)` instead of `sum(f.(x))` is precisely to save allocations. It would be nice if the AD could preserve that, although possibly tricky.

I guess the size of the array of closures will depend on how much they capture, which may include both `x` and `y = f.(x)`. Ideally it would never include `x` since that already has its own array... and sometimes it would be quicker to re-calculate `y` but that seems even harder to arrange.

oxinabox:
Here are benchmarks
Script:
```julia
using Zygote
using BenchmarkTools

const xs = randn(10_000)
@btime Zygote.pullback(sum, abs, $xs)[2](1);
```

machine details
```julia
julia> versioninfo()
Julia Version 1.6.2-pre.0
Commit dd122918ce* (2021-04-23 21:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin20.3.0)
  CPU: Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

### Zygote v0.6.12
```
  515.631 μs (30097 allocations: 1.07 MiB)
  ```
  
### [Zygote#ox/ruleconfig](https://github.com/FluxML/Zygote.jl/pull/990)
using this PR as of https://github.com/JuliaDiff/ChainRules.jl/pull/441/commits/4207633347b51b1a7913c0dba23578a11970de4e
```
  55.476 μs (17 allocations: 703.61 KiB)
```

---

So while I don't think this code is the fastest code that can do this.
I don't this we need to worry about making things worst.
Zygote is just surfiently terrible right now

**Comment by mcabbott** [comment]
FWIW, some other ways (although on a different machine!)
```
julia> @btime Zygote.pullback(sum, abs, $xs)[2](1);
  479.167 μs (30098 allocations: 1.07 MiB)

julia> using Tullio, ForwardDiff

julia> @btime Zygote.pullback(xs -> (@tullio _ := abs(xs[i])  avx=false threads=false), $xs)[2](1);
  27.375 μs (2 allocations: 78.20 KiB)

julia> @btime Zygote.pullback(xs -> (@tullio _ := abs(xs[i])  avx=false threads=false grad=Dual), $xs)[2](1);
  6.492 μs (2 allocations: 78.20 KiB)  # evaluates f twice, the second time with Dual(x,1) for gradient

julia> @btime copy($xs); # just to see alloc
  2.708 μs (2 allocations: 78.20 KiB)
```

**Comment by oxinabox** [comment]
Yeah, I agree faster things are possible.
Can be follow up PRs, as long as we don't make Zygote worse, I am happy with this.

Cool (and I guess not surprising) that reverse via forward is fastest here.
Which is something we now support writing using RuleConfigs.

File: src/rulesets/Base/mapreduce.jl [hunk 7]
```diff
@@ -56,6 +77,17 @@ function rrule(
...
+# Fix dispatch for this pidgeon-hole optimization,
+# Rules with RuleConfig dispatch with priority over without (regardless of other args).
+# and if we don't specify what do do for one that HasReverseMode then it is ambigious
+for Config in (RuleConfig, RuleConfig{>:HasReverseMode})
+    @eval function rrule(
+        ::$Config, ::typeof(sum), ::typeof(abs2), x::AbstractArray{T}; dims=:,
+    ) where {T<:Union{Real,Complex}}
+        return rrule(sum, abs2, x; dims=dims)
+    end
+end
```
oxinabox, line 89: I don't love this.
But I think it is as good as it gets.
We could require every rule is written to have to have a `RuleConfig`,
but we would still get the ambiguity, and so would still need the same code (just with one less thing in the top loop.

If we are happy with this I will make a follow up PR to the ChainRulesCore docs, warning about this.

mzgubic: Eww. But I can't think of anything nicer either. Can we add tests for this as well?

mzgubic (approved):

src/rulesets/Base/mapreduce.jl
mcabbott: Why would `
rrule(config::RuleConfig{&gt;:HasReverseMode}, ::typeof(sum), f::typeof(abs2), xs::AbstractArray)` not be more specific?

oxinabox: It is, which is why we have to generate that.

The problem is with:
```
rrule(config::RuleConfig, ::typeof(sum), f::typeof(abs2), xs::AbstractArray)
```
which is what we could make the required format.
That one is ambiguous with
```
rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(sum), f, xs::AbstractArray)
```

src/rulesets/Base/mapreduce.jl
oxinabox: I did a little experiment into using `mapreduce((net, pullbacks), (val, pb)) = (net + val, push!(pullbacks, pb))`
and I didn't get any performance improvement. Infact I got a large regression, but that could have been me messing stuff up.
I would like to do more testing of this.

mzgubic (approved):
just a couple of minor things, lgtm otherwise, feel free to merge when you want

.gitignore
```diff
@@ -7,3 +7,4 @@ docs/build
 docs/site
 .idea/*
 dev/*
+.vscode/settings.json
```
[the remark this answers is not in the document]
oxinabox: Yes, it's not related to this PR though.

It's like how we have the `idea` editor config ignored

src/rulesets/Base/mapreduce.jl
```diff
@@ -19,6 +19,50 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
...
+
+    function sum_pullback(ȳ)
+        call(f, x) = f(x)  # we need to broadcast this to handle dims kwarg
+        f̄_and_x̄s = call.(pullbacks, ȳ)
+        # no point thunking as most of work is in f̄_and_x̄s which we need to compute for both
+        f̄ = if fieldcount(typeof(f)) === 0 # Then don't need to worry about derivative wrt f
+            NoTangent()
+        else
+            sum(first, f̄_and_x̄s)
+        end
```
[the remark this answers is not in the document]
oxinabox: Yeah I thought that so I timed it.
It's 20% slower.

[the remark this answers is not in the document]
mzgubic: oh wow, I did not expect that much slower! What if we define `sum(::Array{AbstractZero}) = ZeroTangent()` or something similar?

src/rulesets/Base/mapreduce.jl
```diff
@@ -19,6 +19,50 @@ function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
...
+    fx_and_pullbacks = map(x->rrule_via_ad(config, f, x), xs)
+    y = sum(first, fx_and_pullbacks; dims=dims)
+
+    pullbacks = last.(fx_and_pullbacks)
+
+    function sum_pullback(ȳ)
+        call(f, x) = f(x)  # we need to broadcast this to handle dims kwarg
+        f̄_and_x̄s = call.(pullbacks, ȳ)
+        # no point thunking as most of work is in f̄_and_x̄s which we need to compute for both
+        f̄ = if fieldcount(typeof(f)) === 0 # Then don't need to worry about derivative wrt f
```
CarloLucibello, line 55: Not necessarily related to this PR, but it would be good to define a function for `fieldcount(typeof(f)) === 0`, e.g. `hasstructure(f)` or `iscomposite(f)`

simeonschaub: I am not sure that is necessarily the right check though. There could be number-like or array-like functors which do have a well-defined derivative wrt `f`, but don't have any fields. Probably not a big deal for now, but that might be something to keep in mind for general design decisions.

oxinabox: true, they are very unusual though. Probably a good thing to worry about when defining @CarloLucibello 's `iscomposite`/`hasstructure` or what ever we call it, maybe `hastangent`?

Cases I can think of are functors that are also:
 -  number types defined using `primative`,
  - Something that is like `FillArrays.One()` or `OneHot` that pushes  size and index into the type-level.

src/rulesets/Base/mapreduce.jl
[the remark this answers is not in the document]
oxinabox: we would need more like `sum(::typeof(first), ::Array{Tuple{T, Any}) where T isa AbstractZero = T()`
which seems more involved than i want in my life.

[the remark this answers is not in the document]
mzgubic: ah yeah, fair enough

src/rulesets/Base/mapreduce.jl
mcabbott: @tkf's suggestion was to use `Base.issingletontype`, which I think would avoid your `primitive` example. Not sure there's a One, but other cases with a value in the type are regarded as constants:
```
julia> FillArrays.Ones(3) |> typeof |> fieldnames
(:axes,)

julia> FillArrays.One
ERROR: UndefVarError: One not defined

julia> get2(::Val{x}) where x = x^2;

julia> gradient(get2∘Val, 3.14)
(nothing,)

julia> gradient(get2, Val(3.14))
(nothing,)
```

mcabbott: Xref weird example (from the tests) here: https://github.com/FluxML/Zygote.jl/pull/1001#discussion_r654866919

tl;dr is that `global s += 1` is not detected by these tests. That's a test of order of iteration.

Issue #422, opened 2021-05-28
eigen tests fail on nightly


simeonschaub:
See for example https://github.com/JuliaDiff/ChainRules.jl/pull/421/checks?check_run_id=2697576261:

```
Test Summary:                                            |  Pass  Fail  Broken  Total
ChainRules                                               | 23210     3       4  23217
  rulesets                                               | 23210     3       4  23217
    Base                                                 |  6510             4   6514
    Statistics                                           |    12                   12
    LinearAlgebra                                        | 16643     3          16646
      dense                                              |   585                  585
      norm functions                                     |  1592                 1592
      normalize                                          |    90                   90
      matrix functions                                   |   133                  133
      Structured Matrices                                |   412                  412
      Symmetric/Hermitian rules                          |  3556                 3556
      Factorizations                                     |   505     3            508
        lu decomposition                                 |   159                  159
        svd                                              |   226                  226
        eigendecomposition                               |   104     3            107
          eigen/eigen!                                   |    67     3             70
            eigen!(::Matrix{Float64}) frule              |     7                    7
            eigen!(::Matrix{ComplexF64}) frule           |     4     2              6
              tangents are real when outputs are         |     1                    1
            eigen(::Matrix{Float64}) rrule               |     8                    8
            eigen(::Matrix{ComplexF64}) rrule            |     6     1              7
            normalization/phase functions are idempotent |     4                    4
            normalization/phase functions are idempotent |     4                    4
            hermitian matrices                           |    34                   34
          eigvals/eigvals!                               |    37                   37
        cholesky                                         |    16                   16
      BLAS                                               |  9354                 9354
      LAPACK                                             |   416                  416
    Random                                               |    45                   45
    packages                                             |                      No tests
ERROR: LoadError: Some tests did not pass: 23210 passed, 3 failed, 0 errored, 4 broken.
```

simeonschaub:
Will likely take a while for nightlies to propagate through, but then we should hopefully have passing tests on nightly again.

PR #445, opened 2021-06-15
fix #379: numeric instabilities in eigen tests

### LINKED ISSUE #379: stability of primal when testing eigen
See https://github.com/JuliaDiff/ChainRules.jl/pull/375#discussion_r577908632_

In short, tests started failing with huge values like 10^5 in the output of FDM.
Fixing the seed fixes it, but replacing `randn(T, n, n)` with `randn(T, n, n,) .+ 5.0`  seemed to fix it robustly.

@willtebbutt said:
> My suspicion would be that the eigenvalues are flipping around or something?

@sethaxen said
> It could be, but the major source of instability I saw was in the phase of the eigenvectors. The convention LinearAlgebra uses (from LAPACK) is that the entry with the largest magnitude in each eigenvector is rotated to be positive real or negative real, whichever is nearest. So it could become unstable if another entry was close to that one in magnitude. So one way might be to do something like this:
> ```julia
> V = hcat(Random.shuffle([normalize((x .+ 1e-2 .* randn.(ComplexF64)) * sign(randn(ComplexF64))) for x in eachcol(Matrix{Float64}(I, n, n))])...)
> ```
> which creates an identity matrix, adds small random complex numbers to all entries, applies a random phase convention to each eigenvector, then normalizes them. The eigenvectors are then shuffled and made columns of a new matrix. This produces one where the element with the largest magnitude is well-separated from the rest. With a suitable set of eigenvalues, perhaps this would make a well-behaved primal.

simeonschaub:
This was causing failures on nightly again. From running these tests a
bunch of times, there is still an approximately 3% chance that they will
fail randomly, but I think that's probably acceptable.

### DIFF
--- a/test/rulesets/LinearAlgebra/factorization.jl
+++ b/test/rulesets/LinearAlgebra/factorization.jl
@@ -23,6 +23,25 @@ end
 const LU_ROW_MAXIMUM = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum() : Val(true)
 const LU_NO_PIVOT = VERSION >= v"1.7.0-DEV.1188" ? NoPivot() : Val(false)
 
+# well-conditioned random n×n matrix with elements of type `T` for testing `eigen`
+function rand_eigen(T::Type, n::Int)
+    # uniform distribution over `(-1, 1)` / `(-1, 1)^2`
+    _rand(T, n...) = rand(T, n...) .* rand(T <: Complex ? [1, im, -1, -im] : [1, -1], n...)
+
+    # make sure that each eigenvector has one clearly defined entry with maximum magnitude
+    # so that the normalization of EVs is well defined
+    V = _rand(T, n, n)
+    for (col, i) in zip(eachcol(V), shuffle(1:n))
+        col[i] += 3 * (T <: Complex ? cispi(2rand()) : rand([1, -1]))
+        normalize!(col)
+    end
+
+    # make sure the sorting of eigen values is well defined
+    λ = 10(_rand(T, n) .+ (0:3:3(n-1)))
+
+    return V * Diagonal(λ) / V
+end
+
 @testset "Factorizations" begin
     @testset "lu decomposition" begin
         n = 10
@@ -148,21 +167,7 @@ const LU_NO_PIVOT = VERSION >= v"1.7.0-DEV.1188" ? NoPivot() : Val(false)
             n = 10
 
             @testset "eigen!(::Matrix{$T}) frule" for T in (Float64,ComplexF64)
-                # uniform distribution over `(-1, 1)` / `(-1, 1)^2`
-                _rand(T, n...) = rand(T, n...) .* rand(T <: Complex ? [1, im, -1, -im] : [1, -1], n...)
-
-                # make sure that each eigenvector has one clearly defined entry with maximum magnitude
-                # so that the normalization of EVs is well defined
-                V = _rand(T, n, n)
-                for (col, i) in zip(eachcol(V), shuffle(1:n))
-                    col[i] += 3 * (T <: Complex ? cispi(2rand()) : rand([1, -1]))
-                    normalize!(col)
-                end
-
-                # make sure the sorting of eigen values is well defined
-                λ = 10(_rand(T, n) .+ (0:3:3(n-1)))
-
-                X = V * Diagonal(λ) / V
+                X = rand_eigen(T, n)
 
                 Ẋ = rand_tangent(X)
                 F = eigen!(copy(X))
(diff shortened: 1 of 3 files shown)


simeonschaub:
Ah, I was worried about the x86 failures, but seems like the x86 nightlies are just a bit behind, so we are still running into #422.

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -172,10 +185,21 @@ const LU_NO_PIVOT = VERSION >= v"1.7.0-DEV.1188" ? NoPivot() : Val(false)
...
+                V = _rand(T, n, n)
+                for (col, i) in zip(eachcol(V), shuffle(1:n))
+                    col[i] += 3 * (T <: Complex ? cispi(2rand()) : rand([1, -1]))
+                    normalize!(col)
+                end
+
+                # make sure the sorting of eigen values is well defined
+                λ = 10(_rand(T, n) .+ (0:3:3(n-1)))
+
+                X = V * Diagonal(λ) / V
```
simeonschaub, line 203: @oxinabox Do you prefer duplicating this here over moving this into its own function?

oxinabox: I prefer duplicating.
Though we might like to make a helper for this in ChainRulesTestUtils? at some point
Like how we have `generate_well_conditioned_matrix`

simeonschaub: I did look into whether this was just a matter of generating a matrix with a low enough condition number, but found that that didn't have much to say about whether that matrix was suitable for the testing `eigen` or not. This likely means that "well conditioned" here very much depends on what you are trying to do with the matrix, so I don't think we can have such a generalized function.

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -187,7 +211,10 @@ const LU_NO_PIVOT = VERSION >= v"1.7.0-DEV.1188" ? NoPivot() : Val(false)
                 _, X̄_values_ad = @maybe_inferred back(CT(values = λ̄))
                 @test X̄_values_ad ≈ j′vp(_fdm, x -> eigen(x).values, λ̄, X)[1]
                 _, X̄_vectors_ad = @maybe_inferred back(CT(vectors = V̄))
-                @test X̄_vectors_ad ≈ j′vp(_fdm, x -> eigen(x).vectors, V̄, X)[1] rtol=1e-4
+                # need the conversion to `complex` here, since FiniteDiff is currently buggy if functions
+                # return arrays of either real or complex values based solely on the input values (not the
+                # input types)
```
oxinabox, line 216:
```suggestion
                # input types). See https://github.com/JuliaLang/julia/issues/41243
```

test/rulesets/LinearAlgebra/factorization.jl
oxinabox: actually moving it to its own function here could be good.

Your call

oxinabox (approved):

test/rulesets/LinearAlgebra/factorization.jl
simeonschaub: Ok, I changed it. Seth agreed to give this another look today or tomorrow, so I'd like to hold off on merging until then.

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -23,6 +23,25 @@ end
 const LU_ROW_MAXIMUM = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum() : Val(true)
 const LU_NO_PIVOT = VERSION >= v"1.7.0-DEV.1188" ? NoPivot() : Val(false)
 
+# well-conditioned random n×n matrix with elements of type `T` for testing `eigen`
+function rand_eigen(T::Type, n::Int)
+    # uniform distribution over `(-1, 1)` / `(-1, 1)^2`
+    _rand(T, n...) = rand(T, n...) .* rand(T <: Complex ? [1, im, -1, -im] : [1, -1], n...)
+
+    # make sure that each eigenvector has one clearly defined entry with maximum magnitude
+    # so that the normalization of EVs is well defined
```
sethaxen, line 32: Technically it's so that the phase is well defined.

test/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -23,6 +23,25 @@ end
...
+    V = _rand(T, n, n)
+    for (col, i) in zip(eachcol(V), shuffle(1:n))
+        col[i] += 3 * (T <: Complex ? cispi(2rand()) : rand([1, -1]))
+        normalize!(col)
+    end
+
+    # make sure the sorting of eigen values is well defined
```
sethaxen, line 39:
```suggestion
    # make sure the sorting of eigenvalues is well defined
```

sethaxen (approved):
TBH, I don't know as much as I would like about generating well-conditioned matrices, but this seems to steer clear of the problems that caused FD to fail for me for these rules, so, LGTM.

test/rulesets/LinearAlgebra/factorization.jl
simeonschaub: Ah, you are right! I messed up my terminology here.

simeonschaub:
```suggestion
    # so that the complex phase of EVs is well defined
```

simeonschaub:
Great! Thanks for taking another look! I am always a bit unsure with things like this.

simeonschaub:
feels good meme

TFW CI is finally all-green... (OK, coverage complains, but nobody cares about coverage anyways ;))

PR #447, opened 2021-06-15
(Fix #446) Widen _mulsubtrans!! type signature

### LINKED ISSUE #446: Widen signature for _mulsubtrans!! to accept matrices of different element types
The signature for `_mulsubtrans!!` is `_mulsubtrans!!(X::AbstractMatrix{T}, F::AbstractMatrix{T}) `
and should be widened to capture the case where `X` and `F` have different element types.

niklasschmitz:
Fixes #446

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.8"
+version = "0.8.9"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.7"
+version = "0.8.8"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.9"
+version = "0.8.10"
 
 [deps]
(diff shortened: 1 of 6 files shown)


simeonschaub:
I think what we should probably do is promote both eltypes and then only reuse the left matrix if that type is a subtype of the left matrix and allocate a new array with that eltype otherwise.

niklasschmitz:
I agree, promotion seems better. I've updated it now

src/rulesets/LinearAlgebra/utils.jl
```diff
@@ -2,16 +2,18 @@
 # to any particular rule definition
 
 # F .* (X - X'), overwrites X if possible
-function _mulsubtrans!!(X::AbstractMatrix{T}, F::AbstractMatrix{T}) where T<:Real
+function _mulsubtrans!!(X::AbstractMatrix{<:Real}, F::AbstractMatrix{<:Real})
+    T = promote_type(eltype(X), eltype(F))
+    Y = (T == eltype(X)) ? X : similar(X, T)
```
simeonschaub, line 7: Not sure if there are many situations where this could happen, but just to be on the safe side.
```suggestion
    Y = (T <: eltype(X)) ? X : similar(X, T)
```

simeonschaub (approved):
Looks good to me, but I would prefer if someone else could confirm that this is the right behavior here (I think @ararslan originally added those definitions to Nabla?).

ararslan:
This seems familiar like maybe I wrote it initially but I have no recollection otherwise. If you're looking for another set of eyes, perhaps @oxinabox or another active ChainRules user could take a look?

mzgubic (approved):
Please bump the version but LGTM otherwise - thanks for taking this on!

simeonschaub:
Why do I always forget about the version bump... :laughing:

niklasschmitz:
Thanks for the quick response & review!

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.8"
+version = "0.8.9"
```
simeonschaub, line 3: Master is 0.8.9 now, so you will likely need to rebase and change that to 0.8.10

simeonschaub:
Thanks for the contribution!

PR #449, opened 2021-06-17
Test accepting thunks

### LINKED ISSUE #408: `frule`s/`rrule`s should support receiving `Thunk`s
This came up in https://github.com/FluxML/Zygote.jl/pull/966, which wants to support thunks inside Zygote, not just unthunking them as they come out of rrules.

Right now rrules are not explicitly required to support thunks as the input sensitivities. We could enforce this inside ChainRulesTestUtils, and if we do, we should probably decide for 1.0?

mzgubic:
Closes #408 

Needs https://github.com/JuliaDiff/ChainRulesCore.jl/pull/371

to do:
remove examples where the thunks can be unthunked more than once

- [x] Base
- [x] Core
- [x] LinearAlgebra
- [x] Random
- [x] Statistics

### DIFF
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -88,7 +88,7 @@ let
         function rrule(::typeof(abs2), z::Union{Real, Complex})
             function abs2_pullback(ΔΩ)
                 Δu = real(ΔΩ)
-                return (NoTangent(), 2real(ΔΩ)*z)
+                return (NoTangent(), 2Δu*z)
             end
             return abs2(z), abs2_pullback
         end
@@ -213,7 +213,8 @@ let
         end
 
         function rrule(::typeof(*), x::Number, y::Number)
-            function times_pullback(ΔΩ)
+            function times_pullback(Ω̇)
+                ΔΩ = unthunk(Ω̇)
                 return (NoTangent(),  ΔΩ * y', x' * ΔΩ)
             end
             return x * y, times_pullback
--- a/src/rulesets/Base/sort.jl
+++ b/src/rulesets/Base/sort.jl
@@ -20,13 +20,14 @@ function rrule(::typeof(sort), xs::AbstractVector; kwargs...)
     inds = sortperm(xs; kwargs...)
     ys = xs[inds]
 
-    function sort_pullback(Δys)
+    function sort_pullback(ȳ)
+        Δys = unthunk(ȳ)
         function sort_add!(Δxs)
             Δxs[inds] += Δys
             return Δxs
         end
 
-        Δxs = InplaceableThunk(@thunk(sort_add!(zero(xs))), sort_add!)
+        Δxs = InplaceableThunk(@thunk(sort_add!(zero(Δys))), sort_add!)
 
         return NoTangent(), Δxs
     end
--- a/src/rulesets/Base/sort.jl
+++ b/src/rulesets/Base/sort.jl
@@ -26,7 +26,7 @@ function rrule(::typeof(sort), xs::AbstractVector; kwargs...)
             return Δxs
         end
 
-        Δxs = InplaceableThunk(@thunk(sort_add!(zero(Δys))), sort_add!)
+        Δxs = InplaceableThunk(@thunk(sort_add!(zero(xs))), sort_add!)
 
         return NoTangent(), Δxs
     end
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -75,6 +75,8 @@ function frule(
     return F, ∂F
 end
 
+# these functions are defined outside the rrule because otherwise type inference breaks
+# see https://github.com/JuliaLang/julia/issues/40990
 function _lu_pullback(ΔF::Tangent, m, n, eltypeA, pivot, F)
     Δfactors = ΔF.factors
     Δfactors isa AbstractZero && return (NoTangent(), Δfactors, NoTangent())
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -19,8 +19,9 @@ const LU_RowMaximum = VERSION >= v"1.7.0-DEV.1188" ? RowMaximum : Val{true}
 const LU_NoPivot = VERSION >= v"1.7.0-DEV.1188" ? NoPivot : Val{false}
 
 function frule(
-    (_, ΔA), ::typeof(lu!), A::StridedMatrix, pivot::Union{LU_RowMaximum,LU_NoPivot}; kwargs...
+    (_, Ȧ), ::typeof(lu!), A::StridedMatrix, pivot::Union{LU_RowMaximum,LU_NoPivot}; kwargs...
 )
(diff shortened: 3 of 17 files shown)


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -74,64 +75,66 @@ function frule(
...
-            ∂A1 .+= triu!(mul!(rmul!(∂U, U'), L2', ∂A2, -1, 1))
-            ldiv!(L1', ∂A1)
-            rdiv!(∂A, U')
-        end
-        if pivot isa LU_RowMaximum
-            ∂A = ∂A[invperm(F.p), :]
-        end
-        return NoTangent(), ∂A, NoTangent()
-    end
+    lu_pullback(ȳ) = _lu_pullback(ȳ, A, pivot, F)
```
mzgubic, line 137: the body of the function does not change. The name changes from `lu_pullback` to `_lu_pullback` and it gets the extra arguments `A`, `pivot`, and `F`.

I guess we could close over `m`, `n` rather than `A` here, will change when doing the rest of review

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -88,56 +90,64 @@ end
...
 function rrule(::Type{<:Transpose}, A::AbstractVector{<:Number})
-    Transpose_pullback(ȳ::Tangent) = (NoTangent(), vec(ȳ.parent))
-    Transpose_pullback(ȳ::AbstractMatrix) = (NoTangent(), vec(Transpose(ȳ)))
-    return Transpose(A), Transpose_pullback
+    return Transpose(A), _Transpose_vec_pullback
 end
 
+_transpose_mat_pullback(ȳ::Tangent) = (NoTangent(), ȳ.parent)
+_transpose_mat_pullback(ȳ::AbstractVecOrMat) = (NoTangent(), transpose(ȳ))
+_transpose_mat_pullback(ȳ::AbstractThunk) = return _transpose_mat_pullback(unthunk(ȳ))
```
mzgubic, line 141: If these functions are define inside the rrule, type inference breaks. 🤷

oxinabox: should add a comment to this effect

src/rulesets/Base/fastmath_able.jl
```diff
@@ -88,7 +88,7 @@ let
         function rrule(::typeof(abs2), z::Union{Real, Complex})
             function abs2_pullback(ΔΩ)
                 Δu = real(ΔΩ)
-                return (NoTangent(), 2real(ΔΩ)*z)
+                return (NoTangent(), 2Δu*z)
```
oxinabox, line 91: unrelated cleanup?

src/rulesets/Base/sort.jl
```diff
@@ -26,7 +26,7 @@ function rrule(::typeof(sort), xs::AbstractVector; kwargs...)
             return Δxs
         end
 
-        Δxs = InplaceableThunk(@thunk(sort_add!(zero(Δys))), sort_add!)
+        Δxs = InplaceableThunk(@thunk(sort_add!(zero(xs))), sort_add!)
```
oxinabox, line 29: This seems wrong.
The zero of the primal doesn't have to exist.
(If it does, it is probably also the zero of the differential)

Consider `xs::Vector{DateTime}`
then `Δys` is either a `Vector{<:Period}` or a `Vector{<:Tangent{DateTime}}`

In this case it is bugged.
https://github.com/JuliaLang/julia/issues/41348
but if it was a a `Vector{<:Period}` it wouldn't be

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -74,64 +75,66 @@ function frule(
     return F, ∂F
 end
 
+function _lu_pullback(ΔF::Tangent, A, pivot, F)
```
oxinabox, line 78: I guess we had to move this because else it wouldn't type infer?
Would be good to add a comment

test/runtests.jl
```diff
@@ -14,6 +14,8 @@ using StaticArrays
 using Statistics
 using Test
 
+push!(ChainRulesTestUtils.TRANSFORMS_TO_ALT_TANGENTS, x -> @thunk(x))
```
oxinabox, line 17:
```suggestion
# we also test on Thunks now:
push!(ChainRulesTestUtils.TRANSFORMS_TO_ALT_TANGENTS, x -> @thunk(x))
```

oxinabox (approved):
only 1 comment of any real meaning.
Once addressed merge when ready

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -75,13 +75,14 @@ function frule(
...
 
-function _lu_pullback(ΔF::Tangent, A, pivot, F)
+# these functions are defined outside the rrule because otherwise type inference breaks
+# see https://github.com/JuliaLang/julia/issues/40990
+function _lu_pullback(ΔF::Tangent, m, n, eltypeA, pivot, F)
     Δfactors = ΔF.factors
     Δfactors isa AbstractZero && return (NoTangent(), Δfactors, NoTangent())
     factors = F.factors
-    ∂factors = eltype(A) <: Real ? real(Δfactors) : Δfactors
+    ∂factors = eltypeA <: Real ? real(Δfactors) : Δfactors
```
oxinabox, line 84: Check that this works right.
I suspect you need to do

`_lu_pullback(ΔF::Tangent, m, n, :Type{eltypeA}, pivot, F) where eltypeA` 
or julia will not specialize on this and it will result in it being type-unstable.

mzgubic: It does seem to work right
```julia
julia> y, pb = rrule(lu, randn(T, m, n), pivot)

julia> @code_warntype pb(ChainRulesTestUtils.rand_tangent(y))
Variables
  #self#::ChainRules.var"#lu_pullback#1571"{Matrix{Float64}, Val{true}, LU{Float64, Matrix{Float64}}, Int64, Int64}
  ȳ::Tangent{LU{Float64, Matrix{Float64}}, NamedTuple{(:factors, :ipiv, :info), Tuple{Matrix{Float64}, Vector{NoTangent}, NoTangent}}}

Body::Tuple{NoTangent, Matrix{Float64}, NoTangent}
1 ─ %1 = Core.getfield(#self#, :m)::Int64
│   %2 = Core.getfield(#self#, :n)::Int64
│   %3 = Core.getfield(#self#, :A)::Matrix{Float64}
│   %4 = ChainRules.eltype(%3)::Core.Const(Float64)
│   %5 = Core.getfield(#self#, :pivot)::Core.Const(Val{true}())
│   %6 = Core.getfield(#self#, :F)::LU{Float64, Matrix{Float64}}
│   %7 = ChainRules._lu_pullback(ȳ, %1, %2, %4, %5, %6)::Core.PartialStruct(Tuple{NoTangent, Matrix{Float64}, NoTangent}, Any[Core.Const(NoTangent()), Matrix{Float64}, Core.Const(NoTangent())])
└──      return %7
```

but also the `@inferred` tests should have caught a regression, right?

oxinabox: nice, good.
Yes they should ahve

PR #453, opened 2021-06-22
WIP: use `project` tangents

mzgubic:
Usage of https://github.com/JuliaDiff/ChainRulesCore.jl/pull/380.

This will need a breaking version of FiniteDifferences which perturbs the Diagonal only down its diagonal entries, and not also the zeros.

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -28,12 +28,12 @@ function rrule(
         return (
             NoTangent(),
             InplaceableThunk(
-                @thunk(Ȳ * B'),
-                X̄ -> mul!(X̄, Ȳ, B', true, true)
+                @thunk(project(A, Ȳ * B')),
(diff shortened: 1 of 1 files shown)


src/rulesets/Base/arraymath.jl
```diff
@@ -28,12 +28,12 @@ function rrule(
         return (
             NoTangent(),
             InplaceableThunk(
-                @thunk(Ȳ * B'),
-                X̄ -> mul!(X̄, Ȳ, B', true, true)
+                @thunk(project(A, Ȳ * B')),
+                X̄ -> project(A, mul!(X̄, Ȳ, B', true, true))
```
oxinabox, line 32: I don't think this line is right.
mul has already mutated X bar,
X bar already must have had the right structure so we are fine.

Possibly we needed to project one or both of the inputs to the `mul!`

mzgubic:
close in favour of https://github.com/JuliaDiff/ChainRules.jl/pull/459

PR #454, opened 2021-06-24
handle 1 arg matrix cholesky

oxinabox:
Basically just by inserting the default.
I feel like we shouldn't need this, but Nabla can't seem to AD though getting from
`cholesky(M) = cholesky(M, Val(false))`
when only `cholesky(M, Val(false))` has a rule.

Maybe I need a `@unionize_intercept` or something (@willtebbutt ?)
I would rather fix this in Nabla than add this rule

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -431,11 +431,21 @@ end
 #####
 ##### `cholesky`
 #####
-function rrule(::typeof(cholesky), A::Real)
-    C, full_pb = rrule(cholesky, A, :U)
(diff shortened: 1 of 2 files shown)


oxinabox:
I think the answer is Nabla can't handle this and needs a rule

mzgubic (approved):
I don't know why passing Val(false) in the other cases is right, but taking it as an assumption this seems fine.

oxinabox:
> I don't know why passing Val(false) in the other cases is right, but taking it as an assumption this seems fine.

It is the default that the primal uses
It is `cholesky(matrix, pivot=Val(false))`

PR #457, opened 2021-06-28
WIP: use `preproject` implementation

mzgubic:
This is quite ugly since it extracts the relevant information outside the pullback. An extra complication is that
```julia
julia> function noinfers(x::AbstractArray)
           T = eltype(x)
           function myclosure()
               return zeros(T, 2, 2)
           end
           return myclosure
       end
noinfers (generic function with 2 methods)

julia> cl = noinfers([1.0, 2.0, 3.0]); @inferred cl()
ERROR: return type Matrix{Float64} does not match inferred return type Matrix{_A} where _A
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ REPL[82]:1

julia> function infers(x::AbstractArray)
           function myclosure()
               return zeros(eltype(x), 2, 2)
           end
           return myclosure
       end
infers (generic function with 2 methods)

julia> cl = infers([1.0, 2.0, 3.0]); @inferred cl()
2×2 Matrix{Float64}:
 0.0  0.0
 0.0  0.0
```
to which @mcabbott suggested the `_val` solution on Slack. (without seeing the use case, so he might know a more elegant solution)

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -24,18 +24,21 @@ function rrule(
     A::AbstractVecOrMat{<:CommutativeMulNumber},
     B::AbstractVecOrMat{<:CommutativeMulNumber},
 )
+    Ainfo = preproject(A)
+    Binfo = preproject(B)
+    valAtype = Val(typeof(A))
+    valBtype = Val(typeof(B))
+    _val(::Val{T}) where T = T
     function times_pullback(Ȳ)
-        return (
-            NoTangent(),
-            InplaceableThunk(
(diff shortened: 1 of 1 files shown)


oxinabox:
The extra complication of closing over `eltype(a)` being type unstable would be fixed by https://github.com/JuliaLang/julia/pull/40985 (but will have to wait for Julia 1.8 for that, if it is merged)

mzgubic:
close in favour of https://github.com/JuliaDiff/ChainRules.jl/pull/459

PR #458, opened 2021-06-28
No Differentiable Factorial

### LINKED ISSUE #344: Missing rule for factorial
Should be added using `@non_differentiable factorial(n::Integer)`

AlexRobson:
Fixes #344 

There are no tests for non_differentiable as discussed in https://github.com/JuliaDiff/ChainRules.jl/pull/252 . This *does* add a test here as it fits the criterion of either "add(ing) regression tests when we find issues with these rules / decide to add extra rules" but can just remove these tests if it's not wanted.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "0.8.16"
+version = "0.8.17"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -1,6 +1,7 @@
 using Base.Broadcast: broadcastable
 using ChainRules
 using ChainRulesCore
+using ChainRulesCore: NoTangent, rrule
```
mzgubic, line 4:
```suggestion
```
These should be exported and just work because of the line above

mzgubic (approved):
I don't think we need the tests (since we test the macro in ChainRulesCore), and the CI already takes ~35min.

test/runtests.jl
AlexRobson: Cool - I've removed the commit that added the tests (I was half expecting to do this anyway). 

Now the changeset is just the Project.toml and adding the no diff rule.

PR #459, opened 2021-06-29
use ProjectTo functor everywhere, and update to v1 of CRC and CRTU

oxinabox:
This is the partner to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/385
it matches against #457 

It is much cleaner and it *should* be good again inference problems from closing over types,
as the type gets put into a type-parameter of `project_A` and `project_B`

Following minimized version works find for me, haven't tested real thing properly.
```julia
using Test, LinearAlgebra, ChainRulesCore

function projectto_infers(x::AbstractArray)
       project = ProjectTo(x)
       function myclosure(y)
           return project(y)
       end
       return myclosure
   end
   
   
const pb = projectto_infers(Diagonal([1.0, 2.0]))
@inferred pb([1.0 2.0; 3.0 4.0])
```


src/rulesets/Base/array.jl
```diff
@@ -196,7 +199,9 @@ function rrule(::typeof(hvcat), rows, values::Union{AbstractArray, Number}...)
             end
             dY[index...]
         end
-        return (NoTangent(), NoTangent(), dXs...)
+        _call((f, x)) = f(x)
+        proj_dXs = _call.(tuple(zip(project_Vs, dXs)...)) # can't make it type infer, e.g. test_rrule(hvcat, 1, rand(3)', transpose(rand(3)) ⊢ rand(1,3))
+        return (NoTangent(), NoTangent(), proj_dXs...)
```
mzgubic, line 204: @oxinabox any ideas here?

oxinabox: I suspect it is failing to infer the length.
Which is like https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/185#issuecomment-871715279
so see the solutions there.

I think maybe changing the dims arg to the rrule to 
```julia
, dims::Tuple{Vararg{Union{Colon,Int},N}}) where N
```
and then  as an assert:
```
proj_dXs::NTuple{N} = 
```
Or on the other side. as a convert

Or might need it right on the end after the return (which would be a convert)

src/rulesets/Base/array.jl
```diff
@@ -196,7 +199,9 @@ function rrule(::typeof(hvcat), rows, values::Union{AbstractArray, Number}...)
             end
             dY[index...]
         end
-        return (NoTangent(), NoTangent(), dXs...)
+        _call((f, x)) = f(x)
+        proj_dXs = _call.(tuple(zip(project_Vs, dXs)...)) # can't make it type infer, e.g. test_rrule(hvcat, 1, rand(3)', transpose(rand(3)) ⊢ rand(1,3))
```
oxinabox, line 203: This is pretty ugly.
Isn;t it the same as 
```suggestion
        _call(f, x) = f(x)
        proj_dXs = map(_call, project_Vs, dXs)
```
Which may or may not fix inference, but is at least more readable.

src/rulesets/Base/array.jl
mzgubic: The length infers fine. It can't tell what the elements are, even though it can if it is doing them one at a time. This is the ambiguity when `@code_warntype` the `pb` of the example in the comment.
```julia
Body::Tuple{NoTangent, NoTangent, Union{Adjoint{Float64, Matrix{Float64}}, Transpose{Float64, Matrix{Float64}}}, Union{Adjoint{Float64, Matrix{Float64}}, Transpose{Float64, Matrix{Float64}}}}
```

src/rulesets/Base/array.jl
mzgubic: It's like that because a vector of functions does not infer, but a tuple of functions does. e.g. see
```julia
julia> @code_warntype map(zip([-, +], [1.0, 2.0])) do (f, b)
           return f(b)
       end
```
vs
```julia
julia> @code_warntype map(zip((-, +), [1.0, 2.0])) do (f, b)
           return f(b)
       end
```
(but that alone dos not seem to be enough)

oxinabox: right but `map` over two tuples does infer.
```julia
julia> @code_warntype map((-, +), (1.0, 2.0)) do f, b
                  return f(b)
              end
MethodInstance for map(::var"#41#42", ::Tuple{typeof(-), typeof(+)}, ::Tuple{Float64, Float64})
  from map(f, t::Tuple{Any, Any}, s::Tuple{Any, Any}) in Base at tuple.jl:247
Arguments
  #self#::Core.Const(map)
  f::Core.Const(var"#41#42"())
  t::Core.Const((-, +))
  s::Tuple{Float64, Float64}
Body::Tuple{Float64, Float64}
1 ─      nothing
│   %2 = Base.getindex(t, 1)::Core.Const(-)
│   %3 = Base.getindex(s, 1)::Float64
│   %4 = (f)(%2, %3)::Float64
│   %5 = Base.getindex(t, 2)::Core.Const(+)
│   %6 = Base.getindex(s, 2)::Float64
│   %7 = (f)(%5, %6)::Float64
│   %8 = Core.tuple(%4, %7)::Tuple{Float64, Float64}
└──      return %8
```

oxinabox: From what I can tell this fixes it.
I will push these changes

oxinabox:
I have made changes that make the following all work
```
using ChainRulesCore, ChainRules, Test
@inferred(
    rrule(hvcat, 1, rand(3)', transpose(rand(3)))[2](rand(2,3))
)
using ChainRulesTestUtils
test_rrule(hvcat, 1, rand(3)', transpose(rand(3)) ⊢ rand(1,3))
```

src/rulesets/LinearAlgebra/structured.jl
```diff
@@ -7,21 +7,25 @@ const SquareMatrix{T} = Union{Diagonal{T}, AbstractTriangular{T}}
...
 function rrule(::typeof(/), A::AbstractMatrix{<:Real}, B::T) where T<:SquareMatrix{<:Real}
     Y = A / B
+    project_A = ProjectTo(A)
+    project_B = ProjectTo(B)
     function slash_pullback(ȳ)
         Ȳ = unthunk(ȳ)
-        ∂A = @thunk Ȳ / B'
-        ∂B = @thunk _unionall_wrapper(T)(-Y' * (Ȳ / B'))
+        ∂A = @thunk project_A(Ȳ / B')
+        ∂B = @thunk project_B(_unionall_wrapper(T)(-Y' * (Ȳ / B')))
```
oxinabox, line 15: I think the `_unionall_wrapper` might actually be doing the same thing `project` does.
It is worth checking that

src/rulesets/Base/mapreduce.jl
```diff
@@ -40,14 +40,14 @@ function rrule(
 end
 
 
+# TODO: add project in a way that does not break inference
```
oxinabox: Open an issue and link here?

test/rulesets/Base/array.jl
```diff
@@ -28,9 +28,9 @@ end
     dy = 1 ./ reduce(hcat, adjs)
     @test rrule(reduce, hcat, adjs)[2](dy)[3] ≈ rrule(reduce, hcat, collect.(adjs))[2](dy)[3]
 
-    # mix types TODO: broken, applying project breaks type stability
-    #mats = [randn(2, 2), rand(2, 2)']
-    #test_rrule(reduce, hcat ⊢ NoTangent(), mats; check_inferred=VERSION>v"1.1")
+    # mix types
+    mats = [randn(2, 2), rand(2, 2)']
+    test_rrule(reduce, hcat ⊢ NoTangent(), mats; check_inferred=VERSION>v"1.1", test_types=false)
```
oxinabox, line 33:
```suggestion
    test_rrule(reduce, hcat ⊢ NoTangent(), mats; check_inferred=VERSION>v"1.1")
```

test/rulesets/Base/mapreduce.jl
```diff
@@ -36,8 +36,8 @@
 
         # covectors
         x = [-4.0 2.0; 2.0 -1.0]
-        test_rrule(sum, inv, x[1, :]')
-        test_rrule(sum, inv, x[1:1, :]')
+        test_rrule(sum, inv, x[1, :]'; test_types=false)
```
oxinabox, line 40:
```suggestion
        test_rrule(sum, inv, x[1, :]')
```

test/rulesets/Base/mapreduce.jl
```diff
@@ -36,8 +36,8 @@
...
+        test_rrule(sum, inv, x[1:1, :]', test_types=false)
```
oxinabox, line 41:
```suggestion
        test_rrule(sum, inv, x[1:1, :]')
```

oxinabox:
I approve this (but can't click approve)

src/rulesets/Base/array.jl
```diff
@@ -81,10 +81,11 @@ function rrule(::typeof(hcat), Xs::Union{AbstractArray, Number}...)
     Y = hcat(Xs...)  # note that Y always has 1-based indexing, even if X isa OffsetArray
     ndimsY = Val(ndims(Y))  # this avoids closing over Y, Val() is essential for type-stability
     sizes = map(size, Xs)   # this avoids closing over Xs
+    project_Xs = map(ProjectTo, Xs)
     function hcat_pullback(ȳ)
         dY = unthunk(ȳ)
         hi = Ref(0)  # Ref avoids hi::Core.Box
-        dXs = map(sizes) do sizeX
+        dXs = map(project_Xs, sizes) do project, sizeX
             ndimsX = length(sizeX)
```
mcabbott, line 89: Seems like a lot of boilerplate to add this to every rule definition by hand. Can't the computer do this?

oxinabox: Maybe, but it's beyond the scope of this PR.
This is the much reduced boiler plate version.

mzgubic: Possibly we could, though there are a few complications.
1) project needs to be inside thunks, because projecting the thunks would unthunk them (unless we want to nest thunks, which I am told can be a bad thing "because compiler")
2) we'd have to check whether the performance is worse for rules that don't actually need projecting

mcabbott: What I mean is that this is a high-level design question, whether this thing is to be globally applied, or is something to be selectively hand-added to rules which need it. I don't know if I missed this, I couldn't follow everything last week, but was this discussed or did it just sort-of happen?

There was a working implementation here, which I think is applied after un-thunking:
https://github.com/FluxML/Zygote.jl/pull/965/files#diff-e0bc7da8f1a33a59f5ecfa67257c04038f0b4915b3f74bdf39780818fd0010a2R73-R97

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -7,14 +7,16 @@ function frule((_, Δx, Δy), ::typeof(dot), x, y)
...
 
 function rrule(::typeof(dot), x::AbstractArray, y::AbstractArray)
+    project_x = ProjectTo(x)
+    project_y = ProjectTo(y)
     function dot_pullback(Ω̄)
         ΔΩ = unthunk(Ω̄)
         xthunk = InplaceableThunk(
-            @thunk(reshape(y .* ΔΩ', axes(x))),
+            @thunk(project_x(reshape(y .* ΔΩ', axes(x)))),
             dx -> dx .+= reshape(y, axes(x)) .* ΔΩ',
```
mcabbott, line 16: The other design issue I meant to raise is this: if x is real and y is complex, then this in-place rule will give an InexactError. I believe you probably want to insert something like `eltype(project_x)` into that broadcast. Which seems much tricker to do automatically.

Non-broadcasting in-place rules like `mul!` would need something similar, but not quite so simple.

mcabbott:
One way you could avoid boilerplate here is to define a macro which turns
```
@ProjectEverything function rrule(f, x, ys...)
   # do stuff
  return y, back
end
```
into
```
function rrule(f, x, ys...)
   # do stuff
  extra = MultiProject(f, x, ys...)
  return y, extra∘back
end
```
In fact even without a macro, a `MultiProject` which you compose like this would cut a lot of lines here.

mzgubic:
still failing:
`y, pb = rrule(/, rand(3), rand(3)); @inferred pb(rand(3, 3))`

oxinabox:
For the inference problems, I am happy enough to open an issue and come back to them.
For 
> `y, pb = rrule(rc, sum, sum, [[2.0, 4.0], [4.0,1.9]]); pb(y)` (can't deal with project receiving an InplaceableThunk, rc is the fake ruleconfig)

Can we insert a `unthunk` here before  the `project` ?
We do not *always* want to project InplacableThunks via `unthunking` but in this case we do,
The kinda function people realistically use with `sum(f,...)` are ones that we would basically never want to preserve thunking through.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -11,15 +11,9 @@ function rrule(::typeof(dot), x::AbstractArray, y::AbstractArray)
...
-        xthunk = InplaceableThunk(
-            dx -> dx .+= reshape(y, axes(x)) .* ΔΩ',
-            @thunk(project_x(reshape(y .* ΔΩ', axes(x)))),
-        )
-        ythunk = InplaceableThunk(
-            dy -> dy .+= reshape(x, axes(y)) .* ΔΩ,
-            @thunk(project_y(reshape(x .* ΔΩ, axes(y)))),
-        )
-        return (NoTangent(), xthunk, ythunk)
+        x̄ = @thunk(project_x(reshape(y .* ΔΩ', axes(x))))
```
mzgubic, line 14: Is this permanent or until we have something like https://github.com/JuliaDiff/ChainRulesCore.jl/pull/393?

mzgubic (approved):

src/rulesets/LinearAlgebra/dense.jl
oxinabox: Yeah,
more generally I hate the InplacableThunks that are just deferring broadcast materialization.
They are so boilerplatey.
That has its own issue https://github.com/JuliaDiff/ChainRulesCore.jl/issues/274
and is probably going to be easier to solve.

I am more than happy to delete them.
Can always add them back

PR #460, opened 2021-06-29
Add repeat rules from Zygote

AlexRobson:
Adds rrules for repeat. These are essentially copied from Zygote as suggested in #383 and hence taken from [here](https://github.com/FluxML/Zygote.jl/blob/956cbcf3c572c0eb09c146189bb38b1b434634ff/src/lib/array.jl#L130)

Doesn't include missing rules. At least from what I can tell, this is: 

```
function rrule(::typeof(repeat), xs::AbstractArray, counts::Integer...)
```

One change from the original code in Zygote:  closes instead of the value of the size of `xs` rather than `xs` itself. Also adds an `unthunk`. 

Note, I saw #405 after looking at this (that caught the above), but not sure if that is being actively worked on.

### DIFF
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -16,8 +16,10 @@ end
     # zero-arrays: broken
     @test_broken rrule(repeat, fill(1.0), 2) !== nothing
     @test_broken rrule(repeat, fill(1.0), 2, 3) !== nothing
-    @test_broken rrule(repeat, fill(1.0); fkwargs = (inner=2,)) !== nothing
-    @test_broken rrule(repeat, fill(1.0); fkwargs = (inner=2, outer=3,)) !== nothing
+
+    # These dispatch but rrule needs to be fixed to zero-arrays
+    # test_rrule(repeat, fill(1.0); fkwargs = (inner=2,))
+    # test_rrule(repeat, fill(1.0); fkwargs = (inner=2, outer=3,))
 
     @test rrule(repeat, [1,2,3], 4)[2](ones(12))[2] == [4,4,4]
     @test rrule(repeat, [1,2,3], outer=4)[2](ones(12))[2] == [4,4,4]
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -9,15 +9,21 @@ end
     test_rrule(repeat, rand(4, ), 2)
     test_rrule(repeat, rand(4, 5))
     test_rrule(repeat, rand(4, 5); fkwargs = (outer=(1,2),))
-    test_rrule(repeat, rand(4, 5); fkwargs = (inner=(2,4), outer=(1,1,1,3)))
-    test_rrule(repeat, rand(4, 5), 2)
+    test_rrule(repeat, rand(4, 5); fkwargs = (inner=(1,2), outer=(1,3)))
+
+    if VERSION>=v"1.6"
+        # repeat([1 2; 3 4], inner=(2,4), outer=(1,1,1,3)) fails for v<1.6
+        test_rrule(repeat, rand(4, 5); fkwargs = (inner=(2,4), outer=(1,1,1,3)))
+    end
+    test_rrule(repeat, rand(4, 5), 2; check_inferred=VERSION>=v"1.5")
     test_rrule(repeat, rand(4, 5), 2, 3)
 
     # zero-arrays: broken
     @test_broken rrule(repeat, fill(1.0), 2) !== nothing
     @test_broken rrule(repeat, fill(1.0), 2, 3) !== nothing
 
-    # These dispatch but rrule needs to be fixed to zero-arrays
+    # These dispatch but probably needs
+    # https://github.com/JuliaDiff/FiniteDifferences.jl/issues/179
     # test_rrule(repeat, fill(1.0); fkwargs = (inner=2,))
     # test_rrule(repeat, fill(1.0); fkwargs = (inner=2, outer=3,))
 
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -8,8 +8,20 @@ end
     test_rrule(repeat, rand(4, ))
     test_rrule(repeat, rand(4, ), 2)
     test_rrule(repeat, rand(4, 5))
-    test_rrule(repeat, rand(4, 5); fkwargs = (inner = (2,1), outer = (2,2)))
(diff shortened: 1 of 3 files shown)


mzgubic (approved):
Looks really good! Some minor nitpicking follows:
 
I suspect
```julia
test_rrule(repeat, rand(4, 5), 2)
```
would fail (because there is an optional argument in the rule and the new ChainRulesTestUtils version checks that the number of tangents matches the number of arguments)

There are also a few tests in 
https://github.com/FluxML/Zygote.jl/blob/1082ebd3aced63b99c4b6c2956a122ce6a37f97d/test/gradcheck.jl#L232-L239
but from a quick look it seems we've covered them?

mcabbott:
In https://github.com/JuliaDiff/ChainRules.jl/issues/383#issuecomment-838412475 there's a complaint that the rule taking `inner`, `outer` keywords is pretty slow. And the fast rule is only written for a few specific `ndims`. 

I think you can do the general case via broadcasting, I wrote out one implementation here https://gist.github.com/mcabbott/80ac43cca3bee8f57809155a5240519f which you should feel free to steal. But not very carefully checked, it might be worth inventing some evil test cases.

AlexRobson:
> I suspect
> 
> ```julia
> test_rrule(repeat, rand(4, 5), 2)
> ```
> 
> would fail (because there is an optional argument in the rule and the new ChainRulesTestUtils version checks that the number of tangents matches the number of arguments)
> 


Yeah I hit this when I was working with it. I should have included a commeont on this. What's the resolution here - If i'm reading this correctly this is a general problem - any function with omitted positional arguments with defaults will fail under this test. I don't *think* I see any other examples of positional defaults being set within the `rrules`, so perhaps it has just not come up?

What is the intended convention here: to return tangents for all specified arguments or all possible arguments (i.e. `for foo(x, y = 5) = x + 5)`, should this return 1 or 2 tangents (ignoring ∂self) if invoked by e.g. `foo(3)`, given it'll always end up dispatching to the full expression. 

Alternatively, we can write out these explicitely, dropping the defaults. Leads to a bit of duplication, but fine for these small functions. 

```
function rrule(::typeof(repeat), xs::AbstractVecOrMat, m::Integer, n::Integer)...
function rrule(::typeof(repeat), xs::AbstractVecOrMat, m::Integer)...
```

mzgubic:
I only know of a single case, solved here:
https://github.com/JuliaDiff/ChainRules.jl/pull/436/files

@mcabbott as far as I understand it is also slow on Zygote, so we can merge this and the performance will be the same? And then follow up with a better implementation later? Would be good to mention the status on the main issue

mcabbott:
Is there a different main issue besides the one I linked?

Re optional arguments, that seems like a bug, is there an issue tracking it? Slurped arguments do seem to work (as in `repeat([1], 2,3,4,5)`).

test/rulesets/Base/array.jl
```diff
@@ -4,6 +4,14 @@
     test_rrule(reshape, rand(4, 5), 2, :)
 end
 
+@testset "repeat" begin
+    test_rrule(repeat, rand(4, ))
+    test_rrule(repeat, rand(4, ), 2)
+    test_rrule(repeat, rand(4, 5))
+    test_rrule(repeat, rand(4, 5); fkwargs = (inner = (2,1), outer = (2,2)))
+    test_rrule(repeat, rand(4, 5), 2, 3)
```
mcabbott, line 12: Some test cases which you may wish to add, with `@test_broken` if they are broken:
```
repeat([1 2; 3 4], outer=3)  # perhaps works accidentally, compare  repeat([1 2; 3 4], inner=(1,1), outer=2) 
repeat([1 2; 3 4], inner=(2,4), outer=(1,1,1,3))
repeat([1,2,3], 2, 3)

# zero-arrays
repeat(fill(1), 2)
repeat(fill(1), 2, 3)
repeat(fill(1), inner=2)
repeat(fill(1), inner=2, outer=3)

**Comment by AlexRobson** [comment]
Thanks for your input both. I've added a couple of commits to deal with the comments:

- I've used the same approach as used in @mzgubic branch to handle the positional arguments. It is essentially as descrbed above by dropping the positional arguments and using separate methods for them. 
- I've added the extra tests as suggested by @mcabbott. 

The zero array ones are broken for now though, in two ways: 

For `rrule(repeat, fill(1.0), 2)` and friends: 
- Right now this implementation uses `AbstractVector` and `AbbstractVecOrMat` so they are not dispatching.  I believe this should be fixed when we add `function rrule(::typeof(repeat), xs::AbstractArray, counts::Integer...)`. 

For `rrule(repeat, fill(1.0); fkwargs = (inner=2,)))` and friends: 
- This does dispatch, but this is hitting another issue, which I think means the closure in the `rrule` would need tweaking. Specifically, `test_rrule` ends up dispatching to `_test_add!!_behaviour(acc::Float64, val::Array{Float64, 0}` and then failing with `MethodError: no method matching +(::Float64, ::Array{Float64, 0})`. Basically, playing around with it a bit locally, I think the dimensions in the output of the pullback are off with zero arrays.

**Comment by mcabbott** [comment]
Re zero-arrays, it's possible there are issues with whether test_rrule can handle these. It also doesn't seem so obvious what the rule should return -- in lots of other contexts, zero-arrays tend to go away:
```
julia> gradient(x -> x .+ 1, fill(2))
(1,)

julia> gradient(x -> sum(vcat(x, 1)), fill(2))
(1,)
```

File: test/rulesets/Base/array.jl [hunk 2]
```diff
@@ -9,15 +9,21 @@ end
...
     test_rrule(repeat, rand(4, 5), 2, 3)
 
     # zero-arrays: broken
     @test_broken rrule(repeat, fill(1.0), 2) !== nothing
     @test_broken rrule(repeat, fill(1.0), 2, 3) !== nothing
 
-    # These dispatch but rrule needs to be fixed to zero-arrays
+    # These dispatch but probably needs
+    # https://github.com/JuliaDiff/FiniteDifferences.jl/issues/179
     # test_rrule(repeat, fill(1.0); fkwargs = (inner=2,))
```
mzgubic, line 27: are these fixed on your FiniteDifferences branch with the `rand_tangent` fix?

AlexRobson: I believe so. Can uncoment these when doing a follow up for the `function rrule(::typeof(repeat), xs::AbstractArray, counts::Integer...)` general method along with a FiniteDifferences bump. ?

mzgubic (approved):

PR #461, opened 2021-07-01
Add rules for Base.literal_pow

dfdx:
Fixes issue in dfdx/Yota.jl#90

### DIFF
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -182,6 +182,12 @@ const FASTABLE_AST = quote
                 @test ∂y ≈ 0
             end
         end
+
+        @testset "literal_pow" for T in (Float64, ComplexF64)
+            # for real x and n, x must be >0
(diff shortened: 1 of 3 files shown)


test/rulesets/Base/fastmath_able.jl
```diff
@@ -182,6 +182,12 @@ const FASTABLE_AST = quote
                 @test ∂y ≈ 0
             end
         end
+
+        @testset "literal_pow" for T in (Float64, ComplexF64)
+            # for real x and n, x must be >0
+            test_frule(Base.literal_pow, ^, rand(T) + 3, Val(3))
```
oxinabox, line 188: does `test_scalar` work for this?

oxinabox (commented):
I am not sure about using `@scalar_rule` for this.
It was not really designed to accept inputs that are not scalar values
and it is kinda a coincidence that works.
I guess we will notice if it breaks, and if so we can change it then.

dfdx:
Let me prepare another PR with separate `frule()` and `rrule()` so we could compare them side-by-side.

PR #462, opened 2021-07-01
Add rrule for multiarg array addition

dfdx:
Fixes dfdx/Yota.jl#92

This PR doesn't contain `frule` for the same operation, but most other functions in `arraymath.jl` don't have either. Is it normal practice to add only one kind of rules? I don't know much about forward-mode AD ecosystem so it's a bit hard for me to estimate the impact.

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -277,8 +277,7 @@ end
 function rrule(::typeof(+), arrs::AbstractArray...)
     y = +(arrs...)
     arr_axs = map(axes, arrs)
-    function add_pullback(dy_raw)
-        dy = unthunk(dy_raw)
+    function add_pullback(dy)
         return (NoTangent(), map(ax -> reshape(dy, ax), arr_axs)...)
     end
     return y, add_pullback
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -276,9 +276,7 @@ end
 
 function rrule(::typeof(+), arrs::AbstractArray...)
     y = +(arrs...)
-    function add_pullback(dy)
-        # using ntuple here for type stability
-        return (NoTangent(), ntuple(i -> dy, length(arrs))...)
-    end
+    valN = Val(length(arrs)) # makes add_pullback type-stable without closing over arrs
+    add_pullback(dy) = (NoTangent(), ntuple(_ -> dy, valN)...)
     return y, add_pullback
 end
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -276,7 +276,10 @@ end
 
 function rrule(::typeof(+), arrs::AbstractArray...)
     y = +(arrs...)
-    valN = Val(length(arrs)) # makes add_pullback type-stable without closing over arrs
-    add_pullback(dy) = (NoTangent(), ntuple(_ -> dy, valN)...)
+    arr_axs = map(axes, arrs)
+    function add_pullback(dy_raw)
+        dy = unthunk(dy_raw)
+        return (NoTangent(), map(ax -> reshape(dy, ax), arr_axs)...)
+    end
     return y, add_pullback
 end
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -277,10 +277,8 @@ end
 function rrule(::typeof(+), arrs::AbstractArray...)
     y = +(arrs...)
     function add_pullback(dy)
-        return (
-            NoTangent(),
-            ntuple(i -> dy .* ones(eltype(arrs[i]), size(arrs[i])), length(arrs))...
-        )
+        # using ntuple here for type stability
+        return (NoTangent(), ntuple(i -> dy, length(arrs))...)
     end
     return y, add_pullback
 end
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -268,3 +268,19 @@ function rrule(::typeof(-), x::AbstractArray)
     end
     return -x, negation_pullback
 end
+
+
+#####
+##### Addition (Multiarg `+`)
+#####
+
+function rrule(::typeof(+), arrs::AbstractArray...)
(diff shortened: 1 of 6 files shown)


src/rulesets/Base/arraymath.jl
```diff
@@ -268,3 +268,19 @@ function rrule(::typeof(-), x::AbstractArray)
...
+#####
+##### Addition (Multiarg `+`)
+#####
+
+function rrule(::typeof(+), arrs::AbstractArray...)
+    y = +(arrs...)
+    function add_pullback(dy)
+        return (
+            NoTangent(),
+            ntuple(i -> dy .* ones(eltype(arrs[i]), size(arrs[i])), length(arrs))...
```
mzgubic, line 282: Is there a reason we are not using `fill` here instead?

Also, have you considered using thunks? https://juliadiff.org/ChainRulesCore.jl/dev/writing_good_rules.html#Use-Thunks-appropriately

mzgubic:
Thanks for the PR! It's fine to just an `rrule`.

src/rulesets/Base/arraymath.jl
mcabbott: This seems complicated. Isn't the rule just like `dy -> (nothing, dy, dy, dy, ...)`?

dfdx: @mcabbott Indeed! I'll update it.

src/rulesets/Base/arraymath.jl
```diff
@@ -268,3 +268,17 @@ function rrule(::typeof(-), x::AbstractArray)
...
+#####
+##### Addition (Multiarg `+`)
+#####
+
+function rrule(::typeof(+), arrs::AbstractArray...)
+    y = +(arrs...)
+    function add_pullback(dy)
+        # using ntuple here for type stability
+        return (NoTangent(), ntuple(i -> dy, length(arrs))...)
+    end
```
mcabbott, line 282: Not sure it'll matter, but perhaps worth avoiding closing over the arguments, since we only need the length. 
```suggestion
    y = +(arrs...)
    valN = Val(length(arrs)) # makes add_pullback type-stable without closing over arrs
    add_pullback(dy) = (NoTangent(), ntuple(_ -> dy, valN)...)
```

src/rulesets/Base/base.jl
```diff
@@ -161,3 +161,15 @@ end
...
+function frule((_, _, Δx, _), ::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{p}) where p
+    y = Base.literal_pow(^, x, pv)
+    return y, (p * y / x * Δx)
+end
+
+function rrule(::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{p}) where p
+    y = Base.literal_pow(^, x, pv)
+    literal_pow_pullback(dy) = NoTangent(), NoTangent(), (p * y / x * dy), NoTangent()
+    return y, literal_pow_pullback
+end
```
mcabbott, line 175: These are identical to another PR right? Is git / github just confused about what's new and what's not?

https://github.com/JuliaDiff/ChainRules.jl/pull/464/files#diff-dbe52c217d193026d8825d7b34db346dc5c8e7dbab6e28453a03b62b3c196859R164-R175

dfdx: Yes, these are pieces from 2 other PRs, perhaps the result of a poor combination of rebases and merges with changes suggested via Github UI. The codebase shouldn't be affected, but the git history may look a bit weird, so I can simply create a new clean PR (although it will destroy authorship of the suggested edits).

mcabbott: No that sounds like a lot of work. I'm 99% sure it's going to merge smoothly & not duplicate things.

mcabbott (approved):
I now have a merge button, but perhaps I leave it open a little while in case anyone else wants to add thoughts.

src/rulesets/Base/arraymath.jl
mcabbott: Maybe one further question is how much you want to worry about cases like this, Zygote certainly does not:
```julia
julia> ones(3) + ones(3,1)
3×1 Matrix{Float64}:
 2.0
 2.0
 2.0

julia> Zygote.gradient(sum∘+, ones(3), ones(3,1))
(3×1 Fill{Float64}: entries equal to 1.0, 3×1 Fill{Float64}: entries equal to 1.0)
```

src/rulesets/Base/arraymath.jl
```diff
@@ -268,3 +268,15 @@ function rrule(::typeof(-), x::AbstractArray)
...
+#####
+##### Addition (Multiarg `+`)
+#####
+
+function rrule(::typeof(+), arrs::AbstractArray...)
+    y = +(arrs...)
+    valN = Val(length(arrs)) # makes add_pullback type-stable without closing over arrs
+    add_pullback(dy) = (NoTangent(), ntuple(_ -> dy, valN)...)
+    return y, add_pullback
+end
```
mcabbott, line 282: This is a version which would remove trailing dimensions
```suggestion
function rrule(::typeof(+), arrs::AbstractArray...)
    y = +(arrs...)
    arr_axs = map(axes, arrs)
    function add_pullback(dy_raw)
        dy = unthunk(dy_raw)
        return (NoTangent(), map(ax -> reshape(dy, ax), arr_axs)...)
    end
    return y, add_pullback
end
```

test/rulesets/Base/arraymath.jl
```diff
@@ -148,4 +148,12 @@
...
         test_rrule(-, Diagonal(A); output_tangent=Diagonal(Ā))
     end
+
+    @testset "addition" begin
+        A = randn(4, 4)
+        B = randn(4, 4)
+        C = randn(4, 4)
+
+        test_rrule(+, A, B, C)
+    end
```
mcabbott, line 158: And a test with trailing dimensions:
```suggestion
    @testset "addition" begin
        test_rrule(+, randn(4, 4), randn(4, 4), randn(4, 4))
        test_rrule(+, randn(3), randn(3,1), randn(3,1,1))
    end
```

mzgubic:
I've squashed the commits and it seems to have handled it nicely

PR #465, opened 2021-07-02
Add nondiff rules for one ones zero zeros

niklasschmitz:
Porting nograd rules for `one`, `ones`, `zero`, `zeros` from Zygote https://github.com/FluxML/Zygote.jl/blob/master/src/lib/array.jl#L11

Follow-up to https://github.com/JuliaDiff/ChainRules.jl/pull/252

### DIFF
--- a/src/rulesets/Base/nondiff.jl
+++ b/src/rulesets/Base/nondiff.jl
@@ -442,5 +442,5 @@ VERSION >= v"1.1" && @non_differentiable Sys.isopenbsd(::Symbol)
 
 @non_differentiable one(::Any)
 @non_differentiable ones(::Any...)
-@non_differentiable zero(::Any...)
+@non_differentiable zero(::Any)
 @non_differentiable zeros(::Any...)
--- a/src/rulesets/Base/nondiff.jl
+++ b/src/rulesets/Base/nondiff.jl
@@ -440,7 +440,7 @@ VERSION >= v"1.1" && @non_differentiable Sys.isopenbsd(::Symbol)
 
 @non_differentiable similar(::Any...)
 
-@non_differentiable one(::Any...)
+@non_differentiable one(::Any)
 @non_differentiable ones(::Any...)
 @non_differentiable zero(::Any...)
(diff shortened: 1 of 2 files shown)


niklasschmitz:
This was also mentioned in https://github.com/FluxML/Zygote.jl/pull/780 by @oxinabox. Should these rather return `ZeroTangent()` than `NoTangent()` ?

mzgubic:
I suppose `one` and `zero` need `ZeroTangent()` (see https://juliadiff.org/ChainRulesCore.jl/dev/FAQ.html#faq_abstract_zero)

but they work almost identically so I would favour the simplicity of using a macro over being pedantic in this case.

src/rulesets/Base/nondiff.jl
```diff
@@ -439,3 +439,8 @@ VERSION >= v"1.1" && @non_differentiable Sys.isopenbsd(::Symbol)
 @non_differentiable Threads.threadid(::Task)
 
 @non_differentiable similar(::Any...)
+
+@non_differentiable one(::Any...)
```
mzgubic, line 443:
```suggestion
@non_differentiable one(::Any)
```

src/rulesets/Base/nondiff.jl
```diff
@@ -439,3 +439,8 @@ VERSION >= v"1.1" && @non_differentiable Sys.isopenbsd(::Symbol)
...
+@non_differentiable ones(::Any...)
+@non_differentiable zero(::Any...)
```
mzgubic, line 445:
```suggestion
@non_differentiable zero(::Any)
```

mzgubic (approved):

mzgubic:
Thanks for the PR! Will merge when CI passes. Just FYI, it's possible to commit the code suggestions all at once, which prevents CI from running for each suggestion. I think there is a limit to how many jobs can be run per organisation at the same time, and ChainRules tests take a while so a queue forms sometimes.

niklasschmitz:
Thanks @mzgubic !

PR #469, opened 2021-07-12
Support for qr decomposition pullback

rkube:
Added a `rrule` for the qr deomposition. @sethaxen

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -275,6 +275,101 @@ function svd_rev(USV::SVD, Ū, s̄, V̄)
 end
 
 #####
+##### `qr`
+#####
+
+
+function ChainRules.rrule(::typeof(getproperty), F::LinearAlgebra.QRCompactWY, d::Symbol)
+    function getproperty_qr_pullback(Ȳ)
+        # The QR factorization is calculated from `factors` and T, matrices stored in the QRCompactWYQ format, see
+        # R. Schreiber and C. van Loan, Sci. Stat. Comput. 10, 53-57 (1989).
+        # Instead of backpropagating Q̄ and R̄ through (factors)bar and T̄, we re-use factors to carry Q̄ and T to carry R̄
+        # in the Tangent object.
+        ∂T = d === :R ? Ȳ : nothing
+
+        ∂F = Tangent{LinearAlgebra.QRCompactWY}(; factors=∂factors, T=∂T)
+        return (NoTangent(), ∂F)
+    end
+
+    return getproperty(F, d), getproperty_qr_pullback
+end
+
+
+
+function ChainRules.rrule(::typeof(qr), A::AbstractMatrix{T}) where {T}
+    QR = qr(A)
+    m, n = size(A)
+    function qr_pullback(Ȳ::Tangent)
+        # For square (m=n) or tall and skinny (m >= n), use the rule derived by 
+        # Seeger et al. (2019) https://arxiv.org/pdf/1710.08717.pdf
+        #   
+        # Ā = [Q̄ + Q copyltu(M)] R⁻ᵀ
+        #   
+        # where copyltU(C) is the symmetric matrix generated from C by taking the lower triangle of the input and
+        # copying it to its upper triangle : copyltu(C)ᵢⱼ = C_{max(i,j), min(i,j)}
+        #   
+        # This code is re-used in the wide case and therefore in a separate function.
+
+        function qr_pullback_square_deep(Q̄, R̄, A, Q, R)
+            M = R*R̄' - Q̄'*Q
+            # M <- copyltu(M)
+            M = tril(M) + transpose(tril(M,-1))
+            Ā = (Q̄ + Q * M) / R'
+        end
+
+        # For the wide (m < n) case, we implement the rule derived by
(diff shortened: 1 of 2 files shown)


src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -274,6 +274,101 @@ function svd_rev(USV::SVD, Ū, s̄, V̄)
...
+
+function ChainRules.rrule(::typeof(getproperty), F::LinearAlgebra.QRCompactWY, d::Symbol)
+    function getproperty_qr_pullback(Ȳ)
+        # The QR factorization is calculated from `factors` and T, matrices stored in the QRCompactWYQ format, see
+        # R. Schreiber and C. van Loan, Sci. Stat. Comput. 10, 53-57 (1989).
+        # Instead of backpropagating Q̄ and R̄ through (factors)bar and T̄, we re-use factors to carry Q̄ and T to carry R̄
+        # in the Tangent object.
+        ∂T = d === :R ? Ȳ : nothing
+
+        ∂F = Tangent{LinearAlgebra.QRCompactWY}(; factors=∂factors, T=∂T)
```
oxinabox, line 290: `∂factors` isn't defined

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -274,6 +274,101 @@ function svd_rev(USV::SVD, Ū, s̄, V̄)
...
+#####
+
+
+function ChainRules.rrule(::typeof(getproperty), F::LinearAlgebra.QRCompactWY, d::Symbol)
+    function getproperty_qr_pullback(Ȳ)
+        # The QR factorization is calculated from `factors` and T, matrices stored in the QRCompactWYQ format, see
+        # R. Schreiber and C. van Loan, Sci. Stat. Comput. 10, 53-57 (1989).
+        # Instead of backpropagating Q̄ and R̄ through (factors)bar and T̄, we re-use factors to carry Q̄ and T to carry R̄
+        # in the Tangent object.
+        ∂T = d === :R ? Ȳ : nothing
```
oxinabox, line 288: We do not uses `nothing` to represent not used.
We use `ZeroTangent` for not used, (and `NoTangent` for not having a meaningful tangent space)

sethaxen:
> pokes @sethaxen
> (I can't really review this)

Yeah, sorry, I took the deep dive studying the various QR parameterizations a few weeks back in prep for reviewing this but haven't had the chance to yet. Sorry for the delay, @rkube!

sethaxen:
So this is a _really_ tricky set of rules to define, perhaps trickier than any of the other rules we have in ChainRules currently. Here are just a few complications:
- The signatures for `qr` are all changing with Julia v1.7 (below I use the 1.7 signatures)
- `qr` can produce 4 different types in the standard library, summarized below:
```julia
# returns QRCompactWY via LAPACK.geqrt!
qr(A::StridedMatrix{<:BlasFloat}, pivot = NoPivot(); kwargs...)
qr!(A::StridedMatrix{<:BlasFloat}, ::NoPivot; kwargs...)

# returns QR via qrfactUnblocked!
qr(A::AbstractMatrix, pivot = NoPivot())
qr!(A::AbstractMatrix, ::NoPivot)

# returns QRPivoted via qrfactPivotedUnblocked!
qr(A::AbstractMatrix, ::ColumnNorm)
qr!(A::AbstractMatrix, ::ColumnNorm)

# returns SuiteSparse.SPQR.QRSparse
qr(A::SparseMatrixCSC, pivot = NoPivot())
```
- None of the `QR` objects generate the `Q` matrix. Instead, they represent it in a compact form, where `factors` contains Householder reflectors in the strict lower trapezoid, and `R` in the upper trapezoid. Computing rules in terms of these compact elements is challenging, roughly as challenging as implementing the `qr` functions themselves. 
- Calling `.Q` on one of these factorizations produces a `AbstractQ <: AbstractMatrix` object that basically has all of the same fields. The `AbstractQ` objects are `AbstractMatrix`es, which means they by default hit all of our `AbstractMatrix` rules and therefore will end up with `AbstractMatrix` cotangents unless we write custom `rrule`s for every function one might call on a `QR` object. 
- The `AbstractQ` objects are _weird_. For an `nxk` matrix `A`, `size(qr(A).Q) == (n, n)`. However, `Q` also allowed to be multiplied by matrices with size `(k, m)`. So consider code like the following:
```julia
A = randn(10, 5)
Q, _ = qr(A)
v = randn(5)
w = randn(10)
y = Q*w + Q*v
@assert size(y) == (10,)
```
This is completely allowed, but note that the cotangent of `Q` will be `∂Q = ∂y * w' + ∂y * v'`. This adds two matrices of size `(10, 10)` and `(10, 5)`, respectively. This addition will be handled by the AD engine and will error, so it's necessary then to use `ProjectTo` to padd the `(10, 5)` matrix with zeros to make it `(10, 10)`, but this is _very_ wasteful when dealing with very tall matrices where one may never use its `(10, 10)` version. 

I don't think we can just address a subset of these complications one-at-a-time. Once we start adding rules, which will override AD systems' default behavior of differentiating through the `qr!` fallback (for operator-overloading ADs), then we will need to have more rules to make sure all of our rules compose nicely. I need to think more if there's a way that this can be handled without a tremendous amount of complication.

PR #471, opened 2021-07-19
Change argument order in InplaceableThunk

mzgubic:
Follow up to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/396

Could help with https://github.com/FluxML/Zygote.jl/issues/1030


oxinabox (approved):
Arguably some of these should change to do-block form.
But this is clearly an improvement on status quo
So :approved

PR #472, opened 2021-07-21
add rule for copy

CarloLucibello:
Fixes https://github.com/FluxML/Zygote.jl/issues/1037

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -146,14 +146,21 @@ end
 @scalar_rule x \ y (-(Ω / x), one(y) / x)
 
 function frule((_, ẏ), ::typeof(identity), x)
-    return (x, ẏ)
+    return x, ẏ
 end
 
 function rrule(::typeof(identity), x)
-    function identity_pullback(ȳ)
-        return (NoTangent(), ȳ)
-    end
-    return (x, identity_pullback)
+    identity_pullback(ȳ)= NoTangent(), ȳ
+    return x, identity_pullback
+end
+
+function frule((_, ẏ), ::typeof(copy), x)
+    return x, ẏ
+end
+
+function ChainRulesCore.rrule(::typeof(copy), x)
+    copy_pullback(ȳ) = NoTangent(), ȳ
+    return copy(x), copy_pullback
 end
 
 # rouding related,
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -146,20 +146,18 @@ end
 @scalar_rule x \ y (-(Ω / x), one(y) / x)
 
 function frule((_, ẏ), ::typeof(identity), x)
-    return x, ẏ
+    return (x, ẏ)
 end
 
 function rrule(::typeof(identity), x)
-    identity_pullback(ȳ)= NoTangent(), ȳ
-    return x, identity_pullback
-end
-
-function frule((_, ẏ), ::typeof(copy), x)
-    return x, ẏ
+    function identity_pullback(ȳ)
+        return (NoTangent(), ȳ)
+    end
+    return (x, identity_pullback)
 end
 
 function ChainRulesCore.rrule(::typeof(copy), x)
-    copy_pullback(ȳ) = NoTangent(), ȳ
+    copy_pullback(ȳ) = (NoTangent(), ȳ)
     return copy(x), copy_pullback
 end
 
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -125,6 +125,24 @@
         )
     end
 
+    @testset "copy" for T in (Float64, ComplexF64)
+        test_frule(copy, randn(T))
+        test_frule(copy, randn(T, 4))
+        test_frule(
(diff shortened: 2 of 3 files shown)


src/rulesets/Base/base.jl
```diff
@@ -146,14 +146,21 @@ end
...
 function rrule(::typeof(identity), x)
-    function identity_pullback(ȳ)
-        return (NoTangent(), ȳ)
-    end
-    return (x, identity_pullback)
+    identity_pullback(ȳ)= NoTangent(), ȳ
+    return x, identity_pullback
+end
+
+function frule((_, ẏ), ::typeof(copy), x)
```
mzgubic, line 157: why do we also want the frule?

test/rulesets/Base/base.jl
```diff
@@ -125,6 +125,24 @@
         )
     end
 
+    @testset "copy" for T in (Float64, ComplexF64)
+        test_frule(copy, randn(T))
+        test_frule(copy, randn(T, 4))
+        test_frule(
```
mzgubic, line 131: do these work for you? I get
```julia
julia> copy((1.0, ))
ERROR: MethodError: no method matching copy(::Tuple{Float64})
Closest candidates are:
  copy(::LinearAlgebra.CholeskyPivoted) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/cholesky.jl:440
  copy(::T) where T<:SHA.SHA3_CTX at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/SHA/src/types.jl:213
  copy(::LinearAlgebra.LQ) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lq.jl:113
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1
```

mzgubic (commented):
Thanks for the PR! Do you know why the Zygote integration is failing? Seems to be some higher order AD thing?

src/rulesets/Base/base.jl
CarloLucibello: I don't know, I was just mimicking identity, should I remove it?

test/rulesets/Base/base.jl
CarloLucibello: they don't actually. I'll add a test with dictionaries instead

src/rulesets/Base/base.jl
mzgubic: 🤷 maybe, no code is the best code :)

test/rulesets/Base/base.jl
CarloLucibello: I didn't know how to add more sophisticated tests so I just removed the tuple test

CarloLucibello:
no idea why Zygote is failing

test/rulesets/Base/base.jl
mzgubic: A simple test with dicts fails because `rand_tangent` is not defined on Dicts. I will open an issue

mcabbott:
Should this be here or in Zygote? i.e. is this a mathematical property of `copy` or something more specific?

Within Zygote, `copy(::Array)` could presumably be a no-op, since you can't mutate them. That's Zygote-specific. But `copy(::Buffer)` means something else... 

Finally, testing this by finite-differences doesn't achieve much. But testing that it does what you want on a `Dict` would at least ensure that an over-zealous cleanup doesn't (say) restrict this to `::AbstractArray`.

Not much of a review, but maybe things to think about.

CarloLucibello:
> Should this be here or in Zygote? i.e. is this a mathematical property of `copy` or something more specific?

I would say it is a convention of the language, the Base's docstring reads
```
    copy(x)

Create a shallow copy of x: the outer structure is copied, but not all
internal values. For example, copying an array produces a new array with
identically-same elements as the original.
```

> Within Zygote, `copy(::Array)` could presumably be a no-op, since you can't mutate them. That's Zygote-specific. But `copy(::Buffer)` means something else...
> 
`copy(::Buffer)` kind of breaks the convention above since it returns the underlying array. But maybe the rrule defined here is still reasonable since Buffer itself is an AbstactArray therefore it is ok it gets an array-like sensitivity?

> Finally, testing this by finite-differences doesn't achieve much. But testing that it does what you want on a `Dict` would at least ensure that an over-zealous cleanup doesn't (say) restrict this to `::AbstractArray`.
> 
> Not much of a review, but maybe things to think about.

test/rulesets/Base/base.jl
```diff
@@ -117,6 +117,15 @@
         test_rrule(identity, Tuple(randn(T, 3)))
     end
 
+    @testset "copy" for T in (Float64, ComplexF64)
+        test_rrule(copy, randn(T))
+        test_rrule(copy, randn(T, 4))
+    end
+            
+    @testset "copy dict" begin
```
mcabbott, line 125:
```suggestion
    @testset "copy" begin
        test_rrule(copy, randn(4))
```

PR #473, opened 2021-07-21
Stop using rand_tangent from CRTU

oxinabox:
Alt we could move it back to CRTU for 1.0
I think I would like that


oxinabox:
we moved it back

PR #478, opened 2021-07-26
don't try an Inplace unknown matrixes in *

oxinabox:
@AlexRobson this should fix your thing.
We basically need to be pretty careful not to use InplaceableThunks when we don't know that sensible things will be happening for the input.
So this PR basically just removed buggy code, and adds back a special-case that is known not buggy.
A downside is that some places were we could inplace we don't anymore. e.g. `*(::Diagonal, ::Diagonal)`
and `*::(StridedVector, StridedMatrix)`. Could add those back if we really wanted.
But since noone is really using the InplaceableThunk in anger, just holding off on that for now.
solving it generally is https://github.com/JuliaDiff/ChainRulesCore.jl/issues/411

I recommend Turning off showing whitespace when reveiwing

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -19,6 +19,7 @@ end
 ##### `*`
 #####
 
+
 function rrule(
     ::typeof(*),
     A::AbstractVecOrMat{<:CommutativeMulNumber},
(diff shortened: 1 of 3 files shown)


src/rulesets/Base/arraymath.jl
```diff
@@ -19,28 +19,49 @@ end
...
-            @thunk(project_B(A' * Ȳ)),
+            @thunk(A' * Ȳ),
         )
         return NoTangent(), dA, dB
     end
     return A * B, times_pullback
 end
 
+
+
```
oxinabox, line 64:
```suggestion
```

src/rulesets/Base/arraymath.jl
```diff
@@ -66,6 +87,9 @@ function rrule(
...
+
```
oxinabox, line 92:
```suggestion
```

src/rulesets/Base/arraymath.jl
```diff
@@ -19,28 +19,49 @@ end
...
 function rrule(
     ::typeof(*),
     A::AbstractVecOrMat{<:CommutativeMulNumber},
     B::AbstractVecOrMat{<:CommutativeMulNumber},
 )
     project_A = ProjectTo(A)
     project_B = ProjectTo(B)
+    function times_pullback(ȳ)
+        Ȳ = unthunk(ȳ)
+        dA = @thunk(project_A(Ȳ * B'))
```
AlexRobson, line 32: Is it work doing something like this here? 

```
maybe_mul!(A, B) = A * B
maybe_mul!(A::StridedArray, B::StridedArray) = (X = similar(B); mul!(X, A, B)) 
```

Or just push that discussion to https://github.com/JuliaDiff/ChainRulesCore.jl/issues/411

AlexRobson (commented):
Can confirm that this fixes my particular issue, and this would also appear in the test case that was added after some local testing. 

I only had one comment on the code so just to replay my understanding of it: Previously we had inplace `mul!` that was breaking some rrules (such as *(::Diagonal, ::StridedMatrix`). Now, we have an abstract fallback, but included a specialisation for strided arrays where can do this.

src/rulesets/Base/arraymath.jl
oxinabox: > Or just push that discussion to JuliaDiff/ChainRulesCore.jl#411

This

oxinabox:
> Looks good to me

I will take this as approval.

src/rulesets/Base/arraymath.jl
```diff
@@ -19,28 +19,49 @@ end
...
+        dB = @thunk(project_B(A' * Ȳ))
+        return NoTangent(), dA, dB
+    end
+    return A * B, times_pullback
+end
+
+# Optimized case for StridedMatrixes
+# no need to project as already dense, and we are allowed to use InplaceableThunk because
+# we know the destination will also be dense. TODO workout how to apply this generally:
```
mcabbott, line 41: No need to project for structure, but possibly for eltype, right?

oxinabox: this is true. we might run into a problem with  mistakenly complixfying a real matrix.

PR #463, opened 2021-07-01
Port a few rules from Diffractor

Keno:

PR #483, opened 2021-07-29
Add more non-differentiable builtins

Keno:
As reported in https://github.com/JuliaDiff/Diffractor.jl/issues/19


Keno:
If easier this can also be squashed into https://github.com/JuliaDiff/ChainRules.jl/pull/463.

mcabbott:
Xref #482, which put `isdefined` under Base not Core, perhaps that line ought to be deleted.

mzgubic (approved):

PR #484, opened 2021-07-29
Make construtors of all AbstractRNGs as nondifferentiable

oxinabox:
In general it is pretty much nonsense to be doing AD through the constructor of any random number generator.
Not just a `MersenneTwister`.
Julia introduces a new default RNG in 1.7

can we use `@nondifferentiable` for this now?

Probably wants tests.

Hopefully fixed https://github.com/SciML/DiffEqSensitivity.jl/runs/3191036433?check_suite_focus=true#step:6:272

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.3.0"
+version = "1.4.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
Yeah, doing it for `AbstractRNG` is cool! Surely all the rest of the `rand` `rand!` methods etc would capture any other methods used internally by packages defining RNGs

DhairyaLGandhi (approved):

DhairyaLGandhi:
Zygote doesn't know what the arguments represent, and will therefore have to go into any before it reaches `rand!` and then drop it while compiling the pullback.

ChrisRackauckas:
>In general it is pretty much nonsense to be doing AD through the constructor of any random number generator.

Oh boy, my next paper is going to be a trip 😅

oxinabox:
>    Oh boy, my next paper is going to be a trip sweat_smile

I mean leaving REINFORCE etc aside

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.2.0"
+version = "1.3.0"
```
oxinabox, line 3:
```suggestion
version = "1.4.0"
```

PR #485, opened 2021-07-30
fix some issues with rules for `^`

### LINKED ISSUE #26: Strange behaviour of the derivative
I do not know, maybe I am missing something simple here

```julia
using Diffractor
using Plots

function f1(x)
    res = 0.0
    for i in 1:10
        res += x^i
    end
    res
end

f(x) = let var"'" = Diffractor.PrimeDerivativeFwd 
    f1'(x)
end


xs = -0.5:0.01:0.5
plot(xs, f1.(xs), legend = nothing)
plot!(xs, f.(xs))
```

It has this weird glitch in zero.

flameshot-2021-07-30T12-23-07

Info:
```julia
julia> versioninfo()
Julia Version 1.8.0-DEV.74
Commit ba5fffca7b (2021-06-24 02:52 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.0 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 4
  JULIA_PKG_SERVER =

(Diffractor) pkg> st
      Status `~/Projects/BabySteps/Diffractor/Project.toml`
  [d360d2e6] ChainRulesCore v1.0.2
  [9f5e2b26] Diffractor v0.1.1 `~/.julia/dev/Diffractor`
  [91a5bcdd] Plots v1.19.4
```

simeonschaub:
should fix JuliaDiff/Diffractor.jl#26

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -168,11 +168,18 @@ end
 # note: rules for ^ are defined in the fastmath_able.jl
 function frule((_, _, Δx, _), ::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{p}) where p
     y = Base.literal_pow(^, x, pv)
-    return y, (p * y / x * Δx)
+    return y, ifelse(iszero(x), zero(y), p * y / x * Δx)
 end
+frule((_, _, Δx, _), ::typeof(Base.literal_pow), ::typeof(^), x::Real, ::Val{1}) = x^1, Δx
 
 function rrule(::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{p}) where p
     y = Base.literal_pow(^, x, pv)
-    literal_pow_pullback(dy) = NoTangent(), NoTangent(), (p * y / x * dy), NoTangent()
+    function literal_pow_pullback(dy)
+        return NoTangent(), NoTangent(), ifelse(iszero(x), zero(y), p * y / x * dy), NoTangent()
+    end
     return y, literal_pow_pullback
+end
+function rrule(::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{1})
+    literal_pow_one_pullback(dy) = NoTangent(), NoTangent(), dy, NoTangent()
+    return x^1, literal_pow_one_pullback
 end
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -164,14 +164,16 @@ let
         @scalar_rule x - y (true, -1)
         @scalar_rule x / y (one(x) / y, -(Ω / y))
         #log(complex(x)) is required so it gives correct complex answer for x<0
-        @scalar_rule(x ^ y,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(complex(x))),
-        )
+        @scalar_rule(x ^ y, (
+            ifelse(iszero(x), ifelse(isone(y), one(Ω), zero(Ω)), y * Ω / x),
+            Ω * log(complex(x)),
+        ))
         # x^y for x < 0 errors when y is not an integer, but then derivative wrt y
         # is undefined, so we adopt subgradient convention and set derivative to 0.
-        @scalar_rule(x::Real ^ y::Real,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x)))),
-        )
+        @scalar_rule(x::Real ^ y::Real, (
+            ifelse(iszero(x), ifelse(isone(y), one(Ω), zero(Ω)), y * Ω / x),
+            Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x))),
+        ))
         @scalar_rule(
             rem(x, y),
             @setup((u, nan) = promote(x / y, NaN16), isint = isinteger(x / y)),
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -173,5 +173,14 @@
         # for real x and n, x must be >0
         test_frule(Base.literal_pow, ^, 3.5, Val(3))
         test_rrule(Base.literal_pow, ^, 3.5, Val(3))
+
+        test_frule(Base.literal_pow, ^, 0.0, Val(3))
+        test_rrule(Base.literal_pow, ^, 0.0, Val(3))
+
+        test_frule(Base.literal_pow, ^, 3.5, Val(1))
+        test_rrule(Base.literal_pow, ^, 3.5, Val(1))
+
+        test_frule(Base.literal_pow, ^, 0.0, Val(1))
+        test_rrule(Base.literal_pow, ^, 0.0, Val(1))
     end
 end
--- a/test/rulesets/Base/fastmath_able.jl
+++ b/test/rulesets/Base/fastmath_able.jl
@@ -182,6 +182,15 @@ const FASTABLE_AST = quote
                 @test ∂y ≈ 0
             end
         end
+
+        @testset "edge cases with ^" begin
+            # FIXME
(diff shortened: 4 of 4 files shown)


src/rulesets/Base/base.jl
```diff
@@ -168,11 +168,18 @@ end
 # note: rules for ^ are defined in the fastmath_able.jl
 function frule((_, _, Δx, _), ::typeof(Base.literal_pow), ::typeof(^), x::Real, pv::Val{p}) where p
     y = Base.literal_pow(^, x, pv)
-    return y, (p * y / x * Δx)
+    return y, ifelse(iszero(x), zero(y), p * y / x * Δx)
```
sethaxen, line 171: This could cause type-instability if `Δx` is not the same type as `x`. Since the point of `y/x` is to avoid computing `x^(p-1)`, perhaps something like this? The additional `oftype` also ensures that if `x`, `Δx`, and `p` are all integers, we don't unnecessarily make everything a float.
```suggestion
    xpm1 = y / x
    return y, p * oftype(y, ifelse(iszero(x), zero(xpm1), xpm1)) * Δx
```
What do you think?

src/rulesets/Base/fastmath_able.jl
```diff
@@ -164,14 +164,16 @@ let
...
+        ))
         # x^y for x < 0 errors when y is not an integer, but then derivative wrt y
         # is undefined, so we adopt subgradient convention and set derivative to 0.
-        @scalar_rule(x::Real ^ y::Real,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x)))),
-        )
+        @scalar_rule(x::Real ^ y::Real, (
+            ifelse(iszero(x), ifelse(isone(y), one(Ω), zero(Ω)), y * Ω / x),
+            Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x))),
+        ))
```
mcabbott, line 176: Is this right? Applied to the example from https://github.com/FluxML/Zygote.jl/issues/1036 it still gives zero gradient:
```
julia> oldrule(0.0, 0.9)
(0.0, 0.0)

julia> newrule(0.0, 0.9)
(0.0, 0.0)

julia> oldrule(0.0 + eps(), 0.9)
(4.0532396646334465e15, -36.04365338911715)

julia> newrule(0.0 + eps(), 0.9)
(4.0532396646334465e15, -36.04365338911715)
```

simeonschaub: What is `newrule` and `oldrule` here?

mcabbott: Edited, sorry. I just copy-pasted from the above diff, to quickly try things.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -164,14 +164,16 @@ let
         @scalar_rule x - y (true, -1)
         @scalar_rule x / y (one(x) / y, -(Ω / y))
         #log(complex(x)) is required so it gives correct complex answer for x<0
-        @scalar_rule(x ^ y,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(complex(x))),
-        )
+        @scalar_rule(x ^ y, (
```
sethaxen, line 167: This  pre-existing rule is unsafe for non-real and complex numbers. e.g. this will fail for quaternions because you can't call `complex` on them and because it assumes commutativity of multiplication. I don't know if part of this PR, but the types should probably be restricted.

src/rulesets/Base/fastmath_able.jl
simeonschaub: I don't think this is necessarily well-defined, since the derivative of `x ^ (1 - ε)` should be `(1 - ε) / x ^ ε`, so the limit as `x -> 0` diverges.

mcabbott: Sure. But giving some large positive number seems more useful than an error, or NaN, and silently giving zero seems just wrong. 

E.g. if you have this in some optimisation problem, and it would lower your loss to move away from the boundary x=0 to x=0.1, then getting zero gradient there is misleading.

simeonschaub: So should this be `Inf` instead?

src/rulesets/Base/base.jl
sethaxen: Another alternative:
```suggestion
xpm1 = if y isa Integer
    iszero(x) ? zero(y) : div(y, x)
else
    iszero(x) ? zero(y) : y / x
end
return y, p * xpm1 * Δx
```
I'm not certain how the use of the ternary vs `ifelse` would impact performance.

src/rulesets/Base/fastmath_able.jl
mcabbott: Maybe. As I just wrote in parallel here https://github.com/FluxML/Zygote.jl/issues/1036#issuecomment-889832592, maybe it would be better to regularise everything, for x=0 (in this one-sided case) always give the gradients for say x = eps(). This will change smoothly between the different cases.

sethaxen: The `Ω / x` term is just `x^(y-1)`, which for `x=0` is 0 for all `y>1`, 1 for `y=1` and `Inf` for all `y<1`. I like the `eps` idea. It also smooths the approach to `0^0==1`. Not certain all the consequences it could have though.

src/rulesets/Base/fastmath_able.jl
```diff
@@ -164,14 +164,16 @@ let
...
+            ifelse(iszero(x), ifelse(isone(y), one(Ω), zero(Ω)), y * Ω / x),
+            Ω * log(complex(x)),
+        ))
         # x^y for x < 0 errors when y is not an integer, but then derivative wrt y
         # is undefined, so we adopt subgradient convention and set derivative to 0.
-        @scalar_rule(x::Real ^ y::Real,
-            (ifelse(iszero(x), zero(Ω), y * Ω / x), Ω * log(oftype(Ω, ifelse(x ≤ 0, one(x), x)))),
-        )
+        @scalar_rule(x::Real ^ y::Real, (
+            ifelse(iszero(x), ifelse(isone(y), one(Ω), zero(Ω)), y * Ω / x),
```
Keno, line 174: Are we sure this is the best way to compute this? Division is quite expensive. Why not the more straightforward:
```
function rrule(::typeof(^), x::Real, y::Real)
    r = x^(y-1)
    r*x, Δ->(..., r, ...)
end
```

sethaxen: > Are we sure this is the best way to compute this?

If I had to guess, it's so we can use `@scalar_rule` and avoid defining both the `frule` and `rrule`, but I agree that this approach is probably better. For the cases I can think of where this doesn't work (e.g. `x<0` and `y=0.5`), the primal function would itself error.

oxinabox:
I will leave this to @sethaxen to be the person to approve this.
Keno has merge rights so once approved can be merged when ready.
I am unsubscribing from the PR, ping me if i am needed.

src/rulesets/Base/fastmath_able.jl
mcabbott: FWIW, I started trying to do that (never dividing) on this branch:

https://github.com/JuliaDiff/ChainRules.jl/compare/master...mcabbott:fix_pow_mine?expand=1

Built on top of this PR, could perhaps push here, or make a separate PR... but still needs many tests & some details perhaps to think about.

simeonschaub: Sorry, I lost track of this PR a little. If you want to take this over, feel free to either push to this branch or open a new PR. Otherwise, I will try to get back to this some time this week.

mcabbott: Sure, either way. I have a lot of `NaN`s still... some of them marked broken tests now, at least.

PR #487, opened 2021-07-30
Add a few rules needed for higher-order AD of ProjectTo

Keno:
The ProjectTo constructors use a few functions/builtins
that did not previously have rules breaking higher order
AD, since these constructors are now called by the first
order rules.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -26,6 +26,13 @@ function rrule(::typeof(real), z::Number)
     return (real(z), real_pullback)
 end
 
+# Conversions to Float
+
+@scalar_rule float(x) true
+@scalar_rule Float64(x::Real) true
+@scalar_rule Float32(x::Real) true
+@scalar_rule AbstractFloat(x::Real) true
+
 # `imag`
 
 @scalar_rule imag(x::Real) ZeroTangent()
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -174,4 +174,21 @@
         test_frule(Base.literal_pow, ^, 3.5, Val(3))
         test_rrule(Base.literal_pow, ^, 3.5, Val(3))
     end
+
+    @testset "Float conversions" begin
+        for f in (float, Float32, Float64, AbstractFloat)
+            test_frule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+            test_frule(f, 1⊢1.7; rtol=1.0e-3, atol=1.0e-3)
+            test_rrule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+
+            # test_rrule doesn't like integers, so test that case manually
+            let y = rand(Int), x = randn()
+                @test rrule(f, y)[2](x)[2] == x
+            end
+        end
+
+        # Make sure that we didn't accidentally define a rule for all DataTypes
+        @test frule(NoRules, 1.0) === nothing
+        @test rrule(NoRules, 1.0) === nothing
+    end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.4.0"
+version = "1.5.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -10,7 +10,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-ChainRulesCore = "1"
+ChainRulesCore = "1.1"
 ChainRulesTestUtils = "1"
 Compat = "3.31"
 FiniteDifferences = "0.12.8"
--- a/src/rulesets/Core/core.jl
+++ b/src/rulesets/Core/core.jl
@@ -10,6 +10,12 @@
 @non_differentiable Core.apply_type(::Any, ::Any...)
 @non_differentiable Core.typeof(::Any)
 
+if isdefined(Core, :_typevar)
+    @non_differentiable Core._typevar(::Any...)
+end
+@non_differentiable TypeVar(::Any...)
+@non_differentiable UnionAll(::Any, ::Any)
+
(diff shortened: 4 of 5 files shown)


src/rulesets/Base/base.jl
```diff
@@ -26,6 +26,12 @@ function rrule(::typeof(real), z::Number)
     return (real(z), real_pullback)
 end
 
+# Conversions to Float
+
+@scalar_rule float(x::Real) true
+@scalar_rule Float64(x::Real) true
+@scalar_rule Float32(x::Real) true
```
oxinabox, line 33: Shall we do:
```suggestion
@scalar_rule Float32(x::Real) true
@scalar_rule Float16(x::Real) true
@scalar_rule AbstractFloat(x::Real) true
```

src/rulesets/Base/base.jl
```diff
@@ -26,6 +26,12 @@ function rrule(::typeof(real), z::Number)
     return (real(z), real_pullback)
 end
 
+# Conversions to Float
+
+@scalar_rule float(x::Real) true
```
oxinabox, line 31: This is also true for complex i think
```suggestion
@scalar_rule float(x::Union{Real,Complex}) true
```

src/rulesets/Base/base.jl
```diff
@@ -26,6 +26,12 @@ function rrule(::typeof(real), z::Number)
     return (real(z), real_pullback)
 end
 
+# Conversions to Float
```
oxinabox, line 29: These need tests 
Should be pretty simple
`test_scalar(f, 1.2)`
and maybe
`test_scalar(f, 1⊢1.7)`

for each of these

oxinabox:
Once it has tests, merge, bump version and tag when happy

oxinabox (approved):

src/rulesets/Base/base.jl
sethaxen: Is there a reason it needs a type constraint? e.g. if `float` can be called on the primal, then isn't it always safe for us to say the derivative is one? If it can't be called, the frule/rrule will error anyways.

oxinabox: true

test/rulesets/Base/base.jl
```diff
@@ -174,4 +174,21 @@
...
+
+            # test_rrule doesn't like integers, so test that case manually
+            let y = rand(Int), x = randn()
+                @test rrule(f, y)[2](x)[2] == x
+            end
+        end
+
+        # Make sure that we didn't accidentally define a rule for all DataTypes
+        @test frule(NoRules, 1.0) === nothing
+        @test rrule(NoRules, 1.0) === nothing
```
oxinabox, line 192: I think this should be removed from this PR.
And a seperate PR should add just this, as its own testset in its own file

test/rulesets/Base/base.jl
```diff
@@ -174,4 +174,21 @@
...
+
+    @testset "Float conversions" begin
+        for f in (float, Float32, Float64, AbstractFloat)
+            test_frule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+            test_frule(f, 1⊢1.7; rtol=1.0e-3, atol=1.0e-3)
+            test_rrule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+
+            # test_rrule doesn't like integers, so test that case manually
+            let y = rand(Int), x = randn()
+                @test rrule(f, y)[2](x)[2] == x
```
oxinabox, line 186: Really? seems bad that it can't handle it even with a `\vdash`
Can you open a issue on CRTU?

test/rulesets/Base/base.jl
Keno: Maybe, but it's the rules in this PR that accidentally define it if you're not careful.

test/rulesets/Base/base.jl
```diff
@@ -174,4 +174,21 @@
...
     end
+
+    @testset "Float conversions" begin
+        for f in (float, Float32, Float64, AbstractFloat)
+            test_frule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+            test_frule(f, 1⊢1.7; rtol=1.0e-3, atol=1.0e-3)
+            test_rrule(f, 1.2; rtol=1.0e-3, atol=1.0e-3)
+
+            # test_rrule doesn't like integers, so test that case manually
+            let y = rand(Int), x = randn()
```
oxinabox, line 185: WHy are bothinging with this rand stuff?
Why not
```julia
@test rrule(f, 9)[2](5.1)[2] == 5.1
```

oxinabox (approved):
merge when happy

test/rulesets/Base/base.jl
oxinabox: Correct.
I think we should have it, just not here

test/rulesets/Base/base.jl
mzgubic: might just need updating to CRTU 1.0?

test/rulesets/Base/base.jl
Keno: Alright. I would prefer to merge as is though, so we can fix Diffractor and then move things around in a follow up.

test/rulesets/Base/base.jl
Keno: That'd be fine. I just like using random variables. It's like the poor man's symbolics.

test/rulesets/Base/base.jl
Keno:
```
test_rrule: Float64 on Int64: Error During Test at /home/keno/.julia/packages/ChainRulesTestUtils/8380y/src/testers.jl:191
  Got exception outside of a @test
  InexactError: Int64(0.99)
  Stacktrace:
    [1] Int64
      @ ./float.jl:812 [inlined]
    [2] convert
      @ ./number.jl:7 [inlined]
    [3] setindex!
      @ ./array.jl:937 [inlined]
    [4] (::FiniteDifferences.var"#62#64"{Int64, ComposedFunction{ComposedFunction{ComposedFunction{typeof(first), typeof(FiniteDifferences.to_vec)}, ChainRulesTestUtils.var"#fnew#42"{ChainRulesTestUtils.var"#call#50"{NamedTuple{(), Tuple{}}}, Tuple{DataType, Int64}, Tuple{Bool, Bool}}}, FiniteDifferences.var"#Real_from_vec#20"}, Vector{Int64}})(ε::Float64)
      @ FiniteDifferences ~/.julia/packages/FiniteDifferences/W3rQO/src/grad.jl:18
```
on CRTU 1.0

PR #491, opened 2021-08-01
Some Array rules

willtebbutt:


test/rulesets/Base/array.jl
```diff
@@ -1,3 +1,14 @@
+@testset "constructors" begin
+    @testset "undef" begin
+        test_rrule(Array{Float64, 1}, undef, 5)
+        test_rrule(Array{Float32, 3}, undef, 5, 4, 3)
+    end
+    @testset "from existing array" begin
+        test_rrule(Array, randn(2, 5))
+        test_rrule(Array, Diagonal(randn(5)))
+    end
```
mcabbott, line 9: I haven't tried these, but they appear to be allowed by the rule:
```suggestion
        test_rrule(Matrix, Diagonal(randn(5)))
        test_rrule(Matrix, transpose(randn(4)))
        test_rrule(Array{ComplexF64}, randn(3))
    end
```

willtebbutt: Seems reasonable -- thanks for the extra tests.

mcabbott (approved):

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,40 @@
...
+##### `vect`
+#####
+
+@non_differentiable Base.vect()
+
+# Don't worry about projection here. The data passes straight through, so if a cotangent has
+# the wrong type for some reason, it must be the fault of another rule somewhere.
+function rrule(::typeof(Base.vect), X...)
+    function vect_pullback(ȳ)
+        X̄ = ntuple(n -> ȳ[n], length(X))
```
nickrobinson251, line 23: Move `N = length(X)` outside to avoid closing over X?

willtebbutt: Good idea.

willtebbutt:
@mcabbott @nickrobinson251 I've constrained the `vect` implementation a bit, because I couldn't figure out how to implement the most generic (and most useful) pullback. I've opened #492 to discuss further.

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,32 @@
...
+
+@non_differentiable Base.vect()
+
+# Don't worry about projection here. The data passes straight through, so if a cotangent has
+# the wrong type for some reason, it must be the fault of another rule somewhere.
+function rrule(::typeof(Base.vect), X::T...) where {T}
+    l = length(X)
+    function vect_pullback(ȳ)
+        X̄ = ntuple(n -> ȳ[n], l)
+        return (NoTangent(), X̄...)
```
mcabbott, line 25: I think this will infer a bit better:
```suggestion
function rrule(::typeof(Base.vect), x::Vararg{T,N}) where {T,N}
    function vect_pullback(ȳ)
        return (NoTangent(), NTuple{N}(ȳ)...)
```
Trying the existing one at the REPL, it forgets that the length is a constant:
```
julia> @code_warntype rrule(Base.vect, 1,2,3)[2]([1,2,3])
MethodInstance for (::var"#vect_pullback#17"{Int64})(::Vector{Int64})
  from (::var"#vect_pullback#17")(ȳ) in Main at REPL[14]:3
Arguments
  #self#::var"#vect_pullback#17"{Int64}
  ȳ::Vector{Int64}
Locals
  #16::var"#16#18"{Vector{Int64}}
  X̄::Tuple{Vararg{Int64}}
Body::Tuple{NoTangent, Vararg{Int64}}
```

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,32 @@
...
+
+#####
+##### `vect`
+#####
+
+@non_differentiable Base.vect()
+
+# Don't worry about projection here. The data passes straight through, so if a cotangent has
+# the wrong type for some reason, it must be the fault of another rule somewhere.
+function rrule(::typeof(Base.vect), X::T...) where {T}
```
nickrobinson251, line 21: Why the `where T` here? Is it about avoiding promotion? Of so perhaps worth adding a comment?

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,32 @@
...
+    l = length(X)
```
nickrobinson251, line 22: Apparent I was wrong about moving this out

https://github.com/JuliaDiff/ChainRules.jl/issues/492#issuecomment-890586186

🤯

test/rulesets/Base/array.jl
```diff
@@ -1,3 +1,34 @@
+@testset "constructors" begin
+
+    # We can't use test_rrule here (as it's currently implemented) because the elements of
```
nickrobinson251, line 3: Perhaps open an issue and link it here?
I'm not sure if it's worth addressing somehow but maybe worth discussing it

src/rulesets/Base/array.jl
nickrobinson251: I guess not because, T could be an abstract type and we test on non-homogenous inputs

Isn't #492 still an issue here?

mcabbott: I thought it was there to enforce that this is the case of uniform types only. 
```
julia> ff(xs::T...) where {T} = "ok";

julia> ff(1, 3.4)
ERROR: MethodError: no method matching ff(::Int64, ::Float64)
```

nickrobinson251: That's what I thought too, but then I'm confused about the "inhomogeneous" test

src/rulesets/Base/array.jl
willtebbutt: Haha yup, I noticed that too. I love working with lots of people.

src/rulesets/Base/array.jl
nickrobinson251: Oh that test fails with a MethodError, so that resolves my confusion 😆

willtebbutt: Good point, but I don't think it does. This is, I believe, equivalent do doing diagonal dispatch, so everything has to be of the same concrete type for this method to be dispatched to (I had to double check though).

For example
```julia
function foo(X::T...) where {T}
    return X
end
foo(5.0, 4)
```
yields
```julia
ERROR: MethodError: no method matching foo(::Float64, ::Int64)
Closest candidates are:
  foo(::T...) where T at REPL[10]:1
Stacktrace:
 [1] top-level scope
   @ REPL[13]:1
```

src/rulesets/Base/array.jl
willtebbutt: This is more elegant than the equivalent thing I went for -- will adopt this.

src/rulesets/Base/array.jl
willtebbutt: I should have refreshed the page before posting 😂

src/rulesets/Base/array.jl
```diff
@@ -18,13 +18,25 @@ end
...
     function vect_pullback(ȳ)
+        X̄ = ntuple(n -> ȳ[n], N)
         return (NoTangent(), NTuple{N}(ȳ)...)
     end
     return Base.vect(X...), vect_pullback
 end
 
+# Numbers need to be projected because they don't pass straight through the function.
+# More generally, we would ideally project everything.
+function rrule(::typeof(Base.vect), X::Vararg{Number, N}) where {N}
```
mcabbott, line 31: Should be safe to project arrays too, and mixtures (if anyone does that):
```suggestion
function rrule(::typeof(Base.vect), X::Vararg{Union{Number,AbstractArray{<:Number}}, N}) where {N}
```

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,42 @@
...
+
+@non_differentiable Base.vect()
+
+# Don't worry about projection here. The data passes straight through, so if a cotangent has
+# the wrong type for some reason, it must be the fault of another rule somewhere.
+function rrule(::typeof(Base.vect), X::Vararg{T, N}) where {T, N}
+    function vect_pullback(ȳ)
+        X̄ = ntuple(n -> ȳ[n], N)
+        return (NoTangent(), NTuple{N}(ȳ)...)
+    end
```
mcabbott, line 25: This makes the same tuple twice, could just be:
```suggestion
    vect_pullback(ȳ) = (NoTangent(), NTuple{N}(ȳ)...)
```

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,42 @@
...
+end
+
+#####
+##### `vect`
+#####
+
+@non_differentiable Base.vect()
+
+# Don't worry about projection here. The data passes straight through, so if a cotangent has
+# the wrong type for some reason, it must be the fault of another rule somewhere.
```
mcabbott, line 20:
```suggestion
# Fallback case -- because ProjectTo isn't safe on arbitrary types, we just assume they are correct
```

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,42 @@
...
+function rrule(::typeof(Base.vect), X::Vararg{T, N}) where {T, N}
+    function vect_pullback(ȳ)
+        X̄ = ntuple(n -> ȳ[n], N)
+        return (NoTangent(), NTuple{N}(ȳ)...)
+    end
+    return Base.vect(X...), vect_pullback
+end
+
+# Numbers need to be projected because they don't pass straight through the function.
+# More generally, we would ideally project everything.
```
mcabbott, line 30:
```suggestion
# Numbers and arrays are often promoted, to make a uniform vector; ProjectTo here reverses this
```

src/rulesets/Base/array.jl
willtebbutt: I'm not sure that this quite captures the sentiment I was after -- this method is really for the special case associated with this method (diagonal dispatch case), as opposed to being a generic fallback: https://github.com/JuliaLang/julia/blob/bdacfa21ef7d9657be0ff3220947ceca92a340a1/base/array.jl#L126

mcabbott: Oh right, sorry. I mis-read and somehow assumed `X::Vararg{T` didn't impose this. So my comment is completely wrong.

Perhaps this would help me not mess it up next time, and is shorter:
```suggestion
# Case of uniform type `T`: the data passes straight through, so no projection should be required.
```

willtebbutt: Sounds good.

willtebbutt:
Seems like all comments have now been resolved -- @mcabbott I've requested a re-review because a lot has changed since your original approval.

src/rulesets/Base/array.jl
```diff
@@ -1,3 +1,42 @@
...
+
+# Numbers and arrays are often promoted, to make a uniform vector.
+# ProjectTo here reverses this
+function rrule(
+    ::typeof(Base.vect),
+    X::Vararg{Union{Number,AbstractArray{<:Number}}, N},
+) where {N}
+    projects = map(ProjectTo, X)
+    function vect_pullback(ȳ)
+        X̄ = ntuple(n -> projects[n](ȳ[n]), N)
```
mcabbott, line 34: I was wondering whether `projects[n]` gets handled well. `@code_warntype` seems happy... and my attempts to make things easier to unroll all make it slower:
```
julia> @btime rrule(Base.vect, 1,2,3)[2]($(rand(3)))
  28.684 ns (1 allocation: 112 bytes)
(NoTangent(), 0.7437540971290453, 0.6835525631785602, 0.29678387383966687)

julia> @btime rrule(Base.vect, 1+im,2+im,3+im)[2]($(rand(3)))
  235.140 ns (6 allocations: 416 bytes)
(NoTangent(), 0.9212083670665245 + 0.0im, 0.989459216141123 + 0.0im, 0.8454719840778347 + 0.0im)

julia> @btime rrule(Base.vect, 1, 2+3im, 4.0)[2]($(rand(3)))
  609.760 ns (6 allocations: 320 bytes)
(NoTangent(), 0.2914057312235363, 0.23309219863512798 + 0.0im, 0.08023319383991401)

julia> struct StaticGetter{i} end; @inline (::StaticGetter{i})(v) where {i} = v[i]; # from Zygote

julia> function rrule(
            ::typeof(Base.vect),
            X::Vararg{Union{Number,AbstractArray{<:Number}}, N},
        ) where {N}
            valN = Val(N)
            projects = map(ProjectTo, X)
            function vect_pullback(ȳ)
                X̄ = ntuple(n -> StaticGetter{n}()(projects)(ȳ[n]), valN)
                return (NoTangent(), X̄...)
            end
            return Base.vect(X...), vect_pullback
        end
rrule (generic function with 723 methods)

julia> @btime rrule(Base.vect, 1, 2+3im, 4.0)[2]($(rand(3)))
  1.442 μs (11 allocations: 448 bytes)
(NoTangent(), 0.7692288886268137, 0.39993377443044065 + 0.0im, 0.6341039234276757)
```

mcabbott (approved):

willtebbutt:
Added one additional test -- will merge once CI passes.

PR #493, opened 2021-08-02
Less clever sum rrule for the default case

Keno:
The sum rrule relies on broadcasting to figure out the
result shape of the cotangent. However, this has two
disadvantages:

1. We need to keep the original array around
2. Broadcasting machinery is complicated and tough on (higher-order) AD

This adds a special case for `dims=:`, which simply stores
the dimensions of the original array and uses `fill`
in the pullback, which has a simple rrule and is thus
much easier to AD.

### DIFF
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -6,7 +6,35 @@ function frule((_, ẋ), ::typeof(sum), x; dims=:)
     return sum(x; dims=dims), sum(ẋ; dims=dims)
 end
 
-function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
+# Internal helper for filling while maintaining array type.
+# TODO: Ideallty we'd only need typeof(x) here, but Base doesn't have the
+# interfaces for that.
+function _typed_fill(x, ȳ, axes)
+    fill!(similar(x, typeof(ȳ), axes), ȳ)
+end
+
+function rrule(::typeof(_typed_fill), x, ȳ, axes)
+    function _typed_fill_pullback(Ȳ)
+        return (NoTangent(), NoTangent(), sum(Ȳ), NoTangent())
+    end
+    return _typed_fill(x, ȳ, axes), _typed_fill_pullback
+end
+
+function sum_rrule(x, dims::Colon)
+    y = sum(x; dims=dims)
+    let xdims=size(x)
+        function sum_pullback(ȳ)
+            x̄ = InplaceableThunk(
+                x -> x .+= ȳ,
+                @thunk(_typed_fill(x, ȳ, xdims...)),
+            )
+            return (NoTangent(), x̄)
+        end
+        y, sum_pullback
+    end
+end
+
+function sum_rrule(x, dims)
     y = sum(x; dims=dims)
     function sum_pullback(ȳ)
         # broadcasting the two works out the size no-matter `dims`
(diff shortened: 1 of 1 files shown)


src/rulesets/Base/mapreduce.jl
```diff
@@ -6,7 +6,21 @@ function frule((_, ẋ), ::typeof(sum), x; dims=:)
...
 end
 
-function rrule(::typeof(sum), x::AbstractArray{T}; dims=:) where {T<:Number}
+function sum_rrule(x, dims::Colon)
+    y = sum(x; dims=dims)
+    let xdims=size(x)
+        function sum_pullback(ȳ)
+            x̄ = InplaceableThunk(
+                x -> x .+= ȳ,
+                @thunk(fill(ȳ, xdims...)),
```
mcabbott, line 15: Can I suggest using `axes`?
```suggestion
    y = sum(x)
    let x_axes = axes(x)
        function sum_pullback(ȳ)
            x̄ = InplaceableThunk(
                x -> x .+= ȳ,
                @thunk(fill(ȳ, x_axes)),
```

oxinabox:
(I will leave this PR to @mcabbott to review and approve. I have unsubscribed. Ping me if I am wanted)

mcabbott:
The downside of `fill` is that it's not going to preserve a CuArray. If the type is the same you can do `fill!(similar(typeof(x), axes(x)), dy)`, but CR allows for the eltype to vary, and then you need `x` for `fill!(similar(x, typeof(dy), axes(x)), dy)` I think.

With this `ProjectTo` story, HWFloats like `x::AbstractArray{Float32}` imply that the eltype is fixed -- this is known in the forward pass, so could in principle have a happy path. But a trait or something to tell you this isn't really set up yet.

Keno:
Well, we can't use `fill!` in the actual rule because Diffractor doesn't like mutation, but we can define a helper function that does that and just give it an appropriate rrule. In theory there should have been a fill-like array constructor, but https://github.com/JuliaLang/julia/issues/24595 was never finished.

mcabbott:
It would certainly be easy to make a `broadfill(x, dy)` which does what the rule now does, and has its own gradient. Still closes over `x`.

For the scalar case, it does seem a little wasteful to make a constant dense array at all. Zygote presently makes a `Fill` though it avoids this for CuArrays. Ideally this would work everywhere, right?
```
julia> cu(ones(5))' .+ Fill(3f0, 2) .+ 6
2×5 CuArray{Float32, 2}:
 10.0  10.0  10.0  10.0  10.0
 10.0  10.0  10.0  10.0  10.0
```

mcabbott:
Here's one attempt to write this. The fill-like function needn't get axes/size as well as `x`, but if it gets `dims` then it can cater to both the complete sum and the partial one. I think this ought to handle arrays of arrays, structured matrices, CuArrays, etc, but haven't tested much.
```julia
function rrule(::typeof(sum), x::AbstractArray; dims=:)
     function sum_pullback(dy_raw)
        dy = unthunk(dy_raw)
        # Here Ref allows for x being an array of arrays, could skip when eltype(x) <: Number
        dy_cast = dims isa Colon ? Ref(dy) : dy
        x_thunk = InplaceableThunk(
            dx -> dx .+= dy_cast
            @thunk _unsum(x, dy_cast, dims)
        )
        return (NoTangent(), x_thunk)
    end
    return sum(x; dims=dims), sum_pullback
end

_unsum(shape, value, dims) = ProjectTo(shape)(broadcast(last∘tuple, shape, value))

function rrule(::typeof(_unsum), shape, value, dims)
    _unsum_pullback(dz) = (NoTangent(), NoTangent(), sum(unthunk(dz); dims=dims), NoTangent())
    return _unsum(shape, value, dims), _unsum_pullback
end
```

Keno:
That rule may be the best we can do for the moment. Wanna take over this PR?

mcabbott:
Sure. Can I push to this branch, or better to call it a new PR from my fork? 

Will open another thread to discuss ways around capturing etc, afterwards.

Keno:
Either way, you can also just change the branch of this PR.

mcabbott:
Ok, I went the familiar route, and #494 is the new PR.

PR #495, opened 2021-08-05
fix test

mzgubic:
found while doing https://github.com/JuliaDiff/FiniteDifferences.jl/pull/189

no version bump as touches tests only


oxinabox:
the old rule looks right to me.

mzgubic:
there is nothing wrong with the test. there is something wrong with me

Issue #492, opened 2021-08-01
How to implement `vect`


willtebbutt:
#491 implements `vect` for two special cases, but neglects the general case because I couldn't figure out how to implement it in a type-stable manner for `Number`s, or at all in general.

My attempt at an implementation for `Number`s was
```julia
# Numbers need to be projected because they don't pass straight through the function.
# More generally, we would ideally project everything.
function rrule(::typeof(Base.vect), X::Vararg{Number, N}) where {N}
    l = length(X)
    projects = map(ProjectTo, X)
    function vect_pullback(ȳ)
        X̄ = ntuple(n -> projects[n](ȳ[n]), l)
        return (NoTangent(), X̄...)
    end
    return Base.vect(X...), vect_pullback
end
```
but it's type-unstable because the type of `X\bar` couldn't be inferred (Julia 1.6.2). I'm not entirely sure why it can't figure it out.

The `Number` specialisation of this function is interesting because it highlights that when an array is constructed in `vect`, the type of the elements changes from their original types some of the time but not others. For example, this isn't the case with subtypes of `AbstractArray`, which essentially pass through unchanged. I'm not sure how to do with this in general, because
1. there's not a function to explicitly hook in to which implements the conversion
2. `ProjectTo` isn't defined for all types (only `Number`s and `AbstractArray`s AFAICT), so we can't rely on it in general.

Any thoughts on either of these issues?

simeonschaub:
In this case I think you want to move the `l = length(X)` into the pullback, that way the length should be able to be constant-propagated.

mcabbott:
Just using `N` instead of `l` seems to solve it, on 1.7 at least. You could also close over `valN = Val(N)` which sometimes behaves better.

willtebbutt:
Good catch, thanks @simeonschaub . I've actually just gone down the `Vararg{T, N} where {T, N}` route, which seems to solve the problem.

willtebbutt:
> Just using N instead of l seems to solve it, on 1.7 at least. You could also close over valN = Val(N) which sometimes behaves better.

Haha you just beat me to it!

simeonschaub:
> Just using N instead of l seems to solve it, on 1.7 at least.

Oh, nice! I actually wasn't aware that we special-cased closures over static parameters like this.

mcabbott:
Heh. Arrays also sometimes get promoted, e.g.:
```julia
julia> [[1,2], [3,4.0+im]]
[...]

PR #496, opened 2021-08-10
More vect

willtebbutt:
Finishes up some extra stuff from #492 -- as per @mcabbott 's suggestion, I've just added a generic method.

### DIFF
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -37,6 +37,14 @@ function rrule(
     return Base.vect(X...), vect_pullback
 end
 
+# This rule isn't always correct, but Zygote and Diffractor fall over completely without it.
+function rrule(::typeof(Base.vect), X...)
+    vect_pullback(ȳ) = (NoTangent(), ȳ...)
+    return Base.vect(X...), vect_pullback
+end
+
+
+
 #####
 ##### `reshape`
 #####
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -37,10 +37,13 @@ function rrule(
     return Base.vect(X...), vect_pullback
 end
 
-# This rule isn't certain to be correct, if someone has overloaded `Base.vect` for their type
-# in a way that has a non-identity derivative; but that is a very weird thing to do.
-# so we are confident that this is safe.
-#  Zygote and Diffractor fall over completely without this rule.
+# This rule isn't certain to be correct.
+# This rule _is_ correct when elements of `X` pass straight through into the output without
+# modification. e.g. if the output type of `vect` is a `Vector{Any}`, then this should be
+# correct. If, however, the type of some elements of `X` change in this operation in such a
+# way that their cotangents require projection on the reverse pass of AD, this projection
+# won't happen, possibly causing problems later on in the reverse pass.
+# Zygote and Diffractor fall over completely without this rule.
 function rrule(::typeof(Base.vect), X::Vararg{Any,N}) where {N}
     vect_pullback(ȳ) = (NoTangent(), ntuple(n -> ȳ[n], N)...)
     return Base.vect(X...), vect_pullback
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -37,13 +37,7 @@ function rrule(
     return Base.vect(X...), vect_pullback
 end
 
-# This rule isn't certain to be correct.
-# This rule _is_ correct when elements of `X` pass straight through into the output without
-# modification. e.g. if the output type of `vect` is a `Vector{Any}`, then this should be
-# correct. If, however, the type of some elements of `X` change in this operation in such a
-# way that their cotangents require projection on the reverse pass of AD, this projection
-# won't happen, possibly causing problems later on in the reverse pass.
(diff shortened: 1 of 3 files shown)


src/rulesets/Base/array.jl
```diff
@@ -37,6 +37,12 @@ function rrule(
     return Base.vect(X...), vect_pullback
 end
 
+# This rule isn't always correct, but Zygote and Diffractor fall over completely without it.
```
oxinabox, line 40: Is it correct to say

```suggestion
# This rule isn't certain to be correct, if someone has overloaded `Base.vect` for their type
# in a way that has a non-identity derivative; but that is a very weird thing to do.
# so we are confident that this is safe.
#  Zygote and Diffractor fall over completely without this rule.
```

oxinabox (approved):
Approved, subject to my comment being correct.
if it is wrong then i need more info.

src/rulesets/Base/array.jl
willtebbutt: I'm unsure about the correctness of this statement, because I don't fully understand the internals of `vect`. In particular, I don't have a good handle on when the objects being put inside the vector are modified and when they "pass straight through". Do you understand how this works in general?

src/rulesets/Base/array.jl
```diff
@@ -37,6 +37,12 @@ function rrule(
...
+function rrule(::typeof(Base.vect), X...)
+    vect_pullback(ȳ) = (NoTangent(), ȳ...)
```
mcabbott, line 42: I think this will make `@code_warntype` happier:
```suggestion
function rrule(::typeof(Base.vect), X::Vararg{Any,N}) where {N}
    vect_pullback(ȳ) = (NoTangent(), ntuple(n -> ȳ[n], N)...)
```

src/rulesets/Base/array.jl
mcabbott: I thought it converted to the type from `mapreduce(typeof, Base.promote_type, ([1 2], [3,4]'))`, but not 100%.

I also think the comment should just say something like "Can't currently apply ProjectTo to arbitrary types, so leave them alone" as I think that's the extent of the non-correctness -- something might be promoted `vect` in a way that the rule won't un-do, but this is a pretty mild form of incorrectness. Unless you think it can go more seriously wrong?

src/rulesets/Base/array.jl
oxinabox: We should think about making some helpers around this

src/rulesets/Base/array.jl
oxinabox: ok cool.
Yeah, i am fine with missing some `ProjectTo`s, especially in "nonmathematical" operations like this.
As long as all mathematical operations do the right thing when creating the cotangents then later things just moving them around don't have to worry.

I still approve this PR, but some good comment needs to be added.
You can't just say "isn't always correct" without explaining anything about when it isn't

willtebbutt: >  You can't just say "isn't always correct" without explaining anything about when it isn't

Haha that's a fair point. I'll improve the comment.

> As long as all mathematical operations do the right thing when creating the cotangents then later things just moving them around don't have to worry.

I'm not totally convinced this is true. The case that concerns me is somehow a sparse array getting promoted to a dense array in this operation, and the rule for whatever function which produced the sparse array receiving a dense (little-t) tangent matrix on the reverse pass, and not knowing what to do with it. I'm still not 100% up to speed with the new state of the world though, so possibly I'm missing a detail 🤷

willtebbutt:
@mcabbott @oxinabox comment updated. Please let me know if it's descriptive enough.

src/rulesets/Base/array.jl
```diff
@@ -37,6 +37,18 @@ function rrule(
     return Base.vect(X...), vect_pullback
 end
 
+# This rule isn't certain to be correct.
+# This rule _is_ correct when elements of `X` pass straight through into the output without
+# modification. e.g. if the output type of `vect` is a `Vector{Any}`, then this should be
+# correct. If, however, the type of some elements of `X` change in this operation in such a
+# way that their cotangents require projection on the reverse pass of AD, this projection
+# won't happen, possibly causing problems later on in the reverse pass.
+# Zygote and Diffractor fall over completely without this rule.
```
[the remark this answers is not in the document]
willtebbutt:
```suggestion
# This should be fine for everything in the stdlibs though.
# Zygote and Diffractor fall over completely without this rule.
```
How about this?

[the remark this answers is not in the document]
oxinabox: sure. I would even go further and say, that it is fine everythign that doesn't overload `vect` in a weird way.

I would also at this stage be willing to just remove the comment entirely, as I am convienced it will never cause a problem.
`vect` is not something anything should overload?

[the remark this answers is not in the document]
willtebbutt: Okay. I'll just remove the entire comment. If it causes problems, we can revisit.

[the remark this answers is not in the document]
willtebbutt: I've changed the comment. Please resolve if you're happy, and I'll go ahead and get this merged and tagged.

PR #497, opened 2021-08-11
Add rule for `sincospi`

RainerHeintzmann:
Hello,
as discussed on [discourse](https://discourse.julialang.org/t/automatic-differentiation-of-cispi/66075), differentiation of `cispi` fails, which is due to `copysign` being non differentiable. This problem is avoided by defining a rule for `sincospi` in `fastmath_able.jl`. Note the the sign was moved from `\pi` to the sine term (`sinpix`), to warrant that `\pi` does not enforce Float64 as result type, which `-\pi` would do.
Hope that the edit is acceptable, but happy to adapt, if needed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.5.1"
+version = "1.5.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.5.2"
+version = "1.6.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


oxinabox (approved):
Can you bump the minor version number also,
so can tag a new release?

oxinabox:
Looks like it need a guard behind a `if VERSION >= v1.6`, since `sincospi` was added in 1.6.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.5.1"
+version = "1.5.2"
```
mcabbott, line 3: Not super-important, but I think this counts as a feature:
```suggestion
version = "1.6.0"
```

mcabbott:
Thanks!

Project.toml
RainerHeintzmann: :-) Thanks for merging...
At least `ChainRules.jl` is on par with Julia base now ;-)

roflmaostc:
One thing it noticed and was wondering about.

Why do we test with `Float64` and not with `Float32`? For example, this pi type change we would not discover by checking with a `Float64`. But we would do, with checking a `Float32`.

mcabbott:
I think the reason is that finite differences has larger errors there. But I agree this will miss many type promotion bugs.

The rrule from `@scalar_rule` should now have projectors, which will correct this. But still not ideal to convert twice. 

It would be possible to explicitly test I suppose. Testing could (piratically) turn such projection off, or make it an error. And it could convert all Float64 to Float32 just to check types, even if it doesn't do finite differences? Would have to dig quite a bit in ChainRulesTestUtils.

PR #499, opened 2021-08-13
shave another 7 mins off tests

simeonschaub:


oxinabox:
How?I don't understand

oxinabox (approved):

simeonschaub:
> How?I don't understand

We do define a bunch of top-level as well as anonymous functions inside the test, which is probably what adds up.

I have pushed some changes to infer even less, which saves us another 3 minutes or so.

test/runtests.jl
```diff
@@ -13,17 +24,21 @@ using StaticArrays
 using Statistics
 using Test
 
+@nospecialize
+
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@compiler_options"))
+    @eval Base.Experimental.@compiler_options compile=min optimize=0 infer=no
```
kellertuer, line 30: May I ask to this line how this works? I am trying to use this in my code and it seems to not work (compared to starting Julia with those settings as command line parameters). Is there a trick to make this work?

kellertuer: My approach is https://github.com/JuliaManifolds/Manifolds.jl/pull/507/files

PR #500, opened 2021-08-14
use JuliaInterpreter for even faster tests 

simeonschaub:
Turns out we can use JuliaInterpreter to shave another good 30% off test times. It does mean taking on another test-only dependency, but I think that's worth it.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -25,7 +25,7 @@ Random.seed!(1) # Set seed that all testsets should reset to.
 function include_test(path::String; interpret=true)
     println("Testing $path:")  # print so TravisCI doesn't timeout due to no output
     if interpret
-        @time include(path) do ex
+        @time Base.include(@__MODULE__(), path) do ex
             Meta.isexpr(ex, :macrocall) && ex.args[1] == Symbol("@testset") || return ex
             return :(@interpret (() -> $ex)()) # interpret testsets using JuliaInterpreter
         end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,15 +22,11 @@ union!(JuliaInterpreter.compiled_modules, Any[Base, Base.Broadcast, Compat, Line
 
 Random.seed!(1) # Set seed that all testsets should reset to.
 
-function include_test(path::String; interpret=true)
+function include_test(path::String)
     println("Testing $path:")  # print so TravisCI doesn't timeout due to no output
-    if interpret
-        @time Base.include(@__MODULE__(), path) do ex
-            Meta.isexpr(ex, :macrocall) && ex.args[1] == Symbol("@testset") || return ex
-            return :(@interpret (() -> $ex)()) # interpret testsets using JuliaInterpreter
-        end
-    else
-        @time include(path)  # show basic timing, (this will print a newline at end)
+    @time Base.include(@__MODULE__(), path) do ex
+        Meta.isexpr(ex, :macrocall) && ex.args[1] == Symbol("@testset") || return ex
+        return :(@interpret (() -> $ex)()) # interpret testsets using JuliaInterpreter
     end
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2,19 +2,12 @@ using Test, ChainRulesCore, ChainRulesTestUtils
 
 @nospecialize
 
-# don't interpret this, since this breaks some inference tests
-@time include("test_helpers.jl")
-println()
-
-module Tests
-
-using ..Main: Multiplier, NoRules
 using Base.Broadcast: broadcastable
 using ChainRules
 using ChainRulesCore
 using ChainRulesTestUtils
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -23,22 +16,28 @@ using Random
...
-
-if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@compiler_options"))
-    @eval Base.Experimental.@compiler_options compile=min optimize=0 infer=no
-end
+union!(JuliaInterpreter.compiled_modules, Any[Base, Base.Broadcast, Compat, LinearAlgebra, Random, StaticArrays, Statistics])
 
 Random.seed!(1) # Set seed that all testsets should reset to.
 
-function include_test(path::String)
+function include_test(path::String; interpret=true)
```
oxinabox, line 25: we seem to never set `interpret=false`.
We would just remove this option.

test/runtests.jl
```diff
@@ -23,22 +16,28 @@ using Random
...
 using Statistics
 using Test
+using JuliaInterpreter
 
-@nospecialize
-
-if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@compiler_options"))
-    @eval Base.Experimental.@compiler_options compile=min optimize=0 infer=no
-end
+union!(JuliaInterpreter.compiled_modules, Any[Base, Base.Broadcast, Compat, LinearAlgebra, Random, StaticArrays, Statistics])
```
oxinabox, line 21: comment this line?

oxinabox: What if we run FiniteDifferences.jl in compiled?

oxinabox:
Seems to be failing in 1.0, because `include(f, path)` was introduced in 1.5.
Maybe a quick PR to Compat.jl to add it?

test/runtests.jl
baggepinnen: What if JulieInterpreter will have future problems on, e.g., nightly, making testing ChainRules impossible until JI is working again?

test/runtests.jl
simeonschaub: OK, I will.

> What if we run FiniteDifferences.jl in compiled?

I should AB-test that again, but I think a lot of the improvements here are coming from not compiling FiniteDifferences.

simeonschaub:
> Seems to be failing in 1.0, because include(f, path) was introduced in 1.5.
Maybe a quick PR to Compat.jl to add it?

Sounds like a plan!

test/runtests.jl
simeonschaub: I previously had this because I observed some tests being faster this way, but those went away by letting some more modules be compiled, so I might as well delete it.

> What if JulieInterpreter will have future problems on, e.g., nightly, making testing ChainRules impossible until JI is working again?

Guess someone will have to fix JuliaInterpreter then. :) Usually we fix those pretty fast though.

oxinabox: > What if JulieInterpreter will have future problems on, e.g., nightly, making testing ChainRules impossible until JI is working again?

Editting `include_test(x)` to  say `@time include(x)` would  take for all intents and purposes as long as editting `interpret=true` to `interpret=false`.

test/runtests.jl
oxinabox: ah, but we *do* compile Base, which contains the primal functions that are actually being called by FiniteDifferences.jl
I get it.
Makes sense

simeonschaub: Just checked again to be sure and compiling FiniteDifferences is indeed slower.

oxinabox:
We should maybe just remove coveralls, since it can't see through julia interpretter?

test/runtests.jl
```diff
@@ -23,22 +16,24 @@ using Random
...
 
 Random.seed!(1) # Set seed that all testsets should reset to.
 
 function include_test(path::String)
     println("Testing $path:")  # print so TravisCI doesn't timeout due to no output
-    @time include(path)  # show basic timing, (this will print a newline at end)
+    @time Base.include(@__MODULE__(), path) do ex
+        Meta.isexpr(ex, :macrocall) && ex.args[1] == Symbol("@testset") || return ex
+        return :(@interpret (() -> $ex)()) # interpret testsets using JuliaInterpreter
```
oxinabox, line 29: BlueStyle likes 2 spaces for comments on same line
```suggestion
        return :(@interpret (() -> $ex)())  # interpret testsets using JuliaInterpreter
```

oxinabox (approved):
nice work

simeonschaub:
Are we OK with losing coverage information? I do hope that can be fixed in JuliaInterpreter at some point though?

oxinabox:
Yeah, we are ok with that. I am happy to make that trade to get 30% faster tests.
Probably comment out, or outright delete (can always get back from history) the code it the GHA that  reports coverage.
Nonsense wasting out jobs on it if it is always going to return 15%

PR #501, opened 2021-08-15
[WIP] Add rrule for Core._apply_iterate

dfdx:
This is my pretty ugly attempt to create an `rrule()` for `Core._apply_iterate()`. `_apply_iterate()` isn't really well-documented, so let me describe what it does and, consequently, what my code does. 

Apparently, Julia uses `_apply_iterate()` when you splat arguments in your function call. I encountered it while tracing [this Flux function](https://github.com/FluxML/Flux.jl/blob/master/src/layers/conv.jl#L163), here's also and MWE:

```julia
julia> f = (x, sz) -> reshape(x, sz...)
#3 (generic function with 1 method)

julia> args = (rand(3, 4), (12, ))
...

julia> @code_lowered f(args...)
CodeInfo(
1 ─ %1 = Core.tuple(x)
│   %2 = Core._apply_iterate(Base.iterate, Main.reshape, %1, sz)
└──      return %2
)
```
`_apply_iterate(iterate, f, args...)` takes arguments as tuples, unpacks them and applies `f()` to their flat version. `_apply_iterate(iterate, reshape, x, 12)` is equal to simply `reshape(x, 12)`. I couldn't find any explanation for this transformation, but perhaps it makes the compiler's job a little bit easier, so we have to deal with it. 

My `rrule()` implementation consists of 3 parts: 

* flatten arguments
* apply `rrule` to flat arguments to obtain value and function pullback
* restructure the values from the pullback to correspond to the original tuples

The implementation is pretty ugly, and I still need to test it on more examples, but suggestions of a better implementation are highly welcome!

### DIFF
--- a/src/rulesets/Core/core.jl
+++ b/src/rulesets/Core/core.jl
@@ -32,3 +32,34 @@ function rrule(::typeof(ifelse), c, a::Number, b::Number)
     ifelse_pullback(Δ) = (NoTangent(), NoTangent(), ifelse(c, Δ, zero(Δ)), ifelse(c, zero(Δ), Δ))
     return ifelse(c, a, b), ifelse_pullback
 end
+
+function rrule(
+    ::typeof(Core._apply_iterate), ::typeof(iterate), f::F, args...) where F
+    # flatten nested arguments
+    flat = []
+    for a in args
+        push!(flat, a...)
+    end
+    # apply rrule of the function on the flat arguments
+    y, pb = rrule(f, flat...)
+    sizes = map(length, args)
+    function _apply_iterate_pullback(dy)
+        flat_dargs = pb(dy)
+        df = flat_dargs[1]
+        # group derivatives to tuples of the same sizes as arguments
+        dargs = []
+        j = 2
+        for i=1:length(args)
+            darg_val = flat_dargs[j:j + sizes[i] - 1]
+            if length(darg_val) == 1 && darg_val[1] isa NoTangent
+                push!(dargs, darg_val[1])
+            else
+                darg = Tangent{typeof(darg_val)}(darg_val...)
+                push!(dargs, darg)
+            end
+            j = j + sizes[i]
+        end
+        return NoTangent(), NoTangent(), df, dargs...
+    end
+    return y, _apply_iterate_pullback
+end
--- a/test/rulesets/Core/core.jl
+++ b/test/rulesets/Core/core.jl
@@ -10,3 +10,10 @@ end
     test_rrule(ifelse, true, [1.1], [2.0]; check_inferred=false)
     test_frule(ifelse, false, [1.1], [2.0]; check_inferred=false)
 end
+
+@testset "_apply_iterate" begin
+    test_rrule(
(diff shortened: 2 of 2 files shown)


src/rulesets/Core/core.jl
```diff
@@ -32,3 +32,34 @@ function rrule(::typeof(ifelse), c, a::Number, b::Number)
...
+
+function rrule(
+    ::typeof(Core._apply_iterate), ::typeof(iterate), f::F, args...) where F
+    # flatten nested arguments
+    flat = []
+    for a in args
+        push!(flat, a...)
+    end
+    # apply rrule of the function on the flat arguments
+    y, pb = rrule(f, flat...)
```
oxinabox, line 44: This needs to use `rrule_via_ad`
Since nothing promised that an `rrule` will exist.

This mean this rule needs a config

oxinabox:
urrg, yeah, ugly, and difficult.
Ping me when you when you have the nice version that is ready for review.

src/rulesets/Core/core.jl
mcabbott: Does this have overhead?

I'd worry a little bit that providing a rule for some deep-internal function like this would insert ChainRules into lots of places where it needn't go. And I wonder whether this shouldn't be handled by the AD itself? It's about how Julia structures function calls, rather than some mathematical property.

dfdx: Yeah, `Core._apply_iterate()` may be deep-internal, but as far as I understand it arises in user-level code every time we use the splatting operator. If a user writes:

```julia
reshape(x, sz...)
```
they expect this call to be handled by AD independently of what it looks like in the lowered form. And in our case it looks looks like `Core._apply_iterate(iterate, reshape, sz)`.

I thought about replacing `Core._apply_iterate(iterate, f, args...)` directly with `f(f, a, b, c)` before running the reverse pass in the AD, but in general case it's not possible to rewrite array of tuples `args` into flat argument `a, b, c` because you may not have access to details of the tuples. So it's unclear how to represent calls to `_apply_iterate()` in a more ChainRules-friendly way. 

One way to stop this `rrule` from getting to the "wrong places" is to restrict it only to specific functions `f`, e.g.:

```julia
function rrule(
    ::typeof(Core._apply_iterate), ::typeof(iterate), f::typeof(reshape), x, sz...)
   ...
end
```
I suppose it will be safer and maybe even more efficient in some cases. On the downside, if we introduce, say, 20 such rules and in some future versions Julia changes the way splatting is handled, we will need to update all the 20 rules for that version. But perhaps we can talk about it when we have 20 such rules :smile:

dfdx: The more I think about handling it on the AD side, the more I like this idea. In particular, in Yota I have an alternative rule engine ("d-rules") that are checked before ChainRules. Since so far I'm the only person to complain about missing rrule for `_apply_iterate()`, it makes sense to use that machinery instead. If it works well, we can postpone the more generic solution until we get more use cases.

dfdx:
Moved and restricted to Yota for now.

Issue #498, opened 2021-08-12
Rules for `copysign` are missing?


RainerHeintzmann:
`copysign` is used in `ChainRules.jl` in various places, which may have to do with adherence to the IEEE float standard.
Yet, `copysign` throws a, probably intended error:
```julia
julia> Zygote._pullback(copysign, 3.33, 4.4)[2](1)
ERROR: Non-differentiable function Core.Intrinsics.copysign_float
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] macro expansion
   @ ~\.julia\packages\Zygote\TaBlo\src\compiler\interface2.jl:0 [inlined]
 [3] (::typeof(∂(λ)))(Δ::Int64)
   @ Zygote ~\.julia\packages\Zygote\TaBlo\src\compiler\interface2.jl:43
 [4] Pullback
   @ .\floatfuncs.jl:5 [inlined]
 [5] (::typeof(∂(copysign)))(Δ::Int64)
   @ Zygote ~\.julia\packages\Zygote\TaBlo\src\compiler\interface2.jl:0
 [6] top-level scope
   @ REPL[56]:1
```
, but the question is whether this is really required for the pullback rule. E.g. a function like `abs` works as expected but is also, strictly speaking, not differentiable at the origin. Maybe this could work:
```julia
@scalar_rule copysign(y, x) (sign(x)*y, 0)
```
or is there a fundamental reason that `copysign` should not have an allowed gradient, but `abs` should?

willtebbutt:
Nope, this would be a good addition. If you have the time, a PR would be very welcome.

RainerHeintzmann:
Sure, can do so. Maybe I should wait till my other pull request (`sincospi`) is merged and then add this?

willtebbutt:
Sounds good 👍

RainerHeintzmann:
Hmm, interestingly the `sign()` function yields a gradient of nothing:
```julia
julia> Zygote.gradient((x) -> sign(x), 1.2)
(nothing,)
```
should this not be `0`? Bug or feature?

willtebbutt:
I think this is probably a Zygote thing. For historical reasons, Zygote represents zero tangents by `nothing`, whereas ChainRules uses either `0` or `ZeroTangent()`. I would _hope_ that if you called `rrule` directly you would obtain a `0`.

RainerHeintzmann:
Yes, you are right:
```julia
[...]

PR #502, opened 2021-08-16
new rule: copysign

RainerHeintzmann:
As opposed to the previous [discussion](https://github.com/JuliaDiff/ChainRules.jl/issues/498) the correct rule for `copysign(y,x)` should probably be:
```julia
@scalar_rule copysign(y, x) (ifelse(signbit(x)!=signbit(y), -one(y), +one(y)), NoTangent())
```
. You can also find the arguments there why `NoTangent()` is preferred over `0`. 
The testing gets into problems, if too close to zero for x or y, due to the non-continuity of this function.


oxinabox:
thanks!

PR #505, opened 2021-08-18
reenable coverage

simeonschaub:
Might need to wait a little bit until the latest nightly builds
propagate through.


simeonschaub:
Weird, looks like codecov reports a drop in coverage while coveralls still reports 98.86%. Perhaps we should just get rid of codecov?

simeonschaub:
I have dropped the codecov job for now.

oxinabox:
> I have dropped the codecov job for now.

Good, I dislike having both, but i always forget which I like

oxinabox (approved):

PR #506, opened 2021-08-19
do not differentiate through the construction of BitArray

frankschae:
Based on @oxinabox 's reply here: https://github.com/SciML/DiffEqFlux.jl/issues/502#issuecomment-902023817 .

This solves also the issues with the callback construction in https://github.com/SciML/DiffEqFlux.jl/issues/502.

Does it need a test? Should I maybe add one from https://github.com/SciML/DiffEqFlux.jl/issues/502#issuecomment-902015247 in here: https://github.com/JuliaDiff/ChainRules.jl/blob/master/test/rulesets/Base/array.jl ?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.9.1"
+version = "1.10.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.9"
+version = "1.9.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Base/nondiff.jl
+++ b/src/rulesets/Base/nondiff.jl
@@ -84,6 +84,7 @@
 @non_differentiable Vector(::AbstractArray{Bool})
 
 @non_differentiable Array(::AbstractArray{Bool})
+@non_differentiable BitArray(::Any)
 @non_differentiable IndexStyle(::AbstractArray{Bool})
 
 #####


oxinabox:
Cool, LGTM.
We don't generally test `@non_differentiable` rules,
because any bugs in them are not normally in the rule, but in the definition of the `@non_differentiable` macro.
And that has it's own tests.

Sometimes we test them, if we have had them defined wrong in the past.
But that is not the case here

oxinabox:
Will merge once CI passes.
(and I will bump the version and tag a release).
Or you can bump the version and make it easier for me to tag a release.
This is a new feature so it is a minor bump.

frankschae:
cool, Thanks a lot!

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.9"
+version = "1.9.1"
```
oxinabox, line 3: Minor change.
Those is a feature not a bug fix


```suggestion
version = "1.10.0"
```

src/rulesets/Base/nondiff.jl
```diff
@@ -84,6 +84,7 @@
 @non_differentiable Vector(::AbstractArray{Bool})
 
 @non_differentiable Array(::AbstractArray{Bool})
+@non_differentiable BitArray(::Any)
```
mzgubic, line 87: Is there a reason we didn't splat this?
```julia
@non_differentiable BitArray(::Any...)
```

oxinabox: Doesn't it only have 1 constructor?
I didn't actually check

mzgubic:
```julia
julia> methods(BitArray)
# 6 methods for type constructor:
[1] BitArray(::UndefInitializer, dims::Integer...) in Base at bitarray.jl:69
[2] BitArray(::UndefInitializer, dims::Tuple{Vararg{Integer, N}}) where N in Base at bitarray.jl:71
[3] BitArray(x::BitArray) in Base at bitarray.jl:543
[4] BitArray(A::AbstractArray{var"#s79", N} where var"#s79") where N in Base at bitarray.jl:503
[5] (::Type{T})(x::T) where T<:BitArray in Base at bitarray.jl:542
[6] BitArray(itr) in Base at bitarray.jl:574
```
there seems to be more

frankschae: Should I make another PR?

mzgubic: Please do, that would be great, thanks!

PR #520, opened 2021-08-31
Don't ProjectTo in reverse or circshift

### LINKED ISSUE #518: regression: "no method matching ChainRulesCore.ProjectTo" when calling reverse on a vector of functors
Hi,
I'm experiencing a method error I cannot explain.
When I run the following code with [d360d2e6] ChainRulesCore v1.3.0 I get an error:

```Julia
using Flux

m = [
    Dense(2, 2),
    Dense(2, 2),
]

function f(x)
    sum(
        reverse(m)[1](x)
    )
end

gradient(f, [4., 4.])
```

```
ERROR: LoadError: MethodError: no method matching ChainRulesCore.ProjectTo(::Dense{typeof(identity), Matrix{Float32}, Vector{Float32}})
Closest candidates are:
  ChainRulesCore.ProjectTo(::Tridiagonal{T, V} where V<:AbstractVector{T}) where T<:Number at ~/.julia/packages/ChainRulesCore/Voykb/src/projection.jl:382
  ChainRulesCore.ProjectTo(::Bidiagonal{T, V} where V<:AbstractVector{T}) where T<:Number at ~/.julia/packages/ChainRulesCore/Voykb/src/projection.jl:355
  ChainRulesCore.ProjectTo(::UnitLowerTriangular) at ~/.julia/packages/ChainRulesCore/Voykb/src/projection.jl:336
  ...
```
I do not get this error when I use
[d360d2e6] ChainRulesCore v0.10.13

Could this be related to this?
https://github.com/FluxML/Zygote.jl/pull/1051

Edit:
I tried adding the relevant @adjoint that was removed in the PR linked above:

```Julia
Flux.@adjoint function reverse(x::AbstractArray, args...; kwargs...)
    _reverse(t) = reverse(t, args...; kwargs...)
    _reverse(x), Δ->(_reverse(Δ), map(_->nothing, args)...)
end
```
And it works.

versions:
Julia v1.6.2
[587475ba] Flux v0.12.6

oxinabox:
Closes #518 

ProjectTo in these rearranging linear operators seems not worth it.
It causes problems since they are often used on types that are on representing vector spaces -- and ProjectTo is not fully defined on those.
Which then error.
See #518 for more details.

These are fine if we were given good pullback inputs (i.e. if the functions called after did ensure things were good), we preserve those.

(circshift doesn't work on (any?) structurally spaces matrixes anyway)

### DIFF
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -171,7 +171,9 @@ end
 
         # Structured
         y, pb = rrule(reverse, Diagonal([1,2,3]))
-        @test unthunk(pb(rand(3,3))[2]) isa Diagonal
+        # We only preserve structure in this case if given structured tangent (no ProjectTo)
+        @test unthunk(pb(Diagonal([1.1, 2.1, 3.1]))[2]) isa Diagonal
+        @test unthunk(pb(rand(3, 3))[2]) isa AbstractArray
     end
 end
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.11.1"
+version = "1.11.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -303,10 +303,9 @@ function frule((_, xdot), ::typeof(reverse), x::AbstractArray, args...; kw...)
 end
 
 function rrule(::typeof(reverse), x::AbstractArray, args...; kw...)
-    project = ProjectTo(x)
     nots = map(_ -> NoTangent(), args)
     function reverse_pullback(dy)
-        dx = @thunk project(reverse(unthunk(dy), args...; kw...))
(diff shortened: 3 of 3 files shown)


mzgubic (approved):

test/rulesets/Base/array.jl
```diff
@@ -171,7 +171,9 @@ end
 
         # Structured
         y, pb = rrule(reverse, Diagonal([1,2,3]))
-        @test unthunk(pb(rand(3,3))[2]) isa Diagonal
+        # We only preserve structure in this case if given structured tangent (no ProjectTo)
+        @test unthunk(pb(Diagonal([1.1, 2.1, 3.1]))[2]) isa Diagonal
+        @test unthunk(pb(rand(3, 3))[2]) isa AbstractArray
```
mcabbott, line 176: Instead of just testing the absence of projection, how hard would it be to test something that doesn't want it?  Like `rrule(reverse, [(; x=[1,2]), (; x=[3,4])])`.

The structured arrays that do play well with reverse, and might reasonably show up in the wild, are row vectors:
```
julia> reverse([1,2,3]')
1×3 adjoint(::Vector{Int64}) with eltype Int64:
 3  2  1

julia> reverse([1,2,3]', dims=2)
1×3 adjoint(::Vector{Int64}) with eltype Int64:
 3  2  1
```
Not sure they need testing though, they will just work.

mcabbott: Oh, too slow! Never mind.

PR #524, opened 2021-09-03
`@opt_out of sum(array, array)`

### LINKED ISSUE #522: `sum(f, xs)` dispatches on non functions
The rule in https://github.com/JuliaDiff/ChainRules.jl/blob/a130b8f62e21f1dc4feceb4c8d5571fc141a8187/src/rulesets/Base/mapreduce.jl#L66-L89
which is meant to be used where `f` is a function/functor also dispatches on

```julia
julia> using Zygote

julia> using ChainRulesCore

julia> using StatsBase

julia> rrule(Zygote.ZygoteRuleConfig(), sum, rand(3), AnalyticWeights([1.0, 2.0, 3.0]))
ERROR: MethodError: objects of type Vector{Float64} are not callable
Use square brackets [] for indexing an Array.
Stacktrace:
  [1] macro expansion
    @ ~/JuliaEnvs/PortfolioNets.jl/dev/Zygote/src/compiler/interface2.jl:0 [inlined]
  [2] _pullback(ctx::Zygote.Context, f::Vector{Float64}, args::Float64)
    @ Zygote ~/JuliaEnvs/PortfolioNets.jl/dev/Zygote/src/compiler/interface2.jl:9
  [3] rrule_via_ad(::Zygote.ZygoteRuleConfig{Zygote.Context}, ::Vector{Float64}, ::Vararg{Any, N} where N; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Zygote ~/JuliaEnvs/PortfolioNets.jl/dev/Zygote/src/compiler/chainrules.jl:185
  [4] rrule_via_ad(::Zygote.ZygoteRuleConfig{Zygote.Context}, ::Vector{Float64}, ::Vararg{Any, N} where N)
    @ Zygote ~/JuliaEnvs/PortfolioNets.jl/dev/Zygote/src/compiler/chainrules.jl:180
  [5] (::ChainRules.var"#1388#1389"{Zygote.ZygoteRuleConfig{Zygote.Context}, Vector{Float64}})(x::Float64)
    @ ChainRules ~/JuliaEnvs/PortfolioNets.jl/dev/ChainRules/src/rulesets/Base/mapreduce.jl:71
  [6] iterate
    @ ./generator.jl:47 [inlined]
  [7] _collect
    @ ./array.jl:691 [inlined]
  [8] collect_similar
    @ ./array.jl:606 [inlined]
  [9] map
    @ ./abstractarray.jl:2294 [inlined]
 [10] rrule(config::Zygote.ZygoteRuleConfig{Zygote.Context}, ::typeof(sum), f::Vector{Float64}, xs::AnalyticWeights{Float64, Float64, Vector{Float64}}; dims::Function)
    @ ChainRules ~/JuliaEnvs/PortfolioNets.jl/dev/ChainRules/src/rulesets/Base/mapreduce.jl:71
 [11] rrule(config::Zygote.ZygoteRuleConfig{Zygote.Context}, ::typeof(sum), f::Vector{Float64}, xs::AnalyticWeights{Float64, Float64, Vector{Float64}})
    @ ChainRules ~/JuliaEnvs/PortfolioNets.jl/dev/ChainRules/src/rulesets/Base/mapreduce.jl:70
 [12] top-level scope
    @ REPL[40]:1
```

Is there anything we can do about this? There are several things that come to mind but none are amazing:
- define a rule locally (i've done that, but other might run into the exact same problem)
- add a more informative error message checking if `f` is callable (so that users know they need to add a rule)
- add a rule to StatsBase, which does not depend on ChainRulesCore (they probably won't accept the PR?)
- tighten `f` to `f::Function`, but that means the rule does dispatch on functors

I am leaning towards just throwing a better error. Maybe also open in issue in StatsBase?

mzgubic:
closes #522 

This will still error if `f` in `sum(f, xs)` is a non-callable struct, but we can cross that bridge once (if ever) we actually get to it.

### DIFF
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -89,7 +89,15 @@ function rrule(
 end
 
 # https://github.com/JuliaDiff/ChainRules.jl/issues/522
-@opt_out ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(sum), x::AbstractArray, y::AbstractArray; dims=:)
+# The rule above assumes `f` is callable. Arrays are not, this came up when summing
+# arrays with weights in StatsBase
+@opt_out ChainRulesCore.rrule(
+    config::RuleConfig{>:HasReverseMode},
+    ::typeof(sum),
+    x::AbstractArray,
+    y::AbstractArray;
+    dims=:
+)
 
 function frule(
     (_, _, Δx),
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -88,6 +88,9 @@ function rrule(
     return y, sum_pullback
 end
 
+# https://github.com/JuliaDiff/ChainRules.jl/issues/522
+@opt_out ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(sum), x::AbstractArray, y::AbstractArray; dims=:)
+
 function frule(
     (_, _, Δx),
     ::typeof(sum),
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.11.4"
+version = "1.11.5"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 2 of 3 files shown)


src/rulesets/Base/mapreduce.jl
```diff
@@ -88,6 +88,9 @@ function rrule(
     return y, sum_pullback
 end
 
+# https://github.com/JuliaDiff/ChainRules.jl/issues/522
```
oxinabox, line 91: Maybe a one line summary as well?

oxinabox (approved):

PR #534, opened 2021-09-24
Handle sum(f,xs) where f returns a Vector

oxinabox:
Closes https://github.com/FluxML/Zygote.jl/issues/1074

Basically when there is not a `dims` argument given, we need to do a "left-hand" broadcast.
where `call.(pullback, ȳ)` broadcasts `ȳ` up to the shape of `pullback`
but when `dims` is given then we need to do a "outer" broadcast that broadcasts using shape info from both `pullback` and `ȳ`

### DIFF
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -74,8 +74,10 @@ function rrule(
     project = ProjectTo(xs)
 
     function sum_pullback(ȳ)
-        call(f, x) = f(x)  # we need to broadcast this to handle dims kwarg
-        f̄_and_x̄s = call.(pullbacks, ȳ)
+        call(f, x) = f(x)
+        # if dims is :, then need only left-handed only broadcast
+        broadcast_ȳ = dims isa Colon  ? (ȳ,) : ȳ
+        f̄_and_x̄s = call.(pullbacks, broadcast_ȳ)
         # no point thunking as most of work is in f̄_and_x̄s which we need to compute for both
         f̄ = if fieldcount(typeof(f)) === 0 # Then don't need to worry about derivative wrt f
             NoTangent()
--- a/test/rulesets/Base/mapreduce.jl
+++ b/test/rulesets/Base/mapreduce.jl
@@ -92,6 +92,15 @@ struct SumRuleConfig <: RuleConfig{Union{HasReverseMode}} end
         test_rrule(sum, sqrt, randn(5,5) .> 0; fkwargs=(;dims=1))
         # ... and Bool produced by function
         @test_skip test_rrule(sum, iszero, randn(5))  # DimensionMismatch("second dimension of A, 1, does not match length of x, 0")
+
+
+        # Functions that return a Vector
+        # see https://github.com/FluxML/Zygote.jl/issues/1074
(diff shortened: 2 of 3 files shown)


src/rulesets/Base/mapreduce.jl
```diff
@@ -75,7 +75,8 @@ function rrule(
 
     function sum_pullback(ȳ)
         call(f, x) = f(x)  # we need to broadcast this to handle dims kwarg
-        f̄_and_x̄s = call.(pullbacks, ȳ)
+        broadcast_ȳ = dims isa Colon  ? Ref(ȳ) : ȳ
+        f̄_and_x̄s = call.(pullbacks, broadcast_ȳ)
```
oxinabox, line 79: Alternatively, we can do:
```
        f̄_and_x̄s = if dims isa Colon
            map(pb -> pb(ȳ), pullbacks)
        else
            broadcast((pb, ȳi) -> pb(ȳi), pullbacks, ȳ)
        end
```
What think?

mcabbott: I'd vote for always using broadcast. And in fact for making it a one-liner, but tastes vary, not important:
```
broadcast((f,z) -> f(z), pullbacks, dims isa Colon ? Ref(dy) : dy)
```

test/rulesets/Base/mapreduce.jl
```diff
@@ -88,6 +88,14 @@
...
         # ... and Bool produced by function
         @test_skip test_rrule(sum, iszero, randn(5))  # DimensionMismatch("second dimension of A, 1, does not match length of x, 0")
+
+
+        # Functions that return a Vector
+        # see https://github.com/FluxML/Zygote.jl/issues/1074
+        test_rrule(sum, make_two_vec, [1.0, 3.0, 5.0, 7.0])
+        test_rrule(sum, make_two_vec, [1.0 2.0; 3.0 4.0])
+        test_rrule(sum, make_two_vec, [1.0 2.0; 3.0 4.0]; fkwargs=(;dims=2))
+        test_rrule(sum, make_two_vec, [1.0 2.0; 3.0 4.0]; fkwargs=(;dims=1))
```
mcabbott, line 98: Maybe a weirder test:
```suggestion
        test_rrule(sum, make_two_vec, [1.0 2.0; 3.0 4.0]; fkwargs=(;dims=(3,4)))
```
Maybe also worth testing something in which the gradient is in the function not the array, like `gradient(x -> sum(sum(Returns([x,1]), 1:3)), 4.0)`. Maybe it would handle this?
```
gradient(x -> sum(sum(Multiplier([x 0; 0 1]), [[1 2; 3 4], [5 6; 7 8]])), 4.0)
```
I tried `gradient(x -> sum(sum(Multiplier([x 0; 0 1]), [[1,2], [3,4]])), 4.0)` and got an error, not sure if that ought to work.

mcabbott (approved):

test/rulesets/Base/mapreduce.jl
oxinabox: Not sure about the ones where the gradient is in the function.
In particular not sure how to add them without needing an AD system.
(Maybe we should write a RuleConfig for FiniteDifferences.jl and use that in CRTU...)

PR #550, opened 2021-12-03
fix for julia 1.7

### LINKED ISSUE #549: CI is failing for julia 1
Stating the obvious, but we had a nightly failure https://github.com/JuliaDiff/ChainRules.jl/runs/4390009367?check_suite_focus=true
which is probably related to changes in LinearAlgebra coming with julia 1.7

I can reproduce the error locally when running the tests, but not when copy-pasting the code necessary to run the tests into the REPL for some reason. I.e. the following should error:

```julia
using ChainRules
using ChainRulesCore
using Test
using LinearAlgebra
using FiniteDifferences
using ChainRulesTestUtils
using ChainRulesTestUtils: rand_tangent, _fdm

        # generate random matrices of type TA in the domain of f
        function rand_matfun_input(f, TA, T, uplo, n, hermout)
            U = Matrix(qr(randn(T, n, n)).Q)
            if hermout # f(A) will also be a TA
                λ = if f in (acos, asin, atanh)
                    2 .* rand(real(T), n) .- 1
                elseif f in (log, sqrt)
                    abs.(randn(real(T), n))
                elseif f === acosh
                    1 .+ abs.(randn(real(T), n))
                else
                    randn(real(T), n)
                end
            else
                λ = randn(real(T), n)
                λ = if f === atanh
                    2 .* rand(real(T), n) .- 1
                else
                    randn(real(T), n)
                end
            end
            return TA(U * Diagonal(λ) * U', uplo)
        end

f = atanh
TA = Symmetric
T = Float64
TC = Complex{real(T)}
n = 10

                    λ, U = eigen(rand_matfun_input(f, TA, T, :U, n, true))
                    m = div(n, 2)
                    λ[1:m] .= λ[m+1:2m] .+ cbrt(eps(eltype(λ))) / 100
                    A = TA(U * Diagonal(λ) * U')
                    ΔA = TA(randn(T, n, n))
                    _, ∂Y = frule((ZeroTangent(), ΔA), f, A)
                    @test parent(∂Y) ≈ jvp(_fdm, x -> Matrix{TC}(parent(f(TA(x)))), (A.data, ΔA.data))

                    λ[1:m] .= λ[m+1:2m]
                    A2 = TA(U * Diagonal(λ) * U')
                    ΔA2 = TA(randn(T, n, n))
                    _, ∂Y2 = frule((ZeroTangent(), ΔA2), f, A2)
                    @test parent(∂Y2) ≈ jvp(_fdm, x -> Matrix{TC}(parent(f(TA(x)))), (A2.data, ΔA2.data))
```

oxinabox:
Closes #549 

- Change in structured handles factorization becoming less accurate and thus finite differencing beween less accurate
-  Just ups out tolerance of error
- https://github.com/JuliaDiff/ChainRules.jl/runs/4390009367?check_suite_focus=true#step:6:134

And

 - Change in symmetric handles factorization becoming less accurate and thus the finite differencing escaping the domain (we possibly should fix the `_fdm` to point the right way though?)
 - This change just makes sure our initial point is less close to boundry

### DIFF
--- a/test/rulesets/LinearAlgebra/symmetric.jl
+++ b/test/rulesets/LinearAlgebra/symmetric.jl
@@ -307,7 +307,7 @@
                 λ = if f in (acos, asin, atanh)
                     # generate random number between -1 and 0.9. This would be between
                     # -1 and 1 but we want to avoid domain errors from numerical error
-                    1.9 .* rand(real(T), n) .- 1
+                    T(1.9) .* rand(real(T), n) .- 1
                 elseif f in (log, sqrt)
                     abs.(randn(real(T), n))
                 elseif f === acosh
--- a/test/rulesets/LinearAlgebra/symmetric.jl
+++ b/test/rulesets/LinearAlgebra/symmetric.jl
@@ -305,7 +305,9 @@
             U = Matrix(qr(randn(T, n, n)).Q)
             if hermout # f(A) will also be a TA
                 λ = if f in (acos, asin, atanh)
-                    2 .* rand(real(T), n) .- 1
+                    # generate random number between -1 and 0.9. This would be between
+                    # -1 and 1 but we want to avoid domain errors from numerical error
+                    1.9 .* rand(real(T), n) .- 1
                 elseif f in (log, sqrt)
                     abs.(randn(real(T), n))
                 elseif f === acosh
--- a/test/rulesets/LinearAlgebra/structured.jl
+++ b/test/rulesets/LinearAlgebra/structured.jl
@@ -7,8 +7,8 @@
 
         @testset "\\ $T on LHS" for T in (Diagonal, UpperTriangular, LowerTriangular)
             LHS = T(randn(T == Diagonal ? 10 : (10, 10)))
-            test_rrule(\, LHS, randn(Float32, 10); rtol = 1.0e-4, atol = 1.0e-4)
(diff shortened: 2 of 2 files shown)


test/rulesets/LinearAlgebra/symmetric.jl
```diff
@@ -305,7 +305,9 @@
             U = Matrix(qr(randn(T, n, n)).Q)
             if hermout # f(A) will also be a TA
                 λ = if f in (acos, asin, atanh)
-                    2 .* rand(real(T), n) .- 1
+                    # generate random number between -1 and 0.9. This would be between
+                    # -1 and 1 but we want to avoid domain errors from numerical error
+                    1.9 .* rand(real(T), n) .- 1
```
mcabbott, line 310: Needs a T:
```
julia> 1.9 .* rand(real(Float32), 3) .- 1
3-element Vector{Float64}:
 -0.5762628734111785
 -0.7353183567523957
  0.8970144212245941
```

Also, unrelated to this, a few lines down defines lambda twice:
```

                λ = randn(real(T), n)
                λ = if f === atanh
                    2 .* rand(real(T), n) .- 1
                else
                    randn(real(T), n)
                end
                ```

Comment (oxinabox) [reply test/rulesets/LinearAlgebra/symmetric.jl:310]:
```suggestion
                    T(1.9) .* rand(real(T), n) .- 1
```

oxinabox:
I am merging this because merging without proper review seems less bad than leaving CI broken

Issue #479, opened 2021-07-27
Use oneunit where it makes sense


sethaxen:
I suspect some of the scalar rules should be using `oneunit` instead of `one`. For example, `sign`.

sethaxen:
I just had a look at this and got really confused. So I think this needs someone who's thought more about derivatives of dimensional quantities to take care of this and is not after all a good first issue.

e.g. what should the units of (co)tangents be? A tangent of `x` is `dx/dt` for some upstream real `x`, so it should have units of `x/t`, but a cotangent of `x` is `ds/dx` for some downstream real `s`, so it should have units of `s/x`. A few checks.
The Jacobian `J` of `f: x ↦ y` has units of `y/x`. So `ẏ = J * ẋ` has units of `y/x * x/t = y/t`. We also have `x̄ = J' * ȳ` with units of `s/y * y/x = s/x`. All good.
Then we have `<ẋ, x̄>` with the desired units `<dx/dt, ds/dx>` = `ds/dt`.

This would imply that the units of (co)tangents are dependent on more than just the units of the primal, unless we explicitly define (co)tangents as being wrt dimensionless reals `s` and `t`. Then the tangent has the same units as the primal, while the cotangent's units are the inverse of the primal's.

Am I missing something here?

oxinabox:
c.f. https://github.com/PainterQubits/Unitful.jl/pull/504

PR #554, opened 2021-12-13
Use `deg2rad` and `rad2deg` instead of handcoding conversions

devmotion:
This PR replaces handcoded conversions from degrees to radians and radians to degress with `deg2rad` and `rad2deg`.

Arguably this simplifies the code and allows to use specializations in e.g. Base (https://github.com/JuliaLang/julia/blob/cbc2ce819662648da90366334e5c6bf1da1c4717/base/math.jl#L306-L336).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.16.0"
+version = "1.17.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -99,8 +99,8 @@ end
     (ifelse(isint, nan, one(u)), ifelse(isint, nan, -floor(u))),
 )
 
-@scalar_rule deg2rad(x) π / oftype(x, 180)
-@scalar_rule rad2deg(x) oftype(x, 180) / π
+@scalar_rule deg2rad(x) deg2rad(one(x))
+@scalar_rule rad2deg(x) rad2deg(one(x))
 
 @scalar_rule(ldexp(x, y), (2^y, NoTangent()))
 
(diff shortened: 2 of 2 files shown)


mzgubic (approved):
LGTM, thanks! As far as I can tell this has no functional change. I assume the CI failures are unrelated?

devmotion:
Yes, the downstream test failures seem unrelated (I ran into the Zygote error in a different repo as well).

sethaxen:
I know this has already been merged, but for e.g. unitful numbers, the change in this PR is strictly speaking less correct than the original. e.g. for `deg2rad`, if the input `x` has units of `K`, then the Jacobian should have units of `K⁻¹`. The old Jacobian of `π / oftype(x, 180)` ensures this. The new Jacobian of `deg2rad(K)` gives a unitless Jacobian so that the produced tangent has units of `K` instead of being unitless. It's not the only place we have this mistake (see #479), but it's not ideal to introduce it here.

I agree it's better to use `deg2rad` and `rad2deg` here, but I think the right way is to explicitly write the `frule` and `rrule`s:
```julia
frule((_, Δx), ::typeof(deg2rad), x::Number) = (deg2rad(x), deg2rad(Δx))
frule((_, Δx), ::typeof(rad2deg), x::Number) = (rad2deg(x), rad2deg(Δx))
function rrule(::typeof(deg2rad), x::Number)
    project_x = ProjectTo(x)
    deg2rad_pullback(Δy) = (NoTangent(), project_x(rad2deg(Δy)))
    return deg2rad(x), deg2rad_pullback
end
function rrule(::typeof(rad2deg), x::Number)
    project_x = ProjectTo(x)
    rad2deg_pullback(Δy) = (NoTangent(), project_x(deg2rad(Δy)))
    return rad2deg(x), rad2deg_pullback
end
```

devmotion:
> e.g. for deg2rad, if the input x has units of K, then the Jacobian should have units of K⁻¹.

Are you sure? Shouldn't it be `units(deg2rad(x)) / K`? Eg with the Unitful definitions in https://github.com/PainterQubits/Unitful.jl/blob/2a3308eabd23bd5d936d284576991b163ee9fbdd/src/pkgdefaults.jl#L120 I would expect that the units of the Jacobian of `deg2rad` for an input with units `°` are `rad/°`. This would mean one could use something like
```julia
@scalar_rule deg2rad(x) deg2rad(oneunit(x)) / oneunit(x)
```
?

devmotion:
(If my reasoning is correct, this eouöd imply that the previous implementation was not correct unitwise either.)

sethaxen:
> Are you sure? Shouldn't it be `units(deg2rad(x)) / K`?

Sure, I was assuming radians were treated as unitless, which they are, but if they're unitful, then what you suggest is correct.

> This would mean one could use something like
> 
> ```julia
> @scalar_rule deg2rad(x) deg2rad(oneunit(x)) / oneunit(x)
> ```
> 
> ?

Certainly that would be more correct.

But actually, on closer inspection, I take back everything I said above. First, it seems that the base implementations would not behave the "right" way for dimensional numbers anyways: https://github.com/JuliaLang/julia/blob/v1.7.0/base/math.jl#L312-L313.

Second, the reasoning that the Jacobian should have the units `K⁻¹` so that the pushforward ends up dimensionless is flawed, because then the pullback would produce a cotangent with units `K⁻¹` instead of `K`. It seems this is a case where dimensional constraints should be handled by projections. In that case, I don't think this PR is in fact any less correct than the previous implementation.

PR #555, opened 2021-12-13
Use `logten` and `logtwo` from IrrationalConstants

devmotion:
The rules for `exp2`, `exp10`, `log2`, and `log10` can benefit from the irrational constants `logten` and `logtwo` defined in the lightweight IrrationalConstants package (also a dependency of e.g. SpecialFunctions, StatsFuns, and LogExpFunctions). The package has [no dependencies](https://github.com/JuliaMath/IrrationalConstants.jl/blob/main/Project.toml).

The irrational constants take care of type promotion and eg. for `Float32` and `Float64` values of `log(10)` and `log(2)` are precomputed.


mzgubic (approved):
Thanks, also LGTM - just be careful with the versions when tagging both PRs

mcabbott:
What advantages does this have? The existing code ought to pre-compute the logs of the correct types anyway:
```julia
julia> f1(x) = x + log(oftype(x, 10))
f1 (generic function with 1 method)

julia> @code_llvm f1(1.0)
;  @ REPL[31]:1 within `f1`
define double @julia_f1_1874(double %0) #0 {
top:
; ┌ @ float.jl:399 within `+`
   %1 = fadd double %0, 0x40026BB1BBB55516
; └
  ret double %1
}

julia> @code_llvm f1(1.0f0)
;  @ REPL[31]:1 within `f1`
define float @julia_f1_1876(float %0) #0 {
top:
; ┌ @ float.jl:398 within `+`
   %1 = fadd float %0, 0x40026BB1C0000000
; └
  ret float %1
}
```

mzgubic:
It's more readable IMO (and also faster, thought not sure how much it matters since the numbers are tiny)
```julia
julia> f1(x) = x + logten
f1 (generic function with 1 method)

julia> @btime f1(1.0)
  0.040 ns (0 allocations: 0 bytes)
3.302585092994046

julia> f2(x) = x + log(oftype(x, 10))
f2 (generic function with 1 method)

julia> @btime f2(1.0)
  1.716 ns (0 allocations: 0 bytes)
3.302585092994046
```

mzgubic:
Hmm, interesting, the numbers are from julia-1.6, 1.7 gives the same result for both. I could not get rid of the difference in 1.6 by $1.0 though.

devmotion:
IrrationalConstants does not pull in any additional dependencies. To make the discussion a bit more concrete I also timed `using ChainRules` on my computer (Julia 1.7):
```julia
julia> @time using ChainRules # current release
  0.108510 seconds (175.57 k allocations: 12.275 MiB, 2.92% compilation time)

julia> @time using ChainRules # this PR
  0.113230 seconds (184.91 k allocations: 12.979 MiB, 6.09% compilation time)
```

oxinabox:
The bar for adding a dependency to ChainRules.jl is much lower than the bar for ChainRulesCore.
Since only AD packages depend on ChainRules.jl.
And they tend to already have non-trival load times.

I think using this makes us a good citizen and we will likely find other places to use it also.
We can't well say that CRC.jl is super light weight and you shouldn't hesitate to use it if we are not willing to take on similarly light things into CR.jl

devmotion:
Does this mean I should merge this PR (given that tests pass)?

oxinabox:
go a head

PR #557, opened 2021-12-26
Add rrule for parent

yuehhua:
fix

### DIFF
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -504,3 +504,12 @@ function _extrema_dims(x, dims)
     end
     return y, extrema_pullback_dims
 end
+
+#####
+##### `parent`
+#####
+
+function rrule(::typeof(parent), A::Base.SubArray)
+    parent_pullback(Δ) = (NoTangent(), view(Δ, A.indices...))
+    parent(A), parent_pullback
+end
--- a/test/rulesets/Base/array.jl
+++ b/test/rulesets/Base/array.jl
@@ -274,3 +274,12 @@ end
     B = hcat(A[:,:,1], A[:,:,1])
     @test extrema(B, dims=2) == rrule(extrema, B, dims=2)[1]
 end
+
+@testset "parent" begin
+    # Reverse
+    A = rand(3, 4)
+    B = view(A, 2:3, 2)
+    a, b = rrule(parent, B)
+    @test a === A
+    @test b(A)[2] == B
+end


mcabbott:
This doesn't look wrong, but what's the context? It seems unsafe to call `parent` in general; if some function which knows what it's doing calls it, then perhaps it would be better for that function to have a gradient?

Such a function could equally well just access fields, since it must know what type it is handling. In which case you will get a Tangent which this rule won't catch. A rule for that might be a better level to implement this, if needed.

willtebbutt:
I would also be interested to know the context here. `parent` seems totally fine in `Zygote`:
```julia
julia> out, pb = Zygote.pullback(parent, view(ones(6), 2:5))
([1.0, 1.0, 1.0, 1.0, 1.0, 1.0], Zygote.var"#52#53"{typeof(∂(parent))}(∂(parent)))

julia> pb(ones(6))
((parent = [1.0, 1.0, 1.0, 1.0, 1.0, 1.0], indices = nothing, offset1 = nothing, stride1 = nothing),)
```
but when you pre-compose it with `view` something appears to break:
```julia
julia> out, pb = Zygote.pullback(x -> parent(view(x, 2:5)), ones(6))
([1.0, 1.0, 1.0, 1.0, 1.0, 1.0], Zygote.var"#52#53"{typeof(∂(#107))}(∂(#107)))

julia> pb(ones(6))
ERROR: MethodError: Cannot `convert` an object of type Bool to an object of type Vector{Float64}
Closest candidates are:
  convert(::Type{Array{T, N}}, ::StaticArrays.SizedArray{S, T, N, N, Array{T, N}}) where {S, T, N} at ~/.julia/packages/StaticArrays/OWJK7/src/SizedArray.jl:120
  convert(::Type{Array{T, N}}, ::StaticArrays.SizedArray{S, T, N, M, TData} where {M, TData<:AbstractArray{T, M}}) where {T, S, N} at ~/.julia/packages/StaticArrays/OWJK7/src/SizedArray.jl:114
  convert(::Type{Array{T, N}}, ::FillArrays.Zeros{V, N}) where {T, V, N} at ~/.julia/packages/FillArrays/Vzxer/src/FillArrays.jl:431
  ...
Stacktrace:
 [1] convert(#unused#::Type{Union{Nothing, Vector{Float64}}}, x::Bool)
   @ Base ./some.jl:36
 [2] fill!(dest::Vector{Union{Nothing, Vector{Float64}}}, x::Bool)
   @ Base ./array.jl:351
 [3] _zero(xs::Vector{Float64}, T::Type)
   @ Zygote ~/.julia/packages/Zygote/rv6db/src/lib/array.jl:61
 [4] (::Zygote.var"#430#432"{1, Float64, Vector{Float64}, Tuple{UnitRange{Int64}}})(dy::NamedTuple{(:parent, :indices, :offset1, :stride1), Tuple{Vector{Float64}, Nothing, Nothing, Nothing}})
   @ Zygote ~/.julia/packages/Zygote/rv6db/src/lib/array.jl:37
 [5] (::Zygote.var"#2301#back#428"{Zygote.var"#430#432"{1, Float64, Vector{Float64}, Tuple{UnitRange{Int64}}}})(Δ::NamedTuple{(:parent, :indices, :offset1, :stride1), Tuple{Vector{Float64}, Nothing, Nothing, Nothing}})
   @ Zygote ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:67
 [6] Pullback
   @ ./REPL[192]:1 [inlined]
 [7] (::Zygote.var"#52#53"{typeof(∂(#107))})(Δ::Vector{Float64})
   @ Zygote ~/.julia/packages/Zygote/rv6db/src/compiler/interface.jl:41
 [8] top-level scope
   @ REPL[193]:1
```

Is this the kind of problem that you've been encountering @yuehhua ?

yuehhua:
@willtebbutt Yes, that is the case I encounter.

@mcabbott @willtebbutt `view` is frequently used in my framework and `parent` is called to get the original array back. There is still an another way to do the same thing. I recieve a subarray after passing through some layers, and I want to put vectors back to their own indices. A new zero array can be created, get indices of these vectors and put them back. But, I can just call `parent` to avoid memory allocations.

mcabbott:
This still sound like the result ought to produce garbage, `view([99,10,20], 2:3) == view([10,20,NaN], 1:2)` but if you call `parent` you will get very different answers. Why won't this happen to you? Should your function rather accept both an array and some indices, if it needs both?

yuehhua:
The code is complicated and I simplified functions to get examples as follow:

```julia
function foo(graph, X, aggr)
    idx = get_indices(graph)
    return NNlib.scatter(aggr, X, idx)
end

function foo(graph, X, aggr::Nothing)
    idx = get_indices(graph)
    return view(X, :, idx)
end

function bar(graph, X, aggr)
    Y = foo(graph, X, aggr)
    return parent(Y)  # to put results back to orignal size
end
```

willtebbutt:
Before commenting further, I just want to make sure I've understood what your example is doing.

It looks to me like what you're doing is the identity function if `aggr` is `nothing`, and applying `aggr` to a subset of the elements in `X` otherwise, and returning (a copy of the resulting) `X`.

Have I understood correctly?

yuehhua:
What I want to express is that there would be a `aggr` arg, which controls which method to dispatch to. This will return two different kinds of results: one is regular array and the other is a subarray. If it returns a subarray, I want it turn to be the original array.

willtebbutt:
Okay. Is there a reason that you're implementing two methods of `foo`, rather than simply implementing two methods of `bar`, with the second something like
```julia
function bar(graph, X, aggr::Nothing)
    return X
end
```
?

yuehhua:
For a more complicated case (more close to my real code), it will be somwhat like this.

```julia
function foo1(graph, X, aggr)
    idx = get_indices(graph)
    return NNlib.scatter(aggr, X, idx)
end

function foo1(graph, X, aggr::Nothing)
    idx = get_indices(graph)
    return view(X, :, idx)
end

function foo2(graph, X, aggr)
    # do something similar to foo1
end

function foo2(graph, X, aggr::Nothing)
    # do something similar to foo1
end

function foo3(graph, X, aggr)
    # ...
end

function foo3(graph, X, aggr::Nothing)
    # ...
end

function bar(graph, X, aggr1, aggr2, aggr3)
    X2 = foo1(graph, X, aggr1)
    X3 = foo2(graph, X2, aggr2)
    Y = foo3(graph, X3, aggr3)
    return parent(Y)  # to put results back to orignal size
end
```

`bar` act like a backbone and can not be reduced. `foo1`, `foo2`, `foo3` can be truned on/off by dispatching different `aggr`.

For the real code, `bar` is the analog to `propagate` [here](https://github.com/FluxML/GeometricFlux.jl/blob/27c2ce434d442eca227f6530cc8157b80af5d9f8/src/layers/gn.jl#L70).

willtebbutt:
Thanks for taking the time to explain your example more fully. This still feels like a slightly odd example, but it's perfectly valid Julia, so in my view we ought to be able to AD it.

That being said, I don't think that the problem is really in ChainRules per-se. Rather, the [adjoint for view in Zygote](https://github.com/FluxML/Zygote.jl/blob/169563f65f0d9f9221ad528780aab48d157a05fc/src/lib/array.jl#L28) assumes that the cotangent of a view is going to be an AbstractArray of some sort, which isn't always the case.

I found that the following implementation of the `view` adjoint works locally on a small example:
```julia
@adjoint function view(x::AbstractArray, inds...)
  view_pullback(Δ::AbstractArray) = ∇getindex(x, inds)(Δ)
  view_pullback(Δ::NamedTuple) = (Δ.parent, nothing)
  return view(x, inds...), view_pullback
end
```
@yuehhua could you dev Zygote and try this out locally on your problems? If it solves your problem, I would suggest opening a PR to Zygote and tagging us in it.

mcabbott:
Thanks for the example. This still seems a bit weird to me. 

Without AD what you've written seems to be a way to apply a sequence of operations specified by `aggr1, aggr2, aggr3`, each of which is either some function or `nothing`, with the instruction that you should discard the last operation(s) in this chain if they have `nothing`. The implementation uses SubArrays, but it doesn't use them only as lightweight alternatives to `getindex`, since the "masked off" information can still reach the output (sometimes).

However, the gradient rules don't know this, the gradient for `view` is the same as for `getindex`. And so the gradient for the function will not be correct. This is not caused by the rule for parent being wrong; it would be the same if the rule for `getindex` could digest a `Tangent` as Will suggests.

A self-contained example of this is:
```julia
using Zygote, ForwardDiff, ChainRulesCore

function ChainRulesCore.rrule(::typeof(parent), A::Base.SubArray)  # this PR
    parent_pullback(Δ) = (NoTangent(), view(Δ, A.indices...))
    parent(A), parent_pullback
end

function withparent(x)
     # Chain of views and operations, alternating:
     y = @view x[1:2:end]
     z = exp.(y) 
     t = view(z, 1:2)
     # Un-do the last view(s) in the chain:
     parent(t)
end

withparent(1:7)  # 4-element vector, == exp.(1:2:7)

ForwardDiff.gradient(sum∘withparent, collect(1:7))  # 4 nonzero entries

Zygote.gradient(sum∘withparent, collect(1:7))[1]    # only 2 nonzero entries
```

yuehhua:
I found another way bypassing this issue. Thanks.

PR #563, opened 2022-01-13
Test for things adding rules to DataType/UnionAll

oxinabox:
This should let us catch errors like
https://github.com/JuliaDiff/ChainRules.jl/pull/562
before they are merged.

idk if this is the best way, feedback appreciated.

The output this will show is:
```
┌ Error: Bad constructor rrule. typeof(T)` not `Type{T}`
│   method = rrule(::UnionAll, x::AbstractArray, perm) in ChainRules at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/src/rulesets/Base/array.jl:84
└ @ Main ~/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:24
Test Failed at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:25
  Expression: false
ERROR: LoadError: There was an error during testing
in expression starting at /home/oxinabox/JuliaEnvs/ChainRulesWorld/ChainRules.jl/test/global_state.jl:14
```

This PR will fail until #562 is merged. We should merge one of these then rebase the other


test/global_state.jl
```diff
@@ -0,0 +1,44 @@
...
+        function_type = if method.sig <: Tuple{typeof(rrule), RuleConfig, Type, Vararg}
+            parameters(method.sig)[3]
+        elseif method.sig <: Tuple{typeof(rrule), Type, Vararg}
+            parameters(method.sig)[2]
+        else
+            nothing
+        end
+        
+        if function_type == DataType || function_type == UnionAll || function_type == Union
+            @error "Bad constructor rrule. typeof(T)` not `Type{T}`" method
```
mzgubic, line 24:
```suggestion
            @error "Bad constructor rrule. Use `Type{T}` not `typeof(T)`" method
```

mzgubic (commented):
Adding this test here will ensure that no such rule is added to ChainRules, but users writing their own rules could still break things for themselves. It might be helpful to add this to CRTU instead/as well?

oxinabox:
> Adding this test here will ensure that no such rule is added to ChainRules, but users writing their own rules could still break things for themselves. It might be helpful to add this to CRTU instead/as well?

Yeah, I guess we can put this into a function in CRTU and then just call it here

test/global_state.jl
mcabbott: Or `Type{<:T}` probably?

mcabbott:
Maybe `test_rrule` could check `which(rrule, args...)`? If I understand this version right, it'll complain that something is broken, but won't tell you which rule. Perhaps ideally you want both, since the presence of such a rule (even if it never meets `test_rrule`) is a global problem.

mzgubic (approved):

test/runtests.jl
```diff
@@ -74,6 +76,6 @@ end
     println()
 
     include_test("rulesets/Random/random.jl")
-
+
```
oxinabox, line 79:
```suggestion
```

PR #564, opened 2022-01-13
use begin/end for linear indexing in _evalpoly_indeterminates

jverzani:
This uses `begin`/`end` for linear indexing in `_evalpoly_indeterminates`. This should close https://github.com/JuliaMath/Polynomials.jl/issues/313. (The object passed has 0-based indexing for which the current usage is not appropriate.)

### DIFF
--- a/src/rulesets/Base/evalpoly.jl
+++ b/src/rulesets/Base/evalpoly.jl
@@ -144,7 +144,7 @@ if VERSION ≥ v"1.4"
                 ∂p[i] = _evalpoly_backp(p[firstindex(p) - 1 + i], ∂yi)
                 ∂yi = x′ * ∂yi
             end
-            ∂p[N] = _evalpoly_backp(p[end], ∂yi)
+            ∂p[N] = _evalpoly_backp(p[lastindex(p)], ∂yi)
         end
         return ∂x, ∂p
     end
--- a/src/rulesets/Base/evalpoly.jl
+++ b/src/rulesets/Base/evalpoly.jl
@@ -61,6 +61,7 @@ if VERSION ≥ v"1.4"
         return y, ys
     end
     function _evalpoly_intermediates(x, p)
+        @show :hi, p
         N = length(p)
         @inbounds yn = one(x) * p[lastindex(p)]
         ys = similar(p, typeof(yn), N - 1)
@@ -133,7 +134,7 @@ if VERSION ≥ v"1.4"
         x′ = x'
         ∂yi = one(x′) * Δy
         N = length(p)
-        @inbounds ∂p1 = _evalpoly_backp(p[1], ∂yi)
+        @inbounds ∂p1 = _evalpoly_backp(p[firstindex(p)], ∂yi)
         ∂p = similar(p, typeof(∂p1))
         @inbounds begin
             ∂x = _evalpoly_backx(x, ys[N - 1], ∂yi)
@@ -141,10 +142,10 @@ if VERSION ≥ v"1.4"
             ∂p[1] = ∂p1
             for i in 2:(N - 1)
                 ∂x = _evalpoly_backx(x, ys[N - i], ∂x, ∂yi)
-                ∂p[i] = _evalpoly_backp(p[i], ∂yi)
+                ∂p[i] = _evalpoly_backp(p[firstindex(p) - 1 + i], ∂yi)
                 ∂yi = x′ * ∂yi
             end
-            ∂p[N] = _evalpoly_backp(p[N], ∂yi)
+            ∂p[N] = _evalpoly_backp(p[end], ∂yi)
         end
         return ∂x, ∂p
     end
--- a/src/rulesets/Base/evalpoly.jl
+++ b/src/rulesets/Base/evalpoly.jl
@@ -62,13 +62,13 @@ if VERSION ≥ v"1.4"
     end
     function _evalpoly_intermediates(x, p)
         N = length(p)
-        @inbounds yn = one(x) * p[end]
+        @inbounds yn = one(x) * p[lastindex(p)]
         ys = similar(p, typeof(yn), N - 1)
         @inbounds ys[1] = yn
         @inbounds for i in 2:(N - 1)
-            ys[i] = muladd(x, ys[i - 1], p[end - i + 1])
+            ys[i] = muladd(x, ys[i - 1], p[lastindex(p) - i + 1])
         end
-        @inbounds y = muladd(x, ys[N - 1], p[begin])
+        @inbounds y = muladd(x, ys[N - 1], p[firstindex(p)])
         return y, ys
     end
 
--- a/src/rulesets/Base/evalpoly.jl
+++ b/src/rulesets/Base/evalpoly.jl
@@ -3,11 +3,11 @@ if VERSION ≥ v"1.4"
     function frule((_, ẋ, ṗ), ::typeof(evalpoly), x, p)
         Δx, Δp = ẋ, unthunk(ṗ)
         N = length(p)
-        @inbounds y = p[N]
+        @inbounds y = p[end]
         Δy = Δp[N]
         @inbounds for i in (N - 1):-1:1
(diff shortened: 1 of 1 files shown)


src/rulesets/Base/evalpoly.jl
```diff
@@ -62,13 +62,13 @@ if VERSION ≥ v"1.4"
...
-        @inbounds yn = one(x) * p[N]
+        @inbounds yn = one(x) * p[end]
         ys = similar(p, typeof(yn), N - 1)
         @inbounds ys[1] = yn
         @inbounds for i in 2:(N - 1)
-            ys[i] = muladd(x, ys[i - 1], p[N - i + 1])
+            ys[i] = muladd(x, ys[i - 1], p[end - i + 1])
         end
-        @inbounds y = muladd(x, ys[N - 1], p[1])
+        @inbounds y = muladd(x, ys[N - 1], p[begin])
```
nickrobinson251, line 71: i don't think `p[begin]` was valid syntax in Julia v1.0 (which ChainRules still supports), but i think `firstindex` achieves the same:
```suggestion
        @inbounds y = muladd(x, ys[N - 1], p[firstindex(p)])
```

jverzani: Thanks, I just noticed that in testing.

jverzani:
Should I bump the version number here? (Unfamiliar with VersionVigilante)

jverzani:
Sure, will do soon.

jverzani:
Hi @mcabbott 

I'm a little stuck and could use some help with writing a test.

First, unrelated to this PR, I can reliably cause a crash on my system with the following (either the first or second run of the last line)

```
using Revise
using OffsetArrays
using Zygote
using Test

v = [1.0, 2.0, 3.0] # 1 + 2x + 3x^2
f(x) = evalpoly(x, v)
f′(x) = evalpoly(x, (2.0, 6.0))

𝒗 = OffsetVector(v, 0:2)
𝒇(x) = evalpoly(x, 𝒗)


x = 3.0
@test Zygote.gradient(𝒇, x)[1] == f′(x)
```

I then tried to make a minimal polynomial type. The resulting test works correctly under this PR:

```
using Revise
using OffsetArrays
using Zygote
using Test

# 0-based indexed object evaluated with evalpoly
struct MinPoly{T}
    v::Vector{T}
end

Base.firstindex(p::MinPoly) = 0
Base.lastindex(p::MinPoly) = length(p.v)-1
Base.getindex(p::MinPoly{T}, i) where {T} = (i + 1) <= length(p) ? p.v[i+1] : zero(T)
Base.length(p::MinPoly) = length(p.v)
Base.similar(p::MinPoly, args...) = similar(p.v, args...)
Base.collect(p::MinPoly) = collect(p.v)

function Base.evalpoly(x::S, p::MinPoly{T}) where {S, T}
    Σ = one(S) * p[lastindex(p)]
    for i ∈ (lastindex(p)-1):-1:firstindex(p)
        Σ = muladd(Σ, x, p[i])
    end
    Σ
end

(p::MinPoly)(x) = evalpoly(x, p)

# ---
v = [1.0, 2.0, 3.0] # 1 + 2x + 3x^2
f(x) = evalpoly(x, v)
f′(x) = evalpoly(x, (2.0, 6.0))

x = 3.0
p = MinPoly(v)
@test Zygote.gradient(p, x)[1] == f′(x)
```

**However**, I can't get the following tests to work (following the pattern in the testing suite):

```
using ChainRules
using ChainRulesTestUtils

test_frule(evalpoly, x, p)
test_rrule(evalpoly, x, p)
```

I can't quite tell what is expected of these tests.

mcabbott:
Ok I can reproduce the crash! :( 

Note BTW that Base doesn't support offset arrays here, and will silently give wrong answers. (Tracked at https://github.com/JuliaLang/julia/issues/40454 I see.)
```
julia> evalpoly(10, [1.0, 2.0, 3.0])
321.0

julia> evalpoly(10, OffsetArray([1.0, 2.0, 3.0], -1))
332.0

julia> evalpoly(10, OffsetArray([1.0, 2.0, 3.0], +2))
ERROR: BoundsError: attempt to access 3-element OffsetArray(::Vector{Float64}, 3:5) with eltype Float64 with indices 3:5 at index [2]
```
The first question seems to be what the correct answer is here. I think this PR takes the view that the list of coefficients is an iterator, starting from the constant term, and its indices are irrelevant. But the alternative would be to regard non-standard indexing as a way to encode a Laurent series, etc, i.e. to start the polynomial at a possibly negative power. ChainRules must follow what Base decides, for `::AbstractVector`, but at the moment there is no decision.

For types you own, like `MinPoly`, you can of course specify any rules you wish. Especially if they are not `<:AbstractArray`. I wonder if the appropriate thing to do here isn't to insert `Base.require_one_based_indexing` into the rule, and have another rule `rrule(::typeof(evalpoly), x::Number, p::MinPoly) = ...` which un-wraps, calls the rule, re-wraps the answer (in some `Tangent{MinPoly}`). ?

src/rulesets/Base/evalpoly.jl
```diff
@@ -3,11 +3,11 @@ if VERSION ≥ v"1.4"
     function frule((_, ẋ, ṗ), ::typeof(evalpoly), x, p)
         Δx, Δp = ẋ, unthunk(ṗ)
         N = length(p)
-        @inbounds y = p[N]
+        @inbounds y = p[end]
         Δy = Δp[N]
         @inbounds for i in (N - 1):-1:1
             Δy = muladd(Δx, y, muladd(x, Δy, Δp[i]))
-            y = muladd(x, y, p[i])
+            y = muladd(x, y, p[firstindex(p) - 1 + i])
```
sethaxen, line 10: As is, my concern is this will need to re-check `firstindex` for each loop iteration. Why not compute `firstindex(p) - 1` once outside the loop and then re-use inside? Or better yet, compute the range `(firstindex(p) + N - 2):-1:firstindex(p)

sethaxen:
I haven't looked at the linked issue. The rules for `evalpoly` are directly based on manually AD-ing the fallback primal in base Julia and therefore make the same assumptions:
https://github.com/JuliaLang/julia/blob/ac5cc99908d463582e66db3368b9b48fae1e2525/base/math.jl#L152-L174

Note that the primal is only implemented for `AbstractVector` and `Tuple` `p`, but our rule doesn't have that restriction. We should probably make it. The rule will do the wrong thing for any type for which the primal does the wrong thing, and we have the usual problem that if a user adds an `evalpoly` overload for their type, the AD will still hit our rule, which can be a problem anywhere.

sethaxen:
In fact, the original issue is that our rule is hitting a non-`AbstractVector` type, right? Then we can fix that easily by just adding the type constraints from Base and worry about the indexing in a different issue.

mcabbott:
Yes, I think this rule should only be defined for `Union{Tuple, AbstractVector}` to match Base.
```julia
julia> methods(evalpoly)
# 5 methods for generic function "evalpoly":
[1] evalpoly(z::Complex, p::Tuple{Any}) in Base.Math at math.jl:225
[2] evalpoly(z::Complex, p::Tuple) in Base.Math at math.jl:200
[3] evalpoly(x, p::Tuple) in Base.Math at math.jl:176
[4] evalpoly(z::Complex, p::AbstractVector) in Base.Math at math.jl:228
[5] evalpoly(x, p::AbstractVector) in Base.Math at math.jl:189
```
I also think it should throw an error on an offset index. Even though base does not always throw, it should... until there is explicit behaviour for offset indices.

jverzani:
Okay, that makes sense. I hadn't realized Base didn't work as expected until you pointed that out and that is the documented behavior (1-based indexing for the coefficients: `Evaluate the polynomial \sum_k x^{k-1} p[k] for the coefficients p[1], p[2], ...`). I'd argue this should be relaxed in base to any iterable yielding the polynomials coefficients in a standard basis, but that isn't an argument for this package.

Anyways, thanks for your review and thought here.

PR #566, opened 2022-01-14
Less memory-intensive pmap adjoint

gaurav-arya:
The current `pmap` adjoint in `Zygote.jl` collects all the pullbacks from the workers on the root process, which is very memory-intensive. Instead, we'd like to persist the pullbacks on the workers, and use them in the backward pass. This requires the backward pass to use the same matching of workers to array indices as the forward pass, as opposed to load balancing.


ChrisRackauckas:
I think for now if it's not specializing on batch_size > 1, we can just have an issue opened for that.

ChrisRackauckas:
This should be the solution to https://github.com/SciML/DiffEqFlux.jl/issues/673. 

Also @DhairyaLGandhi we should make sure to remove the Zygote pmap derivative once this is in.

ChrisRackauckas:
This needs tests.

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
+function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(pmap), f, p::AbstractWorkerPool, c; kwargs...)
+    key = uuid1()
+    function forw(i, a)
+        y, back = rrule_via_ad(config, f, a)
+        save_backs[(key,i)] = back;
+        return y, myid()
+    end
+
+    I = reshape(eachindex(c), size(c))
+    ys_IDs = pmap(forw, I, c; kwargs...)
```
mcabbott, line 551: I think the claim is that doing `pmap(a -> rrule_via_ad(config, f, a), c)` makes an array on the head process. Is there a simple way to check this, and to benchmark things?

I also wonder: for what kind of `f` does this matter? Should this maximal solution come after you've checked that there are't easier ones, somehow? e.g. for sufficiently pure `f`.

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
+# it should be fine to deal with only a single collection c
+function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(pmap), f, p::AbstractWorkerPool, c; kwargs...)
+    key = uuid1()
+    function forw(i, a)
+        y, back = rrule_via_ad(config, f, a)
+        save_backs[(key,i)] = back;
+        return y, myid()
+    end
+
+    I = reshape(eachindex(c), size(c))
```
mcabbott, line 550: Are any non-array collections allowed? If not, it should probably be `::AbstractArray` in the signature. And `axes` not size here. 

Also, can I vote to call the big collection a bigger letter than `c`? Like A, of at least C?

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
+    function pmap_pullback(ȳ)
+        # TODO: fails when spawn on own processor
+        remote_calls = map((i, ID, ȳ) -> (@spawnat ID save_backs[(key,i)](ȳ)), I, IDs, ȳ)
+        res = map(fetch, remote_calls)
+        for (i, ID) in zip(I, IDs)
+            @spawnat ID delete!(save_backs, (key, i))
+        end
+        f̄ = sum(first, res)
+        c̄ = map(last, res)
+        (nothing, f̄, nothing, c̄)
```
mcabbott, line 564: Will need `(NoTangent(), f̄, NoTangent(), c̄)`, and the robot will want a `return`.

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
     end
     return y, extrema_pullback_dims
 end
+
+#####
+##### `pmap`
+#####
+
+using UUIDs # of course we don't want this;
```
DhairyaLGandhi, line 535: Use `gensym` maybe?

src/rulesets/Base/array.jl
DhairyaLGandhi: `f` may contain data so it's possibly better to avoid serializing it often. We found this issue in Dagger.jl as well, so it seems better not to assume that serializing `f` is free.

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
+
+save_backs = Dict() # could this be a stack instead?
+
+# Now that there is a backwards rule for zip (albeit only in Zygote),
+# it should be fine to deal with only a single collection c
+function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(pmap), f, p::AbstractWorkerPool, c; kwargs...)
+    key = uuid1()
+    function forw(i, a)
+        y, back = rrule_via_ad(config, f, a)
+        save_backs[(key,i)] = back;
```
DhairyaLGandhi, line 546: Is the claim that you only want to hold on to the backwards pass? That can probably be done with `Zygote._pullback`

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
+    ys_IDs = pmap(forw, I, c; kwargs...)
+    ys = map(first, ys_IDs)
+    IDs = map(last, ys_IDs)
+
+    function pmap_pullback(ȳ)
+        # TODO: fails when spawn on own processor
+        remote_calls = map((i, ID, ȳ) -> (@spawnat ID save_backs[(key,i)](ȳ)), I, IDs, ȳ)
+        res = map(fetch, remote_calls)
+        for (i, ID) in zip(I, IDs)
+            @spawnat ID delete!(save_backs, (key, i))
```
DhairyaLGandhi, line 560: `delete` ing seems to be rough. Also the loop needs to be wrapped in`@sync`

DhairyaLGandhi:
We can simulate this on the Zygote level as well. Holding data on the workers should be doable.

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
 end
+
+#####
+##### `pmap`
+#####
+
+using UUIDs # of course we don't want this;
+            # what is a better way of uniquely identifying function calls from the forward pass?
+
+save_backs = Dict() # could this be a stack instead?
```
DhairyaLGandhi, line 538:
```suggestion
const save_backs = Dict() # could this be a stack instead?
```

src/rulesets/Base/array.jl
gaurav-arya: The issue this rule resolves is when the pullback function (`back`) is a closure over a lot of data, significantly more than the size of `y`. Roughly speaking, I think this should happen when `f` contains several non-linear steps so that a lot of data needs to be remembered for the backward pass. This can happen with completely pure `f`, so it may be difficult to check?

src/rulesets/Base/array.jl
gaurav-arya: I was hoping to handle arbitrary collections (so perhaps `I = reshape(1:length(A), axes(A))`) instead), but that would complicate things in the pullback (I'm not sure how to express the tangent of an arbitrary collection, right now I'm just returning an array). So I'll add an `::AbstractArray` signature for now.

src/rulesets/Base/array.jl
mcabbott: Yes, I guess pure isn't it, as you say.

One check is whether `ChainRulesCore.derivatives_given_output` returns something, in which case you need not store the pullback at all. But perhaps something as cheap as `pmap(log, x)` is something nobody will ever do.

How can you test whether this copy back has happened?

src/rulesets/Base/array.jl
gaurav-arya: I did actually originally write this as a Zygote rule! See this gist here: https://gist.github.com/gaurav-arya/9f78bd49a65a3b1219a1f21cf87fe722

src/rulesets/Base/array.jl
```diff
@@ -527,3 +527,42 @@ function _extrema_dims(x, dims)
...
 end
+
+#####
+##### `pmap`
+#####
+
+using UUIDs # of course we don't want this;
+            # what is a better way of uniquely identifying function calls from the forward pass?
+
+const save_backs = Dict() # could this be a stack instead?
```
mcabbott, line 538: Does it matter that a Dict isn't thread safe?

If not, then the obvious way to label entries is just to increment `const COUNTER = Ref(1)`.

gaurav-arya: Took a stab at this!

gaurav-arya: I do worry about the thread-safe issue though. Perhaps it would be best to explicitly restrict to `batch_size = 1`, and port over the Zygote rule to run as a backup for `batch_size > 1`?

mcabbott: You could surely make it thread-safe, via Atomic & all that. But maybe this becomes slow?

I guess I still don't know why one global Dict is better than one Array... which can be created within the rule & then everyone knows where to write. Why is a Dict better? How do you test that this is working?

gaurav-arya: I... don't know why I used a global Dict. I can't see any issues with declaring something within the `rrule`. I'll give that a try now.

gaurav-arya: I guess it makes initializing the array on the processors a little more complex -- do I want to initialize it on every processor in the worker pool before the `pmap`, or initialize it inside of `forw` if it isn't already defined?

gaurav-arya: Oh, I think I see why I used a global dictionary. My understanding is that the worker processors don't know anything about the scoping on the root processor. So, to persist state on a worker process, the simplest solution is to just have a global variable. Does that make sense?

src/rulesets/Base/array.jl
gaurav-arya: Hmm. I've modified the code to use a counter instead of a uuid, but might `gensymb` be an even better solution? Could it be used to create an array with a unique name on all of the worker processors, as opposed to having this array reside in `save_backs[key]` like it currently does? (I don't have any experience with meta-programming in Julia)

gaurav-arya: Do you have any pointers on how I might use 'gensym' here? I've never worked with Julia meta-programming. 

(I've currently replaced uuid with a counter, but I just realized that this would fail in the case of nested 'pmap''s as different processors' counters would start conflicting in the dictionary :/)

gaurav-arya:
Here are some PR to-do's:

- Tests should be more comprehensive, and should test things like nested `pmap`'s, sequential `pmap`s, etc. It seems difficult to this just within ChainRules; would it be a good idea to add such tests downstream in Zygote?

- Resolve any issues with how I'm storing the pullbacks (global dictionary of arrays, with a particular `rrule` call uniquely identifier with `(myid(), COUNTER[]` of root process). I don't know of a better solution, but would welcome suggestions.

- The tests complain about some type inference issues that I don't know how to fix. I'd appreciate help with these, as well as with making sure other `ChainRules` best practices like projecting, handling thinks, etc. are followed

- Possibly specialize for simpler functions `f` to e.g. avoid storing pullbacks. I'm not familiar enough with the experimental features of `ChainRulesCore` to do this correctly.

And then the long-term change which won't happen this PR: replace the fallback for `batch_size > 1` with something that is also not memory intensive.

ChrisRackauckas:
It looks like https://github.com/SciML/DiffEqFlux.jl/issues/673 @JordiBolibar 's issue is out of scope for the PR, but it would still be good to get it in

ChrisRackauckas:
I guess the question I would have is, what's the minimal version we could get in today? I hate stale PRs, they are a waste of everyone's time. Is it easier to just do it direct in Zygote? If so, why not just get that done and wait to generalize later?

gaurav-arya:
Hey, I think the shortest path to getting this merged in somewhere would be to address the type inference issues that fail the tests (presumably simple to address, but I don't know how) and merge it here in `ChainRules`. 

I currently don't have the bandwidth to work on it :( So the alternative would be to leave this PR stale so that I can polish it, generalize it, etc. in a week or two when I am able to

mcabbott:
I don't think it can be merged here without more serious. And, as I said above, I'm not sure how to review this without examples / benchmarks which show that all of this is doing something. Maybe Zygote takes a more relaxed view...

gaurav-arya:
Makes sense. I'll revisit this PR when I can then, and address the to-do's I noticed above. Feel free to close it or leave it as a draft, I don't know what the more appropriate action is

JordiBolibar:
Hi, any news on this fix? I'm really looking forward to this one 😄

gaurav-arya:
Hey! I'm gonna give it another stab this weekend 🙂

gaurav-arya:
Status update: batch sizes > 1 should be handled now. I'll work on getting a benchmark soon. For more elaborate tests of e.g. nested `pmap`s or multiple `pmap` in a row, it seems hard to do it using only ChainRulesTestUtils, so perhaps these could go upstream in Zygote. 

However, the way I'm persisting pullbacks doesn't seem right. For one, it doesn't support distributed garbage collection easily. What would really help is if I could discuss the details of the PR with someone who is very familiar with the distributed computing ecosystem of Julia i.e. things like remote calls, remote references, etc. so that I can get it done properly.

ChrisRackauckas:
I won't have the time, but maybe @vchuravy has a second?

gaurav-arya:
Benchmark: [benchmark.pdf](https://github.com/JuliaDiff/ChainRules.jl/files/8019180/benchmark.pdf)

Companion Zygote PR: coming soon

Additional Notes:

- I realized that the counter+dictionary logic I had come up with to persist state was basically identical to [what is done in DistributedArrays.jl](https://github.com/JuliaParallel/DistributedArrays.jl/blob/master/src/core.jl). So I've switched to depend on `DistributedArrays`, which should hopefully make the code more readable.
- Garbage collection on the distributed processors should now be taken care of by `DistributedArrays`. However, there is the following caveat from the [docs](https://juliaparallel.github.io/DistributedArrays.jl/stable/#Garbage-Collection-and-DArrays): "since the size of the DArray object itself is small, a problem arises as gc on the master faces no memory pressure to collect the DArray immediately. This results in a delay of the memory being released on the participating workers."
- I've tried to handle arbitrary collections `X`. (In particular, it's crucial to handle `X <: Base.Iterators.Zip`.) A look through of my code with an eye on type stability / genericness / correct use of projections would be great.
- There are some basic tests here in `ChainRules` -- feel free to add to these as needed. In addition, there are tests in `Zygote` for functions with multiple `pmap`'s, nested `pmap`'s, and `pmap`'s called with multiple collections (in which case the `rrule` here is chained with the `Iterators.Zip` adjoint in `Zygote`).

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,59 @@
...
+    indices = getindex.(ys_IDs_indices, 3) # remember the index of the pullback in the array on each processor
+ 
+    # create a list of positions in X handled by each processor
+    unique_IDs = sort(unique(IDs))
+    positions = [Vector{eltype(eachindex(ys_IDs_indices))}() for _ in 1:length(unique_IDs)] # I don't understand type stability well, is this bad :/
+    for i in eachindex(ys_IDs_indices)
+        push!(positions[searchsortedfirst(unique_IDs, IDs[i])], i)
+    end
+
+    pmap_pullback = function (Ȳ)
```
CarloLucibello, line 30:
```suggestion
    function pmap_pullback(Ȳ)
```

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    ys_IDs_indices = pmap(forw, p, X; kwargs...)
+    ys = getindex.(ys_IDs_indices, 1) # the primal values
+    IDs = getindex.(ys_IDs_indices, 2) # remember which processors handled which elements of X
+    indices = getindex.(ys_IDs_indices, 3) # remember the index of the pullback in the array on each processor
+ 
+    # create a list of positions in X handled by each processor
+    unique_IDs = sort(unique(IDs))
+    positions = [Vector{eltype(eachindex(ys_IDs_indices))}() for _ in 1:length(unique_IDs)] # I don't understand type stability well, is this bad :/
```
CarloLucibello, line 25:
```suggestion
    T = eltype(eachindex(ys_IDs_indices))
    positions = [Vector{T}() for _ in 1:length(unique_IDs)] # I don't understand type stability well, is this bad :/
```

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,59 @@
+#####
+##### `pmap`
+#####
+
+# Now that there is a backwards rule for zip (albeit only in Zygote),
+# it should be fine to deal with only a single collection X
+function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(pmap), f, p::AbstractWorkerPool, X; batch_size=1, kwargs...)
```
CarloLucibello, line 7: doesn't seem the batch_size is being used

src/ChainRules.jl
```diff
@@ -10,6 +10,8 @@ using Random
 using RealDot: realdot
 using SparseArrays
 using Statistics
+using Distributed
+using DistributedArrays
```
mcabbott, line 14: Distributed is a standard library but DistributedArrays is a package with these dependencies:

https://juliahub.com/ui/Packages/DistributedArrays/Gfdem/0.6.6?page=1

I don't know what others think about adding dependencies. What exactly is gained, what are the costs?

src/ChainRules.jl
```diff
@@ -30,6 +32,7 @@ include("rulesets/Base/arraymath.jl")
 include("rulesets/Base/indexing.jl")
 include("rulesets/Base/sort.jl")
 include("rulesets/Base/mapreduce.jl")
+include("rulesets/Base/distributed.jl")
 
 include("rulesets/Statistics/statistics.jl")
```
mcabbott, line 37: Distributed is on the same level as Statistics, not under Base

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,61 @@
...
+    function forw(x)
+        y, back = rrule_via_ad(config, f, x)
+        push!(darr[:L][1], back)
+        return y, myid(), length(darr[:L][1])
+    end
+
+    ys_IDs_indices = pmap(forw, p, X; kwargs...)
+    ys = getindex.(ys_IDs_indices, 1) # the primal values
+    IDs = getindex.(ys_IDs_indices, 2) # remember which processors handled which elements of X
+    indices = getindex.(ys_IDs_indices, 3) # remember the index of the pullback in the array on each processor
```
mcabbott, line 21: What types are these? 

Is it worth making arrays from the array of tuples, not just directly reading them out as needed?

But more, since `forw` already mutates something as it goes along, to store the pullbacks, is it necessary to also create an array of tuples? Or could the details similarly be stored as it goes?

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,61 @@
...
+    output_sz = size(ys)
+ 
+    # create a list of positions in X handled by each processor
+    unique_IDs = sort(unique(IDs))
```
mcabbott, line 25: What's gained by sorting these?

And why not `sort!(unique!(`?

test/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,12 @@
+@testset "pmap" begin
+    test_rrule(pmap, inv, WorkerPool(), rand(10), check_inferred=false) # test empty worker pool
+    test_rrule(pmap, inv, default_worker_pool(), rand(10), check_inferred=false)
+    test_rrule(pmap, inv, default_worker_pool(), rand(4, 4), check_inferred=false) # test matrix input
+    test_rrule(pmap, inv, default_worker_pool(), rand(10), fkwargs=(batch_size=2,), check_inferred=false) # test batch_size > 1
+    test_rrule(pmap, Multiplier(2.0), default_worker_pool(), rand(3), check_inferred=false) # test adjoint of f
```
mcabbott, line 6: There's a lot of pretty complicated code. Do these tests cover a reasonable list of edge cases? Like:
* What can be used instead of default_worker_pool()? Would they notice if this was wrong?
* Does it matter whether batch sizes divide the number of workers evenly? If `addprocs()` is called somewhere?
* Will it obey pmap's keywords? `distributed=false`, `retry_delays`?
* What functions `f` are allowed, can I `pmap` a Flux model? It can't guarantee an order, which places some restrictions. Can it give an error if these are violated?
* Or how about a function which isn't type-stable, or worse, sometimes gives `NoTangent()`? Will `similar(res_batches[1], output_sz)` get confused?
* Does this only work on arrays of numbers? Then it should not accept `X::Any`.

test/runtests.jl
```diff
@@ -17,11 +17,14 @@ using StaticArrays
 using Statistics
 using Test
 using JuliaInterpreter
+using Distributed
 
 union!(JuliaInterpreter.compiled_modules, Any[Base, Base.Broadcast, LinearAlgebra, Random, StaticArrays, Statistics])
 
 Random.seed!(1) # Set seed that all testsets should reset to.
 
+const CFG = ChainRulesTestUtils.ADviaRuleConfig() # for testing rules that call back into AD
```
mcabbott, line 26: If this moves it should be to "test_helpers.jl"

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,61 @@
+#####
+##### `pmap`
+#####
+
+# Now that there is a backwards rule for zip (albeit only in Zygote),
+# it should be fine to deal with only a single collection X
+function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(pmap), f, p::AbstractWorkerPool, X; kwargs...)
+    project_X = ProjectTo(X)
+
+    darr = dfill([], (nworkers(p) + 1,), vcat(myid(), workers(p))) # Include own proc to handle empty worker pool
```
mcabbott, line 10: Functions taken from a package should probably be qualified.

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,61 @@
...
+    function pmap_pullback(Ȳ)
+        Ȳ = unthunk(Ȳ)
+
+        # runs the pullback for each position handled by proc ID in forward pass
+        function run_backs(ID, positions)
+            Ȳ_batch = Ȳ[positions]
+            indices_batch = indices[positions]
+            res_batch = remotecall_fetch(() -> 
+                asyncmap((ȳ, i) -> darr[:L][1][i](ȳ), Ȳ_batch, indices_batch), # run all the backs in a local asyncmap
+                ID)
```
mcabbott, line 41: I don't think I know what this is doing.

Does look like it should be a `do` block though.

src/rulesets/Base/distributed.jl
```diff
@@ -0,0 +1,61 @@
...
+    # create a list of positions in X handled by each processor
+    unique_IDs = sort(unique(IDs))
+    T = eltype(eachindex(ys_IDs_indices))
+    positions = [Vector{T}() for _ in 1:length(unique_IDs)]
+    for i in eachindex(ys_IDs_indices)
+        push!(positions[searchsortedfirst(unique_IDs, IDs[i])], i)
+    end
+
+    function pmap_pullback(Ȳ)
+        Ȳ = unthunk(Ȳ)
```
mcabbott, line 33: I think we should not re-use the same symbol for something of a (potentially) different type.

mcabbott (commented):
I took a look but I'm not sure I can fully review this, I don't know much about Distributed stuff. And whether dependencies are OK is perhaps above my pay grade around here.

I feel it needs more serious tests & benchmarks. I presume the plan is to map pretty expensive functions, so perhaps many of the book-keeping bits wash out? But sometimes very expensive things over small arrays and sometimes moderately expensive things over large arrays?

src/rulesets/Base/distributed.jl
gaurav-arya: I still need to keep the original IDs around, but I can get rid of a mutation.

I sorted so that I could match each ID to the unique processor quickly with `searchsortedfirst`
IMG_0325
.

test/rulesets/Base/distributed.jl
gaurav-arya: - Any worker pool can be used; default_worker_pool is the default for `pmap` when none is provided.
- That doesn't matter; the rule just remembers the mapping that `pmap` chose to use in the forward pass and does the same in the backward pass. Here is a diagram:
IMG_0325
- I added a test for `distributed=false`. It won't error with `retry_delays`, but it won't do retries in the backward pass, which could be a problem, I'm not sure about those sorts of situations. In general, this rule makes assumptions about the robustness of worker process, e.g. that they don't die between the forward and backward pass
- Any function `f` should work
- It could; I don't think there's actually any point in aiming for type stability in there, so I should change that. Going to hold off on it for now because I think there may be bigger issues with projection in that last part
- The aim is to work for arbitrary collections. In particular, note that `pmap` handles multiple collections by calling the single-collection version with all the collections zipped. By writing this rule for an arbitrary collection, I especially wanted to handle zipped ararys.

mcabbott: Re 5, I don't mean type-stability for performance, I mean that you may create a container which doesn't hold what it needs to. Which is why you need to test with an `f` whose gradient is sometimes zero. 

Re 4, any `f`: Must it be defined `@everywhere`? Can it close over arrays defined only locally? 

Re 1, if "Any worker pool can be used", then ideally you would test that it does in fact use the pool specified, both forward and backward.

gaurav-arya: - You're right, I think that could definitely cause a bug.
- It can be any `f` that is valid for use in `pmap`.
- It does more than just using the same pool, it uses the exact same matching of workers to tasks.
- Sure, I could explicitly check that, although this isn't something that's really in danger of happening since it explicitly uses precisely the same IDs as the forward pass.

gaurav-arya:
Hey, I'm going to temporarily close this PR. I think that there are a lot of ChainRules-specific issues that I'm dealing with because there's no non-distributed rule in ChainRules. And dealing with these issues at the same time as the pretty complex distributed logic is pretty overwhelming. Hopefully I can re-open it once:

- A `pmap` rule is added in Zygote (where there is a `map` + a slower `pmap` rule to base the new rule on). I'll try to do this soon-ish
- A rule for `map` is added to ChainRules.

PR #572, opened 2022-01-17
Fix `getindex` bug

mzgubic:
Closes https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/233


mzgubic:
indeed, see https://github.com/JuliaDiff/ChainRulesCore.jl/pull/534

PR #578, opened 2022-01-21
use ProjectTo in Array addition

oxinabox:
the `reshape` was a primative version of ProjectTo i think?

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -392,9 +392,9 @@ end
 
 function rrule(::typeof(+), arrs::AbstractArray...)
     y = +(arrs...)
-    arr_axs = map(axes, arrs)
+    projects = map(ProjectTo, arrs)
     function add_pullback(dy)
-        return (NoTangent(), map(ax -> reshape(dy, ax), arr_axs)...)
+        return (NoTangent(), map(project->project(dy), projects)...)
     end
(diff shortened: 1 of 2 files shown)


mcabbott:
I see it has `test_rrule(+, randn(3), randn(3,1), randn(3,1,1))` for this reshape, but maybe test Diagonal + Matrix or something?

oxinabox:
I added 
```julia
        test_rrule(+, randn(3,3), Diagonal(randn(3)), randn(3,3,1))
        test_rrule(+, randn(3,3), Diagonal(randn(3)), Symmetric(randn(3,3)))
```

and interestingly both fail.
So some debugging to do there

mcabbott:
The bug is that this doesn't work:
```julia
julia> ProjectTo(Diagonal([1,2,3]))(randn(3,3,1))
ERROR: MethodError: no method matching (::ProjectTo{Diagonal, ... }}}}}}}})(::Array{Float64, 3})

julia> Diagonal(randn(3)) + randn(3,3,1)  # but this does
3×3×1 Array{Float64, 3}:
```

mcabbott:
The fix is part of https://github.com/JuliaDiff/ChainRulesCore.jl/pull/446, BTW. Which I think should be merged.

oxinabox:
Closing infavor of #783 because I can't be bothered rasing

PR #579, opened 2022-01-22
add `sparse` rrule

CarloLucibello:
Partial replacement for #246

cc @sethaxen


src/rulesets/SparseArrays/sparsematrix.jl
```diff
@@ -0,0 +1,18 @@
...
+        ΔV = project_V(ΔΩ[I .+ m .* (J .- 1)])
+        return NoTangent(), NoTangent(), NoTangent(), ΔV, NoTangent(), NoTangent(), NoTangent()
+    end
+
+    return sparse(I, J, V, m, n, combine), sparse_pullback
+end
+
+function rrule(::typeof(sparse), A::Union{AbstractVector, AbstractMatrix})
+    function sparse_pullback(Ω̄)
+        return NoTangent(), Ω̄
```
CarloLucibello, line 15: is this ok or we need to project or something?

src/rulesets/SparseArrays/sparsematrix.jl
```diff
@@ -0,0 +1,18 @@
...
+    function sparse_pullback(Ω̄)
+        ΔΩ = unthunk(Ω̄)
+        ΔV = project_V(ΔΩ[I .+ m .* (J .- 1)])
+        return NoTangent(), NoTangent(), NoTangent(), ΔV, NoTangent(), NoTangent(), NoTangent()
+    end
+
+    return sparse(I, J, V, m, n, combine), sparse_pullback
+end
+
+function rrule(::typeof(sparse), A::Union{AbstractVector, AbstractMatrix})
```
mcabbott, line 13: Should these rules be attached to `sparse` or one step later to things like `Type{<:SparseMatrixCSC}`? I don't have examples but catching all as many paths like `T(A)` and `convert(T, A)` etc. as possible sounds desirable.

CarloLucibello: We may also need rules for `SparseMatrixCSC`, but for sure we need a rule for `sparse` since it calls [sparse!](https://github.com/JuliaSparse/SparseArrays.jl/blob/8d397ccc5a2fa6ac6832623ae291cb3acff72d4f/src/sparsematrix.jl#L913) for the heavy-duty.

mcabbott: Yes, I meant for the 1-arg sparse. The one which takes all the vectors is a different story.

CarloLucibello: Ah sorry, didn't notice what you were pointing at. I'm not super confident in defining rrules for constructors.
Should it be `rrule(::Type{<:SparseMatrixCSC},...)` or `rrule(::Type{<:SparseMatrixCSC},...)` or the  two are equivalent?

mcabbott: I fiddle until it works, but my guess is that the first would capture all mode specific cases. Maybe not `convert` though.

CarloLucibello: ok I defined the rrules for the type and removed this one. I did check separately that Zygote's gradients of `sparse(A)` and `sparse(v)` go through

mcabbott (approved):
Looks good to me.

PR #588, opened 2022-02-11
Don't mutate cotangent in rule for `exp`

### LINKED ISSUE #512: Rule for `exp` mutates its input
Should these agree?
```
julia> using Zygote

julia> gradient(x -> sum(abs, exp(x)' + exp(x)'), [1 2; 3 4.0])
([169.91192742898497 320.5548284667717; 232.38736013199076 434.4143324629697],)

julia> gradient(x -> sum(abs, 2 * exp(x)'), [1 2; 3 4.0])
([226.54923657197995 427.40643795569565; 309.849813509321 579.2191099506263],)
```
I think the problem is that the rule for `exp` mutates the gradient it receives, which isn't safe as there's no guarantee this isn't shared with other rules -- here, the rule for + is doing `Δ -> (Δ, Δ)`.

From here https://github.com/JuliaDiff/ChainRules.jl/pull/381#issuecomment-902724655, but not sure when this was introduced. Haven't tried to check whether other rules do this too.

Xref https://github.com/FluxML/Zygote.jl/pull/981 maybe --- it would be nice to change the convention to mutate `Δ` freely by default, and copy only when necessary.

sethaxen:
Fixes #512.

I'm not certain how to test this, since I was unable to replicate in ChainRules the bug observed in Zygote in that issue, but with the fix here, we get
```julia
julia> using Zygote

julia> gradient(x -> sum(abs, exp(x)' + exp(x)'), [1 2; 3 4.0])
([226.54923657197995 427.40643795569565; 309.84981350932094 579.219109950626],)

julia> gradient(x -> sum(abs, 2 * exp(x)'), [1 2; 3 4.0])
([226.54923657197995 427.40643795569565; 309.84981350932094 579.219109950626],)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.26"
+version = "1.26.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/LinearAlgebra/matfun.jl
+++ b/src/rulesets/LinearAlgebra/matfun.jl
@@ -132,7 +132,7 @@ function rrule(::typeof(exp), A0::StridedMatrix{<:BlasFloat})
             # Ensures ∂X is mutable. The outer `adjoint` is unwrapped without copy by
             # the default _matfun_frechet_adjoint!
             ΔX = unthunk(X̄)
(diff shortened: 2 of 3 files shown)


sethaxen:
> You could add a test that explicitly calls the rrule & checks for mutation.

I had trouble testing before, but I found I could replicate if the cotangent is an `Adjoint`. Added a test.

> Ideally perhaps CRTU would check that the input is not mutated?

Agreed!

PR #592, opened 2022-02-24
revisit thunk/unthunk in sparse operations

CarloLucibello:
Can one generally assume that the incoming sensitivity in a rrule has already been unthunked? 
Should one `thunk`  a returned value in the pullback also in the case all other returns are `NoTangent`s?

### DIFF
--- a/src/rulesets/SparseArrays/sparsematrix.jl
+++ b/src/rulesets/SparseArrays/sparsematrix.jl
@@ -2,8 +2,7 @@ function rrule(::typeof(sparse), I::AbstractVector, J::AbstractVector, V::Abstra
     project_V = ProjectTo(V)
     
     function sparse_pullback(Ω̄)
-        ΔΩ = unthunk(Ω̄)
-        ΔV = project_V(ΔΩ[I .+ m .* (J .- 1)])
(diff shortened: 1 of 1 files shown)


oxinabox:
In short: no and no.
Both these should be in the docs.
Possibily we should add a cross reference around this point of the docs to the appropriate sections?
As these are good questions

> Can one generally assume that the incoming sensitivity in a rrule has already been unthunked?

No, one must assume that it may be thunked.
Which is frustrating, but required for us to get anything out of thunking at all.
(Really want to remove thunks one day, they are a hack and make things annoying)
Explination is in: https://juliadiff.org/ChainRulesCore.jl/dev/rule_author/writing_good_rules.html#Thunks

> Should one thunk a returned value in the pullback also in the case all other returns are NoTangents?

That is something we have discussed a few times.
Since probably the one the user is going to actually only want to use the one that is NoTangent.
And that means we shouldn't thunk it.
Bit this isn't explictly said one way or the other.
If it was it would be in https://juliadiff.org/ChainRulesCore.jl/dev/rule_author/writing_good_rules.html#Use-Thunks-appropriately
possibly further discussion could be added these.

PR #593, opened 2022-02-24
backport #571

mzgubic:
From: https://github.com/JuliaDiff/ChainRulesCore.jl/pull/542#issuecomment-1048951081

backports https://github.com/JuliaDiff/ChainRules.jl/pull/571 to ChainRules 1.21, as it was before https://github.com/JuliaDiff/ChainRules.jl/pull/577/files

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.21" 
+version = "1.21.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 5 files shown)


mzgubic:
the ReversePropagation error exists on master and seems unrelated.

mcabbott (approved):
This looks fine to me.

Is `backport-1.21` the right branch name? If there may be other things backported, perhaps the same branch should end up with several tags, in which case something like `julia1.0` not mentioning a version might be better?

devmotion:
> Is backport-1.21 the right branch name? If there may be other things backported, perhaps the same branch should end up with several tags, in which case something like julia1.0 not mentioning a version might be better?

But if you backport anything to this branch, you can only release it as 1.21.x, can't you? So `backport-1.21` seems reasonable IMO.

mcabbott:
Oh sorry, somehow I thought there was one more dot in that version number. 1.21 is fine.

PR #597, opened 2022-03-08
Mark allunique non_differentiable

niklasschmitz:
This PR marks the Bool-valued `Base.allunique` as non-differentiable.


niklasschmitz:
Thanks @mzgubic !

mcabbott:
Maybe `allequal` too? with Compat 3.42.0

mzgubic:
@niklasschmitz would you be willing to add what @mcabbott suggested?

niklasschmitz:
Sure, while we're at it :) added it, thanks @mcabbott for the pointer

PR #601, opened 2022-03-20
Fix `rrule` for  complex-valued determinants

### LINKED ISSUE #600: `rrule` for complex-valued `det` gives wrong result
The `rrule` for `det(U)` of a matrix where `det(U)` is a complex number is wrong.

The current implementation is here:
https://github.com/JuliaDiff/ChainRules.jl/blob/9023d898a0b957bd9b3baab6bc38b54822d6963a/src/rulesets/LinearAlgebra/dense.jl#L132

The correct rule is

```julia
∂x = x isa Number ? ΔΩ : conj(Ω) * ΔΩ * inv(x)'
```

That is, it has to use the conjugate of `Ω = det(U)`. The resulting gradients match what finite differences yields. The correct rule can also be verified analytically: https://nbviewer.org/gist/goerz/65b117c6c6bca0dd2a650d9b8b0b05a5

Here is a test script that illustrates the failure, both with `ChainRules` directly, and for an "upstream application" in Zygote:

```julia
using Zygote: Zygote
using FiniteDifferences: FiniteDifferences
using ChainRules
using ChainRulesTestUtils
using LinearAlgebra
using Test

const 𝕚 = 1im
const PRECISION = 1e-8   # how much does Zygote and FD have to agree?


function random_unitary(N=4)
    X = rand(N, N) + exp.(𝕚 * 2π * rand(N, N)) # Random
    H = X + X'                                # ⇒ Hermitian
    return exp(𝕚 * H)                         # ⇒ Unitary
end


function well_conditioned_matrix(N=4)
    A = randn(ComplexF64, N, N)
    return A * A' + I
end


function re_det(U)
    real(det(U))  # broken
end


function run_test_zygote(func; complex=true, make_matrix=random_unitary)

    if complex
        U0 = make_matrix()
    else
        U0 = real.(make_matrix())
    end
    fdm = FiniteDifferences.central_fdm(5, 1)

    zgrad = Zygote.gradient(func, U0)[1]
    fgrad = FiniteDifferences.grad(fdm, func, U0)[1]
    print(">>> Testing $(nameof(func))(U0, complex=$complex)... ")
    res_zygote = @test norm(zgrad - fgrad) < PRECISION
    if res_zygote isa Test.Pass
        println("OK")
    end
end


@testset "TEST" begin
    test_rrule(det, well_conditioned_matrix())
    test_rrule(det, random_unitary())
    run_test_zygote(re_det, complex=false)
    run_test_zygote(re_det)
end
```

With the current code, the script fails. Updating the `rrule` as described above makes it pass.

Currently, the test suite in `ChainRules` uses a "well-conditioned" matrix defined as

https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/05caea982e905902299acef4558ba4a7b4e5ad96/src/data_generation.jl#L2-L5

for the complex-valued test case. Unfortunately, even though this matrix is complex-valued, its determinant is not. This explains why the current tests are not catching this bug.

A better test case would be a random unitary matrix, which can be obtained from the current test matrix `B` as `exp(1im * (B + B'))`. Such a random unitary is also well-conditioned, but has a complex-valued determinant.

goerz:
Closes #600

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -129,7 +129,7 @@ frule((_, Δx), ::typeof(det), x::Number) = (det(x), Δx)
 function rrule(::typeof(det), x::Union{Number, AbstractMatrix})
     Ω = det(x)
     function det_pullback(ΔΩ)
-        ∂x = x isa Number ? ΔΩ : Ω * ΔΩ * inv(x)'
+        ∂x = x isa Number ? ΔΩ : conj(Ω) * ΔΩ * inv(x)'
         return (NoTangent(), ∂x)
     end
     return Ω, det_pullback
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -129,7 +129,7 @@ frule((_, Δx), ::typeof(det), x::Number) = (det(x), Δx)
 function rrule(::typeof(det), x::Union{Number, AbstractMatrix})
     Ω = det(x)
     function det_pullback(ΔΩ)
-        ∂x = x isa Number ? ΔΩ : conj(Ω) * ΔΩ * inv(x)'
+        ∂x = x isa Number ? ΔΩ : inv(x)' * dot(Ω, ΔΩ)
         return (NoTangent(), ∂x)
     end
     return Ω, det_pullback
--- a/test/rulesets/LinearAlgebra/dense.jl
+++ b/test/rulesets/LinearAlgebra/dense.jl
@@ -119,6 +119,12 @@
                 test_rrule(f, B)
             end
         end
+        @testset "$f(unitary matrix)" begin
+            B = generate_well_conditioned_matrix(ComplexF64, 4)
+            U = exp(1im * (B + B'))
+            test_frule(f, U)
+            test_rrule(f, U)
+        end
     end
     @testset "logabsdet(::Matrix{$T})" for T in (Float64, ComplexF64)
         B = randn(T, 4, 4)
--- a/test/rulesets/LinearAlgebra/dense.jl
+++ b/test/rulesets/LinearAlgebra/dense.jl
@@ -119,9 +119,9 @@
                 test_rrule(f, B)
             end
         end
-        @testset "$f(unitary matrix)" begin
+        @testset "$f(complex determinant)" begin
             B = randn(ComplexF64, 4, 4)
-            U = exp(1im * (B + B'))
+            U = exp(B - B')
             test_frule(f, U)
             test_rrule(f, U)
         end
--- a/test/rulesets/LinearAlgebra/dense.jl
+++ b/test/rulesets/LinearAlgebra/dense.jl
@@ -120,7 +120,7 @@
             end
         end
         @testset "$f(unitary matrix)" begin
-            B = generate_well_conditioned_matrix(ComplexF64, 4)
+            B = randn(ComplexF64, 4, 4)
             U = exp(1im * (B + B'))
             test_frule(f, U)
             test_rrule(f, U)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.28.0"
+version = "1.28.1"
 
 [deps]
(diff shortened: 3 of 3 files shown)


src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -129,7 +129,7 @@ frule((_, Δx), ::typeof(det), x::Number) = (det(x), Δx)
 function rrule(::typeof(det), x::Union{Number, AbstractMatrix})
     Ω = det(x)
     function det_pullback(ΔΩ)
-        ∂x = x isa Number ? ΔΩ : Ω * ΔΩ * inv(x)'
+        ∂x = x isa Number ? ΔΩ : conj(Ω) * ΔΩ * inv(x)'
```
sethaxen, line 132: Perhaps we should explicitly multiple the two scalars together. Also, this computes the product in the correct order for numbers for which multiplication isn't commutative (though probably the rule is not even correct for such numbers anyways):
```suggestion
        ∂x = x isa Number ? ΔΩ : inv(x)' * dot(Ω, ΔΩ)
```

test/rulesets/LinearAlgebra/dense.jl
```diff
@@ -119,6 +119,12 @@
                 test_rrule(f, B)
             end
         end
+        @testset "$f(unitary matrix)" begin
+            B = generate_well_conditioned_matrix(ComplexF64, 4)
+            U = exp(1im * (B + B'))
```
sethaxen, line 124: Does `B` need to be well-conditioned? Because if `Z` in `U=exp(Z)` is skew-Hermitian, then `U` is well-conditioned.

```suggestion
            B = randn(ComplexF64, 4, 4)
            U = exp(B - B')
```

sethaxen (commented):
Thanks for the fix! A couple very minor suggestions. And, can you also increment the patch version?

src/rulesets/LinearAlgebra/dense.jl
goerz: Sure, that looks good to me!

test/rulesets/LinearAlgebra/dense.jl
goerz: `B` does not have to be well-conditioned... in fact, `B` can be completely random. Any unitary matrix is well-conditioned (same as your `U=exp(B-B')`, I think). My actual use case that found this bug involves unitary matrices, so that's why I went with a unitary test matrix.

The only reason I was using the `B` as a "seed" is that it delegates to `generate_well_conditioned_matrix(Random.GLOBAL_RNG, T, N)` in `ChainRulesTestUtils`, and I wasn't quite sure what the `GLOBAL_RNG` is about. I thought maybe the test-suite initializes the RNG so that each test run is reproducible (uses the same random numbers each time). Putting `B = randn(ComplexF64, 4, 4)` might break that behavior, but then I just realized that plenty of other places in the test also use `randn` directly.

So yeah, if `B = randn(ComplexF64, 4, 4)` is fine, then I can use either `U = exp(1im * (B + B'))` or (presumably) `U = exp(B - B')`

sethaxen: `Random.GLOBAL_RNG` is what is used by the `rand`/`randn` functions in `Random` if no `rng` is provided, so using `randn` here does the same thing. And yes, we also use `randn` with no specified `rng` through our test suite.

> So yeah, if `B = randn(ComplexF64, 4, 4)` is fine, then I can use either `U = exp(1im * (B + B'))` or (presumably) `U = exp(B - B')`

It's not very important, but the latter requires one less allocation, so I'd prefer it. The two expressions both target the same distribution on the unitary matrices.

goerz: Ok, so I switched `B` to a random matrix, but left `U` to be unitary. Let me know if that's ok.

test/rulesets/LinearAlgebra/dense.jl
```diff
@@ -119,6 +119,12 @@
                 test_rrule(f, B)
             end
         end
+        @testset "$f(unitary matrix)" begin
+            B = randn(ComplexF64, 4, 4)
+            U = exp(1im * (B + B'))
```
sethaxen, line 124: I still prefer this simpler expression:
```suggestion
            U = exp(B - B')
```

goerz:
> And, can you also increment the patch version?

Done

goerz:
Ok, I think this should be good to go

sethaxen (approved):

PR #604, opened 2022-03-28
Remove bad non-differentiable

### LINKED ISSUE #603: Silent Gradient Drop
[This](https://github.com/JuliaStats/Distributions.jl/blob/c9d6c28f415025bf489ac3bec2f8eec46b0eefbd/src/genericrand.jl#L48) fallback method for `rand` in `Distributions.jl` hits [this](https://github.com/JuliaDiff/ChainRules.jl/blob/f13e0a45d10bb13f48d6208e9c9d5b4a52b96732/src/rulesets/Random/random.jl#L25) rule, which is declared non-differentiable.

This results in a silent failure, where there ought to be an error if the given AD (in my case Zygote), is unable to handle in-place differentiation.

willtebbutt:
Resolves #603

### DIFF
--- a/src/rulesets/Random/random.jl
+++ b/src/rulesets/Random/random.jl
@@ -19,11 +19,6 @@ end
 @non_differentiable rand(::Type{<:Real}, ::Integer...)
 @non_differentiable rand(::Integer...)
 
-# There are many different 1-3 arg methods, but not varargs
-@non_differentiable rand!(::Any)
-@non_differentiable rand!(::Any, ::Any)
-@non_differentiable rand!(::Any, ::Any, ::Any)
-
 @non_differentiable randexp(::Any...)
 
 @non_differentiable randexp!(::AbstractArray)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.28.1"
+version = "1.28.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


src/rulesets/Random/random.jl
```diff
@@ -19,11 +19,6 @@ end
 @non_differentiable rand(::Type{<:Real}, ::Integer...)
 @non_differentiable rand(::Integer...)
 
-# There are many different 1-3 arg methods, but not varargs
-@non_differentiable rand!(::Any)
-@non_differentiable rand!(::Any, ::Any)
-@non_differentiable rand!(::Any, ::Any, ::Any)
-
```
devmotion, line 26: Maybe keep less general versions such as `rand!(::AbstractArray)` and `rand!(::AbstractRNG, ::AbstractArray)`, similar to `randn!`?

willtebbutt: I've added back in a couple of methods.

willtebbutt:
Looks like this currently causes some downstream issues -- will attempt to isolate the particular rules that we need to keep Zygote and Diffractor tests passing.

mcabbott:
Diffractor never passes.

Zygote failures are things like this: 
```
/home/runner/work/ChainRules.jl/ChainRules.jl/downstream/test/gradcheck.jl:1730
[313](https://github.com/JuliaDiff/ChainRules.jl/runs/5733081278?check_suite_focus=true#step:6:313)
  Test threw exception
[314](https://github.com/JuliaDiff/ChainRules.jl/runs/5733081278?check_suite_focus=true#step:6:314)
  Expression: gradient((x->begin
[315](https://github.com/JuliaDiff/ChainRules.jl/runs/5733081278?check_suite_focus=true#step:6:315)
                sum(rand(Random.GLOBAL_RNG, Float32, 1, 1))
[316](https://github.com/JuliaDiff/ChainRules.jl/runs/5733081278?check_suite_focus=true#step:6:316)
            end), 1) == (nothing,)
```
for which the immediate fix is that there should be rules something like
```
@non_differentiable rand(::Type{<:Number}, ::AbstractRNG, ::Tuple)
@non_differentiable rand(::Type{<:Number}, ::AbstractRNG, ::Integer...)
```

devmotion:
This seems suprising since the PR does not touch `rand`, I would assume removing rules for `rand!` should not affect something like `sum(rand(args...))`?

mcabbott:
Yes, the method is:
```
rand(r::AbstractRNG, ::Type{X}, dims::Dims) where {X} = rand!(r, Array{X}(undef, dims), X)
```
If we allow `rand!(rng, array)` then it's possible we should allow `rand!(rng, array, eltype)` too.

marius311:
Stumbled on this after a long dive (w/ help from Zack Li and Brian Chen on Slack) trying to understand this incorrectly dropped gradient:
```julia
gradient(σ -> sum(rand(Xoshiro(1), MvNormal(zeros(2), σ*I))), 1) # nothing
```
Conversely, with this PR, instead of dropping it, you (correctly) get an error since the `rand` code in Distributions.jl is mutating. If you load DistributionsAD, then it works. But yea would be good not to silently and incorrectly drop it, so +1 from me on this PR (at least w.r.t. to this, no comment on anything else breaking)

oxinabox (approved):

PR #605, opened 2022-03-29
CI: with: coverage: 'false'

mcabbott:
This may disable code coverage more tracking and speed up tests?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,6 +41,8 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+        with:
+          coverage: 'false'
 #      - uses: julia-actions/julia-processcoverage@v1
 #      - uses: codecov/codecov-action@v1
 #        with:


simeonschaub:
Ah, I think I actually wanted to reenable coverage once we had support for it in Julia interpreter, but never got around to it. I think it's probably useful enough to reenable now, but I'd also be curious what the actual overhead is if you have done some benchmarks.

mcabbott:
This sped up CI by 4x in https://github.com/mcabbott/Tullio.jl/pull/144 so I wondered what would happen here, but so far it doesn't look promising. I have not done any local benchmarking. 


My hope was that coverage tracking on nightly only might be enough, where (it seems, and I am told) this has less overhead.

simeonschaub:
Yeah, I'd assume that the relative overhead of logging code coverage would be a lot smaller when running in an interpreter, so I'm not surprised that there's not much of a difference.

simeonschaub:
Ah, wait! Coverage in JuliaInterpreter only works on 1.8+, so it's fine to set coverage to false in all previous versions. If you look in the nightly Job, it's actually reenabled there.

mcabbott:
Maybe I close this as it didn't help much.

PR #607, opened 2022-04-12
use ProjectTo in rules for constructors of triangular matrixes

oxinabox:
This seems correct?


mzgubic (commented):
Yeah, looks correct, but it's not clear to my why CI fails?

PR #610, opened 2022-04-23
Improve type stability of `cat_pullback`

ToucheSir:
Currently, https://github.com/JuliaDiff/ChainRules.jl/blob/8c34f19d3a8a8a224c9fbe20524d2a08c8b9bf81/src/rulesets/Base/array.jl#L356-L362 will generate a `Tuple{Union{UnitRange{Int64}, Colon}, ...}`. When the number of input dims >=3, this appears to fail some union-splitting threshold for `getindex` on certain array types. Here's a MWE with FillArrays, which Zygote uses somewhat extensively:
```julia
using Test, ChainRules, ChainRulesCore, FillArrays

function grad(x1, x2, sens, dims::Val)
  y, back = rrule(cat, x1, x2; dims)
  back(sens)
end

# Infers
@inferred grad(ones(1, 1, 1), ones(1, 1, 1), ones(1, 2, 1), Val(2))
@inferred grad(Ones(1, 1), Ones(1, 1), ones(1, 2), Val(2))

# Does not infer
@inferred grad(Ones(1, 1, 1), Ones(1, 1, 1), Ones(1, 2, 1), Val(2))
@inferred grad(ones(1, 1, 1), ones(1, 1, 1), Ones(1, 2, 1), Val(2))
```
This PR resolves the instability by always creating a `UnitRange`. However, I am not sure if this is sufficiently semantically close to `Colon` to pass muster, so RFC :) .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.28.2"
+version = "1.28.3"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -357,7 +357,7 @@ function rrule(::typeof(cat), Xs::Union{AbstractArray, Number}...; dims)
                 if d in cdims
                     d > ndimsX ? (prev[d]+1) : (prev[d]+1:prev[d]+sizeX[d])
                 else
-                    d > ndimsX ? 1 : (:)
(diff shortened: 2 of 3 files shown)


mcabbott:
This seems fine to me. Could you add something like this example as a test?

The rules for hcat and vcat have a similar branch, `d > ndimsX ? 1 : (:)`. But perhaps that never matters.

ToucheSir:
> The rules for hcat and vcat have a similar branch, `d > ndimsX ? 1 : (:)`. But perhaps that never matters.

It also doesn't matter for 1 and 2D inputs here, which is why I suspect union splitting is bailing out. Cthulhu shows 4 variants of `getindex(::Fill, ...)` generated for the 2D case, so I guess the threshold is somewhere between 2^2 and 2^3?

mcabbott (approved):

Issue #611, opened 2022-05-07
Wrong call on `cholesky` `rrule`


theogf:
Identified in https://github.com/JuliaStats/PDMats.jl/issues/159

When calling the `cholesky` `rrule`, there is an error in the code at https://github.com/JuliaDiff/ChainRules.jl/blob/c5dbe030af390599848830ff43a5dffc04be69e2/src/rulesets/LinearAlgebra/factorization.jl#L526

`Ū = ΔC.U` will be `NoTangent()` since the `Tangent` of  `Cholesky` object does not contain the field `U`.
I think it should be `Ū = ΔC.factors`

sethaxen:
> `Ū = ΔC.U` will be `NoTangent()` since the `Tangent` of `Cholesky` object does not contain the field `U`.

`Cholesky` does not have the field `U`, but it does have the property `U`. Then there's this `rrule` for `getproperty`:
https://github.com/JuliaDiff/ChainRules.jl/blob/c5dbe030af390599848830ff43a5dffc04be69e2/src/rulesets/LinearAlgebra/factorization.jl#L534-L553

Note that the current rules won't compose well if any downstream code accesses `factors`, but everything should work well if they access `U` or `L`. This should probably be improved though to look more like the rules for `lu`:

https://github.com/JuliaDiff/ChainRules.jl/blob/c5dbe030af390599848830ff43a5dffc04be69e2/src/rulesets/LinearAlgebra/factorization.jl#L80-L169

theogf:
So this assumes that ΔC here is a `Cholesky` object right?
https://github.com/JuliaDiff/ChainRules.jl/blob/c5dbe030af390599848830ff43a5dffc04be69e2/src/rulesets/LinearAlgebra/factorization.jl#L524

sethaxen:
> So this assumes that `ΔC` here is a `Cholesky` object right?

I don't see anywhere it assumes that. The preceding line has the type annotation `ΔC::Tangent`. The (co)tangent for `Cholesky` will not be a `Cholesky`.

theogf:
Right so `ΔC` is a `Tangent`! So `ΔC.U` will call `getproperty`  here
https://github.com/JuliaDiff/ChainRulesCore.jl/blob/2d75b4be102bb41ba3ac6df6dec8bb9617b20f0f/src/tangent_types/tangent.jl#L104
And since `hasfield(NamedTuple, :U)` is false, it will return `NoTangent()`...

[...]

PR #612, opened 2022-05-07
Right field call for cholesky

theogf:
Should solve #611


sethaxen:
This is not the right fix. See https://github.com/JuliaDiff/ChainRules.jl/issues/611#issuecomment-1120309720

PR #613, opened 2022-05-10
Add rules for `det` and `logdet` of `Cholesky`

devmotion:
This PR adds rules for `det(::Cholesky)` and `logdet(::Cholesky)` (currently defined in Zygote for real numbers: https://github.com/FluxML/Zygote.jl/blob/885a904ed958c74cdaa2af7a971a6b5a2da908a7/src/lib/array.jl#L744-L748), and restricts the existing rules of `det` to `StridedMatrix{<:Number}`, as already done in #245 for `logdet`.

With these changes it will become possible to compute the gradient of `det \circ PDMats.PDMat` with Zygote (see https://github.com/JuliaStats/PDMats.jl/issues/159). As an example (works even after removing https://github.com/FluxML/Zygote.jl/blob/885a904ed958c74cdaa2af7a971a6b5a2da908a7/src/lib/array.jl#L744-L748):
```julia
julia> using PDMats, Zygote, LinearAlgebra

julia> x = [1. 0.2; 0.2 1.];

julia> Zygote.gradient(logdet ∘ PDMat, x)
([1.0416666666666667 -0.41666666666666674; 0.0 1.0416666666666667],)

julia> Zygote.gradient(det ∘ PDMat, x)
([1.0 -0.4000000000000001; 0.0 1.0000000000000002],)

julia> Zygote.gradient(logdet ∘ PDMat, Symmetric(x))
([1.0416666666666667 -0.20833333333333337; -0.20833333333333337 1.0416666666666667],)

julia> Zygote.gradient(det ∘ PDMat, Symmetric(x))
([1.0 -0.20000000000000004; -0.20000000000000004 1.0000000000000002],)
```

cc: @theogf @simsurace


mcabbott:
> existing rules of `det` to `StridedMatrix{<:Number}`

This produces:
```julia
julia> gradient(det, [1 2; 3 4]')
ERROR: Can't differentiate foreigncall expression
```
Isn't this precisely the case for which `ChainRulesCore.@opt_out` was invented?

mcabbott:
> The following example works

If you choose the numbers just right. 

> Why should det be defined more generally if logdet is restricted to StridedMatrices

Because it was done before `@opt_out` etc. Ideally it would now be fixed.

devmotion:
> Ideally it would now be fixed.

Yeah, it should be fixed by defining det and logdet for LU factorizations and removing the ones for matrices completely. But this seemed like a task for a different PR.

oxinabox:
> Yeah, it should be fixed by defining det and logdet for LU factorizations and removing the ones for matrices completely. But this seemed like a task for a different PR.

We should indeed do this.
Is there a reason not to use `@opt_out` until that has been done?
(I am open to hearing it).

This is kinda a balancing act right now.
Do we break  StaticArrays and AdjointMatrix (which are very commonly used), or do we break PDMats which are commonly used in probabilistic and ML applications where gradients are often wanted.
Without further convincing I am leaning towards leaving things as they are, since it is more confusing when somethin that used to work stops working, than when something that was broken continues to be broken.
But I am open to being convinced

oxinabox (changes requested):
splitting the type restrictions into a seperate PR would let me merge the rest of this without delay.

devmotion:
Yes, I thought about these issues as well during the last few days and came basically to the same conclusion.

As discussed in the issues and comments linked above, I still think the existing rule for `det` is particularly bad - not only since it captures too many matrix types that it shouldn't (wrappers such as `AbstractPDMat` but also e.g. `Symmetric`) but also since it doesn't even work correctly for eg `Matrix{Float64}` since it errors for singular matrices.

However, I think break all currently working cases such as `StaticArray` and `Adjoint` is definitely undesirable. I wonder if the `logdet` restriction to `StridedMatrix` broke these as well or how similar problems were avoided.

In any case, I think the best and fastest way forward is to remove the `det` changes from this PR and open a separate PR for them.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -118,15 +118,15 @@ end
...
+function frule((_, Δx), ::typeof(det), x::StridedMatrix{<:Number})
     Ω = det(x)
     # TODO Performance optimization: probably there is an efficent
     # way to compute this trace without during the full compution within
     return Ω, Ω * tr(x \ Δx)
 end
 frule((_, Δx), ::typeof(det), x::Number) = (det(x), Δx)
 
-function rrule(::typeof(det), x::Union{Number, AbstractMatrix})
+function rrule(::typeof(det), x::Union{Number, StridedMatrix{<:Number}})
```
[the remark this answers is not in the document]
devmotion: I reverted these changes.

devmotion:
@oxinabox can you have another look? I removed the `det` changes, I think the PR should be easier to review and merge now.

devmotion:
Bump :slightly_smiling_face:

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -551,3 +551,24 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where {T <: Cholesky}
     end
     return getproperty(F, x), getproperty_cholesky_pullback
 end
+
+# `det` and `logdet` for `Cholesky`
+function rrule(::typeof(det), C::Cholesky)
+    y = det(C)
+    s = conj!((2 * y) ./ _diag_view(C.factors))
+    function det_Cholesky_pullback(ȳ)
+        ΔC = Tangent{typeof(C)}(; factors=Diagonal(ȳ .* s))
```
sethaxen, line 560: Would something like this be better? There's one fewer allocation, and we don't need to assume that `s` is mutable.
```suggestion
   diagF = _diag_view(C.factors)
    function det_Cholesky_pullback(ȳ)
        ΔC = Tangent{typeof(C)}(; factors=Diagonal(2(ȳ * conj(y)) ./ conj.(diagF)))
```

sethaxen: If the determinant is 0 (can happen if `check=false` is passed to `cholesky`), this will inject `NaN`s, even if the cotangent is 0. Since we try to treat cotangents as strong zeros, it would be nice to handle this case by ensuring that such `NaN`s end up as zeros.

devmotion: > There's one fewer allocation, and we don't need to assume that s is mutable.

Seems like an improvement to me, thanks! :slightly_smiling_face: 

There's a whitespace missing in the first line of your suggestion it seem:

```suggestion
    diagF = _diag_view(C.factors)
    function det_Cholesky_pullback(ȳ)
        ΔC = Tangent{typeof(C)}(; factors=Diagonal(2(ȳ * conj(y)) ./ conj.(diagF)))
```

devmotion: > If the determinant is 0

It would happen if `y = 0` (the determinant) but also if `ȳ = 0`. Should we care about the last case as well? Or is it correct to return `NaN` there?

devmotion: Ah, never mind, of course, at least one element of `diagF` is zero iff`y = 0`. I.e., we only have to care about `y = 0`.

devmotion: Maybe even a bit clearer (without having to know about precedence of operators):
```suggestion
    diagF = _diag_view(C.factors)
    function det_Cholesky_pullback(ȳ)
        ΔC = Tangent{typeof(C)}(; factors=Diagonal( (2 * (ȳ * conj(y))) ./ conj.(diagF)))
```

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -551,3 +551,24 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where {T <: Cholesky}
...
+        return NoTangent(), ΔC
+    end
+    return y, det_Cholesky_pullback
+end
+
+function rrule(::typeof(logdet), C::Cholesky)
+    y = logdet(C)
+    s = conj!((2 * one(eltype(C))) ./ _diag_view(C.factors))
+    function logdet_Cholesky_pullback(ȳ)
+        ΔC = Tangent{typeof(C)}(; factors=Diagonal(ȳ .* s))
```
devmotion, line 570: I guess something similar to above could be done here as well:
```suggestion
    diagF = _diag_view(C.factors)
    function logdet_Cholesky_pullback(ȳ)
        ΔC = Tangent{typeof(C)}(; factors=Diagonal( 2 .* ȳ ./ conj.(diagF) ))
```

devmotion: Here we could also end up with `NaN` if the determinant and hence at least one element of `diagF` is zero and `ȳ = 0`. I guess that's fine though?

src/rulesets/LinearAlgebra/factorization.jl
devmotion: I guess something like the following could work?
```julia
# compute `x / conj(y)`, handling `x = y = 0`
function _x_divide_conj_y(x, y)
    z = x / conj(y)
    # in our case `iszero(x)` implies `iszero(y)`
    return iszero(x) ? zero(z) : z
end
function rrule(::typeof(det), C::Cholesky)
    y = det(C)
    diagF = _diag_view(C.factors)
    function det_Cholesky_pullback(ȳ)
        ΔC = Tangent{typeof(C)}(; factors=Diagonal( _x_divide_conj_y(2 * (ȳ * conj(y)), diagF) ))
```

devmotion: @sethaxen I updated the PR and added tests for singular matrices.

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -551,3 +551,31 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where {T <: Cholesky}
     end
     return getproperty(F, x), getproperty_cholesky_pullback
 end
+
+# `det` and `logdet` for `Cholesky`
+function rrule(::typeof(det), C::Cholesky)
+    y = det(C)
+    diagF = _diag_view(C.factors)
+    function det_Cholesky_pullback(ȳ)
+        ΔF = Diagonal(_x_divide_conj_y.(2 * ȳ * conj(y), diagF))
```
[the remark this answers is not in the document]
devmotion: Usually I like to reuse existing code but this seems a bit strange to me. I think it reduces readability, and it seems it causes surprising and unnecessary operations such as `sqrt`?

src/rulesets/LinearAlgebra/factorization.jl
```diff
@@ -551,3 +551,31 @@ function rrule(::typeof(getproperty), F::T, x::Symbol) where {T <: Cholesky}
...
+    z = x / conj(y)
+    # in our case `iszero(x)` implies `iszero(y)`
+    return iszero(x) ? zero(z) : z
+end
+
+function rrule(::typeof(logdet), C::Cholesky)
+    y = logdet(C)
+    diagF = _diag_view(C.factors)
+    function logdet_Cholesky_pullback(ȳ)
+        ΔC = Tangent{typeof(C)}(; factors=Diagonal((2 * ȳ) ./ conj.(diagF)))
```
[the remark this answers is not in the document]
devmotion: I was wondering that as well (there are some - now hidden - comments above) but it felt like usually we don't handle such things in a special way if it can only be triggered by specific cotangents but is not an immediate consequence of the inputs. Or do we?

src/rulesets/LinearAlgebra/factorization.jl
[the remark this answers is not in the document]
devmotion: Probably due to these additional operations and more complicated computations the `prod`-based suggestion is also consistently a bit slower:

With current PR:
```julia
julia> using LinearAlgebra, ChainRules, BenchmarkTools

julia> function benchmark_det_Cholesky(n::Int)
           A = rand(n, n)
           C = cholesky(A' * A + I)
           @btime ChainRules.rrule(det, $C)
           return
       end
benchmark_det_Cholesky (generic function with 1 method)

julia> benchmark_det_Cholesky(5);
  29.937 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(10);
  31.029 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(50);
  60.686 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(100);
  103.206 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(1_000);
  876.981 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(10_000);
  66.534 μs (2 allocations: 80 bytes)
```

With the `prod` suggestion:
```julia
julia> using LinearAlgebra, ChainRules, BenchmarkTools

julia> function benchmark_det_Cholesky(n::Int)
           A = rand(n, n)
           C = cholesky(A' * A + I)
           @btime ChainRules.rrule(det, $C)
           return
       end
benchmark_det_Cholesky (generic function with 1 method)

julia> benchmark_det_Cholesky(5);
  36.872 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(10);
  37.621 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(50);
  63.905 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(100);
  105.336 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(1_000);
  880.792 ns (2 allocations: 80 bytes)

julia> benchmark_det_Cholesky(10_000);
  66.729 μs (2 allocations: 80 bytes)
```

Of course, for larger matrices this seemingly constant overhead matters less.

But my main concern is the readability here. The `prod` code is quite involved and IMO it's far from obvious what it's doing and that its use is correct here.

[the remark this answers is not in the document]
sethaxen: These are fair points, and this is a pretty minor comment, so I don't think this should be a blocker for this PR,

src/rulesets/LinearAlgebra/factorization.jl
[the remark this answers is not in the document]
sethaxen: I don't think we test this for all the rules we should, but a principle that we seem to be agreed on is that zero (co)tangents should be strong zeros (see e.g. https://github.com/JuliaDiff/ChainRulesCore.jl/pull/551#issuecomment-1129121382).

So in this case if `ȳ==0`, then the cotangent of `factors` should be a zero matrix. Otherwise you end up with cases like `zero(logdet(cholesky(A; check=false)))`, which pulls back a zero cotangent through this rule, injecting `NaN`'s into all downstream cotangents, even though the output is unrelated to the value of `A`.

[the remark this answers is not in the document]
devmotion: I see, I was not aware of this principle. Would maybe good to add it to the docs and possibly CRTestUtils :slightly_smiling_face: 

I'll update the PR accordingly.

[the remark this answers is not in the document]
devmotion: Should be fixed in https://github.com/JuliaDiff/ChainRules.jl/pull/613/commits/d831cd49fc8a7669e8e10930bf3214da61ba22dc

sethaxen (approved):
LGTM! Assuming the same tests pass that pass on main, I think this is ready to merge. Thanks, @devmotion!

PR #616, opened 2022-05-12
Eliminate redundant call to `norm`

danielwe:
Was staring at code and noticed this


mzgubic (approved):
Thanks for opening! Could you bump the minor version please?

danielwe:
done

Issue #85, opened 2019-08-16
Add rule for `mean(f, x; dims)`


nickrobinson251:
We have rules for `mean`, except for `mean(f, x; dims)` which is new as of Julia v1.3

mzgubic:
The implementation could look a bit like `sum(f, xs)`:

https://github.com/JuliaDiff/ChainRules.jl/blob/ce78d3d3e8aaf6303e1aa7085fdbdfc2d36d1b64/src/rulesets/Base/mapreduce.jl#L69-L94

mcabbott:
Ideally it would probably share code, have a function which for `mean` gets `scale=1/size(...)` or something.

oxinabox:
reopened as i had to revert the fix

PR #619, opened 2022-05-17
Revert "Rule for `mean(f,x)`"

oxinabox:
Reverts JuliaDiff/ChainRules.jl#615

because 
This broke Invenia's code base. Something went wrong with how this works with NamedDims.jl.
I am reverting this til it can be debugged

```
2022-05-17T18:24:11Z	MethodError: no method matching iterate(::Nothing)
2022-05-17T18:24:11Z	Stacktrace:
2022-05-17T18:24:11Z	  [1] indexed_iterate(I::Nothing, i::Int64)
2022-05-17T18:24:11Z	    @ Base ./tuple.jl:89
2022-05-17T18:24:11Z	  [2] rrule(config::Zygote.ZygoteRuleConfig{Zygote.Context}, ::typeof(Statistics.mean), f::NamedDims.NamedDimsArray{(:target, :node), Float32, 2, Matrix{Float32}}, x::StatsBase.AnalyticWeights{Float32, Float32, Vector{Float32}}; dims::Function)
2022-05-17T18:24:11Z	    @ ChainRules ~/.julia/packages/ChainRules/OJ3vT/src/rulesets/Statistics/statistics.jl:26
2022-05-17T18:24:11Z	  [3] rrule(config::Zygote.ZygoteRuleConfig{Zygote.Context}, ::typeof(Statistics.mean), f::NamedDims.NamedDimsArray{(:target, :node), Float32, 2, Matrix{Float32}}, x::StatsBase.AnalyticWeights{Float32, Float32, Vector{Float32}})
2022-05-17T18:24:11Z	    @ ChainRules ~/.julia/packages/ChainRules/OJ3vT/src/rulesets/Statistics/statistics.jl:26
2022-05-17T18:24:11Z	  [4] chain_rrule
2022-05-17T18:24:11Z	    @ ~/.julia/packages/Zygote/DkIUK/src/compiler/chainrules.jl:217 [inlined]
2022-05-17T18:24:11Z	  [5] macro expansion
2022-05-17T18:24:11Z	    @ ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0 [inlined]
```

Reopens https://github.com/JuliaDiff/ChainRules.jl/issues/85, reopens https://github.com/FluxML/Zygote.jl/issues/1128

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.32.0"
+version = "1.32.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/Statistics/statistics.jl
+++ b/src/rulesets/Statistics/statistics.jl
@@ -6,29 +6,17 @@ _denom(x, dims::Colon) = length(x)
 _denom(x, dims::Integer) = size(x, dims)
 _denom(x, dims) = mapreduce(i->size(x, i), Base.mul_prod, unique(dims), init=1)
 
-function rrule(::typeof(mean), x::AbstractArray{<:Union{Real,Complex,AbstractArray}}; dims=:)
-    y_sum, sum_pullback = rrule(sum, x; dims)
+# TODO: We have `mean(f, x; dims)` as of 1.3.0-DEV.36
+# https://github.com/JuliaDiff/ChainRules.jl/issues/85
+function rrule(::typeof(mean), x::AbstractArray{<:Real}; dims=:)
+    y_sum, sum_pullback = rrule(sum, x; dims=dims)
     n = _denom(x, dims)
     function mean_pullback(ȳ)
-        _, ∂x = sum_pullback(unthunk(ȳ) / n)
+        _, ∂sum_x = sum_pullback(ȳ)
+        ∂x = unthunk(∂sum_x) / n
         return (NoTangent(), ∂x)
     end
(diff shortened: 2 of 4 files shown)


rofinn:
Assuming all other test besides Diffractor.jl I'll merge this.

NOTE: The Diffractor.jl test is being disabled in #620 anyway.

mcabbott:
This is like #522, there's a StatsBase signature with two arrays. Slightly pirate, since std. lib. does allow that, although it doesn't work:
```
julia> using Statistics

julia> mean([1,2], [3,4])
ERROR: MethodError: objects of type Vector{Int64} are not callable
```

rofinn:
Hmm, I think merging the revert seems safer to me for now. I don't think the nightly CI errors should block this, but maybe @mcabbott thinks this shouldn't go in if the fix is sufficiently simple?

oxinabox:
Reverting the revert is also fast.

In general I am strongly in favor of reverting things that cause problems sooner rather than any delay as people start adding work arounds for them and that makes things more disruptive.


Let's do this and then we do not have to rush to get the final solution in.

mcabbott:
Also, https://github.com/FluxML/Zygote.jl/issues/836 is probably the original issue for this.

mcabbott:
Bump, after 6 months? 

See this is the problem with reverting. While everyone was interested, we could have fixed the weird (not-quite-kosher) StatsBase signature in like a day, back in May. But instead, nobody has quite got around to it in half a year.

oxinabox:
I mean yes, but everyone was also too busy. As you can guess from my current burst of activity I am finally having time to work on this stuff again.
Still yes lets look at it, create a reproducer, and fix this.
I will do so presently

PR #620, opened 2022-05-17
Comment out Reverse dependency test of Diffractor

oxinabox:
Diffractor has been broken on nightly for months now.
Til it is fixed it just adds noise to our CI


oxinabox:
This doesn't actually need a version bump.

PR #621, opened 2022-05-21
make `Base.binomial()` non-diff

Moelf:


oxinabox:
I wonder if just defining `Core.Intrinsics.trunc_int` as nondifferentable would be enough

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.33.0"
+version = "1.33.1"
```
oxinabox, line 3:
```suggestion
version = "1.34.0"
```

This is a new feature, not a bug fix

ChrisRackauckas:
> I wonder if just defining Core.Intrinsics.trunc_int as nondifferentable would be enough

Maybe, but we could also just fix it.

Issue #625, opened 2022-05-25
Matrix multiplication rule for StridedArray fails with mixed `eltype`s


Jutho:
The custom rule
https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/arraymath.jl#L47
with `InplaceableThunk` fails when `A` and `B` have different `eltype`s, in particular when `A` is real and `B` is complex (or vice versa).

The minimal replacement is
```julia
function rrule(
    ::typeof(*),
    A::StridedMatrix{<:CommutativeMulNumber},
    B::StridedVecOrMat{<:CommutativeMulNumber},
)
    project_A = ProjectTo(A)
    project_B = ProjectTo(B)
    function times_pullback(ȳ)
        Ȳ = unthunk(ȳ)
        dA = InplaceableThunk(
            X̄ -> mul!(X̄, Ȳ, B', true, true),
            @thunk(project_A(Ȳ * B')),
        )
        dB = InplaceableThunk(
            X̄ -> mul!(X̄, A', Ȳ, true, true),
            @thunk(project_B(A' * Ȳ)),
        )
        return NoTangent(), dA, dB
    end
    return A * B, times_pullback
end
```
as for the generic rule at [line 27](https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/arraymath.jl#L27). However, this will only fix the non-additive part of the `InplaceableThunk`. The additive part will still fail if `X̄` is real, and at least one of `Ȳ` or `B` is complex. A more complete correction could look like

```julia
    _realview(A::AbstractArray{Complex{T}}) where {T} = view(reinterpret(reshape, T, A), 1, axes(A)...)
    _imagview(A::AbstractArray{Complex{T}}) where {T} = view(reinterpret(reshape, T, A), 2, axes(A)...)

    function _muladdproject!(C::StridedMatrix{<:Complex}, A::StridedVecOrMat, B::StridedVecOrMat)
        mul!(C, A, B, one(eltype(C)), one(eltype(C)))
    end
    function _muladdproject!(C::StridedMatrix{<:Real}, A::StridedVecOrMat{<:Real}, B::StridedVecOrMat{<:Real})
        mul!(C, A, B, one(eltype(C)), one(eltype(C)))
    end
    function _muladdproject!(C::StridedMatrix{<:Real}, A::StridedVecOrMat{<:Complex}, B::StridedVecOrMat{<:Real})
[...]

PR #626, opened 2022-05-30
restrict strided array multiplication `rrule`

Jutho:
Fix the specialized `rrule` for `StridedArray` multiplication to equal `eltype`. Fixes issue [#625](https://github.com/JuliaDiff/ChainRules.jl/issues/625).


oxinabox (approved):

oxinabox:
LGTM, as this will fall back to the AbstractArray case for eltypes not matching.
So it will still all work

PR #631, opened 2022-06-17
Add rules for division by Cholesky

### LINKED ISSUE #611: Wrong call on `cholesky` `rrule`

sethaxen:
`\(::Cholesky, ::AbstractVecOrMat)` and `/(::AbstractMatrix, ::Cholesky)` internally call `ldiv!` and `rdiv!`, respectively. This PR adds `rrule`s for these functions to cover this up. This also allows the correspond adjoint for `\` to be removed from Zygote.

Fixes #611

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.40.0"
+version = "1.41.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.35.3"
+version = "1.36.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -578,3 +578,38 @@ function _x_divide_conj_y(x, y)
     z = x / conj(y)
     return iszero(x) ? zero(z) : z
 end
+
+# these rules exists because the primals mutates using `ldiv!` and `rdiv!`
+function rrule(::typeof(\), A::Cholesky, B::AbstractVecOrMat{<:Union{Real,Complex}})
+    U, getproperty_back = rrule(getproperty, A, :U)
+    Z = U' \ B
+    Y = U \ Z
+    project_B = ProjectTo(B)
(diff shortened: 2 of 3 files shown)


devmotion:
I'm not sure if this should be added here. Are there other examples in ChainRules where a rule is implemented only because the primal is mutating? At first glance this seems like a Zygote specific issue that should be addressed there. Or more generally a problem of mutation-incompatible ADs, so maybe one should only define it for `RuleConfig{>:Union{ReverseMode,NoMutation}}` (or something similar).

It also seems this problem might affect factorizations more generally? https://github.com/JuliaLang/julia/blob/ae89f8b00a6e785832e4f47de73599e032aa9674/stdlib/LinearAlgebra/src/factorization.jl#L103

sethaxen:
> Are there other examples in ChainRules where a rule is implemented only because the primal is mutating?

Yes, there are number of rules defined due to mutation; most are not documented as such, but see https://github.com/JuliaDiff/ChainRules.jl/search?q=mutates. The rules don't belong in Zygote, since other ADs like Diffractor will also need them. In the past, we have added rules that cover up mutation so that they can be removed from Zygote.

> Or more generally a problem of mutation-incompatible ADs, so maybe one should only define it for `RuleConfig{>:Union{ReverseMode,NoMutation}}` (or something similar).

Yeah that's an interesting idea. I wonder why we haven't done that in the past. @oxinabox @mcabbott do you know?

> It also seems this problem might affect factorizations more generally? https://github.com/JuliaLang/julia/blob/ae89f8b00a6e785832e4f47de73599e032aa9674/stdlib/LinearAlgebra/src/factorization.jl#L103

Yes, but unfortunately the solution is specific to each factorization. We handle this one here since there's a corresponding adjoint in Zygote we'd like to remove.

ToucheSir:
> > Or more generally a problem of mutation-incompatible ADs, so maybe one should only define it for `RuleConfig{>:Union{ReverseMode,NoMutation}}` (or something similar).
> 
> Yeah that's an interesting idea. I wonder why we haven't done that in the past. @oxinabox @mcabbott do you know?

I suppose one hang-up is that there is no `HasMutation`/`NoMutation` type in CRC :P. Were there one, we'd update `ZygoteRuleConfig` to use it.

oxinabox:
>      I suppose one hang-up is that there is no HasMutation/NoMutation type in CRC :P. Were there one, we'd update ZygoteRuleConfig to use it.

I  am in no hurry to add it.
Later we can, but first i want to see a real AD that has CR support and handles mutatation well.
Yota also doesn't.
And nor does Nabla.

and ReverseDiff is opt in per rule anyway,

PR #633, opened 2022-06-17
Use new cholesky pivot syntax in v1.8

### LINKED ISSUE #477: upcoming changes to `cholesky` in base

sethaxen:
Fixes #477 by adopting a similar approach as #434, except that it uses a type union to still support the old syntax. This will need to be removed for future Julia versions when the deprecated syntax is removed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.35.3"
+version = "1.35.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


devmotion (approved):
This looks good to me. Mainly a version bump seems to be missing?

I think this would fix some downstream issues in Zygote on newer Julia versions. These issues popped up also in downstream packages such as https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/issues/135. I assume this PR might fix them since one issue seems to be that currently the rules for `cholesky` in ChainRules are not picked up (see e.g. https://github.com/FluxML/Zygote.jl/runs/7124557589?check_suite_focus=true#step:6:1540).

oschulz:
Closes #647

oschulz:
Should also fix current Zygote failure on `cholesky(::PDiagMat)` in Julia v1.8.

PR #637, opened 2022-06-23
add remaining `@nograd` from Zygote

### LINKED ISSUE #635: Implement the existing Zygote `@nograd` as `@non_differentiable`
Base:
- [x] ones
- [x] zeros
- [ ] Base.OneTo
- [ ] Colon()
- [x] one
- [x] zero
- [ ] sizehint!
- [x] count
- [ ] Base.Channel
- [ ] Base.typeintersect
- [ ] floor
- [ ] ceil
- [ ] trunc
- [ ] round
- [x] div
- [ ] eps
- [ ] Base.eval
- [x] Symbol

Core
- [x] Core.TypeVar
- [x] Core.UnionAll

Distributed
- [ ] subtypes(Distributed.AbstractWorkerPool)

mzgubic:
Closes #635 

Also removes `floor`, `ceil`, and `round` non-diff from [previous PR](#636). I overlooked that we already have

```julia
# rouding related,
# we use `zero` rather than `ZeroTangent()` for scalar, and avoids issues with map etc
@scalar_rule round(x) zero(x)
@scalar_rule floor(x) zero(x)
@scalar_rule ceil(x) zero(x)
```
which looks like there is a reason we don't use `@non_differentiable`, though it isn't clear to me why.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,11 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.36"
+version = "1.37.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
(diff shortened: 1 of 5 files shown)


mcabbott (approved):

mcabbott:
While not strictly related, perhaps this PR could add `LinearAlgebra.isposdef` to close this issue:

https://github.com/FluxML/Zygote.jl/issues/1240#issuecomment-1151202934

oxinabox (approved):
LGTM.
Adding an extra dependency here on a stdlib is no problem.
(adding one for CRC.jl would be a problem)

PR #638, opened 2022-06-25
Trivial case for `cumsum`

mcabbott:
Needed for https://github.com/FluxML/Zygote.jl/pull/1253


mzgubic (approved):
Thanks! LGTM, just needs a version bump.

Issue #640, opened 2022-06-28
Support L, U and factors in Cholesky


rofinn:
In relation to other comments [here](https://github.com/invenia/Nabla.jl/issues/216) and [here](https://github.com/invenia/Nabla.jl/pull/217) I think we should read support for `getproperty(::Tangent{<:Cholesky}, ...)` with `:U` or `:L` along with `:factors`. This would make release 1.35.3 slightly less breaking.

https://github.com/invenia/Nabla.jl/pull/217/files#diff-a51816a450e57348c1c90d807b0a793092bb7e368d759102888820eb2e0ee025R55

PR #641, opened 2022-06-29
Add getproperty for Tangent{<:Cholesky}

rofinn:
Initial code for closing #640. I'm open to changing what it checks, but this seems like any easy way to minimize breaking changes for now.

### DIFF
--- a/src/rulesets/LinearAlgebra/factorization.jl
+++ b/src/rulesets/LinearAlgebra/factorization.jl
@@ -578,3 +578,11 @@ function _x_divide_conj_y(x, y)
     z = x / conj(y)
     return iszero(x) ? zero(z) : z
 end
+
+# Support both factors and U / L properties on cholesky tangent types
+# TOODO: Maybe we should check the type of U or L first?
+function Base.getproperty(tangent::Tangent{P, T}, sym::Symbol) where {P <: Cholesky, T <: NamedTuple}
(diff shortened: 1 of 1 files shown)


devmotion:
My intuition is that we would want `getproperty(x::Tangent{<:Cholesky}, :U) == getproperty(x::Tangent{<:Cholesky}, :L)'`. I think that's not satisfied by the PR?

I guess it might also be useful to define `getproperty(x::Tangent{<:Cholesky}, :UL)` (at least for completeness and to avoid `ZeroTangent`s in that case as well).

sethaxen:
One of the challenges with this approach is that we can't capture the logic present in this rule:
https://github.com/JuliaDiff/ChainRules.jl/blob/6ff4c319f8fd25f27636d28144d78c92f81d8753/src/rulesets/LinearAlgebra/factorization.jl#L531-L543

Because we don't have access to the `Cholesky` object's `uplo` field, we don't know the relationship between `factors`, `U`, and `L`. e.g. if `uplo=='L'`, then a user requesting `ΔC.L` should be given `LowerTriangular(ΔC.factors)`, and a user requesting `ΔC.U` should be given `LowerTriangular(ΔC.factors)'`, but the case is completely different if `uplo=='U'`. Currently this rule will only work when `Symbol(C.uplo) == sym`. I'm not certain if/how this can be resolved, short of modifying `Tangent` to include metadata from the primal object.

PR #658, opened 2022-08-08
Don't use UnionAlls in Varargs

andreasnoack:
This has been deprecated. I think it would be helpful to users to test with `--depwarn=error` but I'm not sure how to enable that when using `julia-runtest`. @DilumAluthge Do you know if that is possible?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.43.0"
+version = "1.43.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


devmotion:
> I think it would be helpful to users to test with `--depwarn=error` but I'm not sure how to enable that when using `julia-runtest`.

Apparently one can set `depwarn` to `error`: https://github.com/julia-actions/julia-runtest/blob/7ea3b3e0bd2f7de6d8104e238f69a715c988a5d1/action.yml#L16-L18 I.e., it seems one would have to use something like
```
uses: julia-actions/julia-runtest@v1
with:
  depwarn: error
```

andreasnoack:
Thanks. Let's try to see if it works.

oxinabox:
I don't think we want `depwarn=error`.
It's deprecated.
It's not like it is going to be removed before Julia 2.0, right?

Further we do use some packages here, and those might deprecate things while we are still using them.
Sometimes using a deprecated thing is the best way to maximise compatibility.
Not in this case, but in general.
So I approve the core of this PR, but not the changes to our CI setup.

andreasnoack:
I've dropped the CI commit.

When deprecations were printed by default the overhead of taking the code path was often so high that it didn't matter that it technically wasn't breaking since the code ran too slowly to be useful. Apparently there is still a bit of overhead from taking these paths so it might be useful to detect them. We'll continue to do that downstream.

mcabbott (approved):

PR #668, opened 2022-08-22
resolve `rrule` ambiguities

### LINKED ISSUE #654: `repeat` rrule ambiguity for Bool arrays
```julia
julia> rrule(repeat, falses(1), 1)
ERROR: MethodError: rrule(::typeof(repeat), ::BitVector, ::Int64) is ambiguous. Candidates:
  rrule(::typeof(repeat), xs::AbstractArray, counts::Integer...) in ChainRules at /home/brianc/.julia/packages/ChainRules/o1vND/src/rulesets/Base/array.jl:191
  rrule(::typeof(repeat), var"543"::AbstractArray{Bool}, var"544"...; repeat_pullback) in ChainRules at /home/brianc/.julia/packages/ChainRules/o1vND/src/rulesets/Base/nondiff.jl:65
Possible fix, define
  rrule(::typeof(repeat), ::AbstractArray{Bool}, ::Vararg{Integer})
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
```
Found while working on https://github.com/FluxML/Zygote.jl/issues/1234. Is there a way to constrain the vararg type in https://github.com/JuliaDiff/ChainRules.jl/blob/1770bb29ca42d4e07643284e0a4917ad6ea35b57/src/rulesets/Base/nondiff.jl#L65 to remove this ambiguity?

mzgubic:
Closes #654 

And adds a test to prevent regressions in `rrule`.

There are still plenty ambiguities for `frule` because their first arg when using `@non_diff` is sometimes `::Any`, which produces an ambiguity with the generic `RuleConfig` dispatch.


src/rulesets/Base/nondiff.jl
```diff
@@ -54,15 +54,14 @@
 @non_differentiable invperm(::AbstractArray{Bool})
 @non_differentiable invpermute!(::Any, ::AbstractArray{Bool})
 @non_differentiable mapslices(::Any, ::AbstractArray{Bool})
-@non_differentiable Matrix(::AbstractArray{Bool})
```
mzgubic, line 57: this is actually taken care of in https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Base/array.jl#L15-L19

and the project to makes sure it is always NoTangent() for arrays of booleans, and it compiles away (at least the time is the same as with the dedicated rule)

mcabbott (approved):

PR #670, opened 2022-08-25
Explicitly `unthunk` in a few rules

mcabbott:
The `rrule` for `+` would silently unthunk N times, by calling `reshape`. (Maybe `test_rrule` should have a thunk which counts to test for such mistakes?)

And most broadcasting would fail on thunks, by calling something like `ndims(@thunk [1,2])`. (Not sure why this didn't cause failures.)


mzgubic (approved):
LGTM, just need a version bump

PR #674, opened 2022-09-02
broadcast type preservation

maartenvd:
appears to solve the diffractor-zygote case in

https://github.com/FluxML/Zygote.jl/issues/1301

@mcabbott


mcabbott:
Do you know what the test errors are? Weird it doesn't print them in CI.

Xref comments on testing this a https://github.com/FluxML/Zygote.jl/pull/1302

Issue #678, opened 2022-10-08
cat with Val tuple dims fails


cossio:
```
X = randn(4,5,3)
gs = gradient(X) do X
    sum(cat(X, X; dims=Val((1,2))))
end
```

gives an error:

```
ERROR: MethodError: no method matching Int64(::Tuple{Int64, Int64})
Closest candidates are:
  (::Type{T})(::AbstractChar) where T<:Union{Int32, Int64} at char.jl:51
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  (::Type{T})(::BigInt) where T<:Union{Int128, Int16, Int32, Int64, Int8} at gmp.jl:359
  ...
Stacktrace:
  [1] rrule(::typeof(cat), ::Array{Float64, 3}, ::Vararg{Array{Float64, 3}}; dims::Val{(1, 2)})
    @ ChainRules ~/.julia/packages/ChainRules/KVV0e/src/rulesets/Base/array.jl:350
  [2] chain_rrule_kw
    @ ~/.julia/packages/Zygote/dABKa/src/compiler/chainrules.jl:230 [inlined]
  [3] macro expansion
    @ ~/.julia/packages/Zygote/dABKa/src/compiler/interface2.jl:0 [inlined]
  [4] _pullback(::Zygote.Context{false}, ::Base.var"#cat##kw", ::NamedTuple{(:dims,), Tuple{Val{(1, 2)}}}, ::typeof(cat), ::Array{Float64, 3}, ::Array{Float64, 3})
    @ Zygote ~/.julia/packages/Zygote/dABKa/src/compiler/interface2.jl:9
  [5] _pullback
    @ ./REPL[34]:2 [inlined]
  [6] _pullback(ctx::Zygote.Context{false}, f::var"#37#38", args::Array{Float64, 3})
    @ Zygote ~/.julia/packages/Zygote/dABKa/src/compiler/interface2.jl:0
  [7] pullback(f::Function, cx::Zygote.Context{false}, args::Array{Float64, 3})
    @ Zygote ~/.julia/packages/Zygote/dABKa/src/compiler/interface.jl:44
  [8] pullback
    @ ~/.julia/packages/Zygote/dABKa/src/compiler/interface.jl:42 [inlined]
  [9] gradient(f::Function, args::Array{Float64, 3})
    @ Zygote ~/.julia/packages/Zygote/dABKa/src/compiler/interface.jl:96
[...]

PR #679, opened 2022-10-08
fix cat rrule

cossio:
fix https://github.com/JuliaDiff/ChainRules.jl/issues/678

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.44.6"
+version = "1.44.7"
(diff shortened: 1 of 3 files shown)


oxinabox:
Please bump the version.
Which will also rerun CI incase this is some weird transient thing

mzgubic:
We are seeing the segfaults sometimes at nightly, so that's unrelated. The 1.6 failures look real though, and don't appear on nightly. I haven't looked at the changes yet to have ideas on how it might be related 🤷

mzgubic (approved):
Nice, thanks for persisting with the CI failures and finding the solution. The Nightly failures we see on main branch as well and they are unrelated.

PR #680, opened 2022-10-12
Remove frule for getindex(::Tuple, i)

Keno:
Having this chain rule is sub-optimal, because it prevents early-SROA in Diffractor-like systems that would like to perform some optimizations before applying AD (but can't do any optimization on functions that have custom rules). By letting it go down to the `getfield`, regular SROA can apply. Any AD system should handle `getfield` anyway, so I don't think there's a strong reason to have this.

Similar reasoning applies to the reverse rules also, but they aren't currently actively causing me problems, so this PR only removes the frule, since I don't think many other packages are using them. We can revisit the rrules later.

### DIFF
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -1,3 +1,8 @@
+# Int rather than Int64/Integer is intentional
+function frule((_, ẋ), ::typeof(getfield), x::Tuple, i::Int)
+    return x.i, ẋ.i
+end
+
 "for a given tuple type, returns a Val{N} where N is the length of the tuple"
 _tuple_N(::Type{<:Tuple{Vararg{Any, N}}}) where {N} = Val(N)
 
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -179,29 +179,6 @@ function ∇getindex!(dx::AbstractGPUArray, dy, inds...)
 end
 
 #####
-##### first, tail
-#####
-
-function frule((_, ẋ), ::typeof(first), x::Tuple)
-    return first(x), first(ẋ)
-end
-
-function rrule(::typeof(first), x::T) where {T<:Tuple}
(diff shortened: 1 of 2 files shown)


Keno:
Probably makes sense to actually add an frule for `getfield` in the same breath.

mcabbott (approved):
I will take your word for it, but right now this branch leads to the following failure:
```
julia> Diffractor.PrimeDerivativeFwd(x -> (x,2x,3x)[2])(1.0)
2.0

julia> Diffractor.PrimeDerivativeFwd(x -> sum((x,2x,3x)[1:2]))(1.0)
ERROR: type CompositeBundle has no field partials
Stacktrace:
  [1] getproperty
    @ ./Base.jl:37 [inlined]
  [2] (::Diffractor.∂☆{1})(f::Diffractor.UniformBundle{1, typeof(getfield), ZeroTangent}, x::Diffractor.CompositeBundle{1, Tuple{Float64, Float64, Float64}, Tuple{Diffractor.TangentBundle{1, Float64, Tuple{Float64}}, Diffractor.TangentBundle{1, Float64, Tuple{Float64}}, Diffractor.TangentBundle{1, Float64, Tuple{Float64}}}}, s::Diffractor.TangentBundle{1, Int64, Tuple{NoTangent}}, inbounds::Diffractor.UniformBundle{1, Bool, ZeroTangent})
    @ Diffractor ~/.julia/dev/Diffractor/src/stage1/forward.jl:165
```

Keno:
I need to update Diffractor. We can hold this branch until I can push that.

oxinabox:
Can we rebase and merge this?

Keno:
Rebased and also gave the same treatment to `first`,`tail`, which @staticfloat ran into having the same issue.

oxinabox:
CI is failing,
not just because we haven't merged #718  
even the 1.6 CI is failing.
This is because the tests haven't been deleted.
this PR should also delete the matching tests.

> Probably makes sense to actually add an frule for getfield in the same breath.


This hasn't been done yet?

oscardssmith:
I've deleted the unnecessary tests and added the getfield frule. Let's see if this works.

oxinabox:
All CI failures were preexisting
This can be merged now

PR #683, opened 2022-10-24
handle permutedims rrule with ZeroTangent

chengchingwen:


src/rulesets/Base/array.jl
```diff
@@ -143,18 +143,21 @@ end
 function rrule(::typeof(permutedims), x::AbstractVector)
     project = ProjectTo(x)
     permutedims_pullback_1(dy) = (NoTangent(), project(permutedims(unthunk(dy))))
+    permutedims_pullback_1(::ZeroTangent) = (NoTangent(), ZeroTangent())
```
mcabbott, line 146: IDK why there are several Zeros, but perhaps if doing this it may as well accept all of them:
```suggestion
    permutedims_pullback_1(z::AbstractZero) = (NoTangent(), z)
```

chengchingwen: IDK either, but I follow how CR.jl previously do.
https://github.com/JuliaDiff/ChainRules.jl/blob/a9a84ba6cb8aa9ce079af9401600e7c96a8aff3a/src/rulesets/LinearAlgebra/norm.jl#L59

mcabbott: Instances of the wider pattern, which I didn't write, include:

https://github.com/JuliaDiff/ChainRules.jl/blob/b1daa7aa6849ed66d0004de73137207e6a2d007b/src/rulesets/SparseArrays/sparsematrix.jl#L33

https://github.com/JuliaDiff/ChainRules.jl/blob/b1daa7aa6849ed66d0004de73137207e6a2d007b/src/rulesets/SparseArrays/sparsematrix.jl#L33

mcabbott (approved):

src/rulesets/Base/array.jl
chengchingwen: Oh, then should I update this PR, or just leave as it is?

PR #691, opened 2023-02-08
Added differentiation rule for `promote_type`

kaandocal:
This is intended to fix https://github.com/FluxML/Zygote.jl/issues/1372


ToucheSir:
Looks like all the failures are actually broken tests which now work. GH Actions log shows they're unrelated to this PR. @oxinabox given that, would you prefer the tests be marked unbroken here or in a separate PR?

oxinabox:
Lets do it in a separate PR.
I do wonder if should also be promoting derivative type here also but lets wait til someone requests that

PR #694, opened 2023-02-15
Optimize power rule

YingboMa:


mcabbott:
The hard part was getting all the weird edge cases to work. If introducing extra branches like `if isinteger(p)`, you may need to expand the list of test cases to check both halves.

oxinabox:
@YingboMa  are you going to add the missing test so we can merge this?

Issue #46, opened 2019-06-06
FillArrays


willtebbutt:
There are currently instances when we want to use arrays of zeros in which size information is retained - this is not something supported by the `Zero` differential, but which is supported by `FillArrays.jl`'s `Zeros` type. I would be nice to use this here, but are there perhaps issues with it not defining `setindex`?

Related to #15

willtebbutt:
FWIW, I'm very keen on having first-class AD support for `FillArrays` as I use them extensively in `Stheno.jl`, and `Zygote.jl` now makes use of them internally rather than rolling its own fill type.

oxinabox:
`tr` is an example that should be `FIllArrays`.
Zygote does so.

PR #717, opened 2023-05-18
Bring over OneElement for scalar getindex

oxinabox:
I think this is maybe is a precondition for https://github.com/FluxML/Zygote.jl/pull/1328
so that we know that deleting the Zygote rules will not cause performance regressions.

Still WIP as i chase down failing tests.

### DIFF
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -81,15 +81,36 @@ For the `rrule` of `y = x[inds...]`, this function is roughly
 `setindex(zero(x), dy, inds...)`, returning the array `dx`.
 Differentiable. Includes `ProjectTo(x)(dx)`.
 """
-function ∇getindex(x::AbstractArray, dy, inds...)
+function ∇getindex(x::AbstractArray{T,N}, dy, inds...) where {T,N}
     # `to_indices` removes any logical indexing, colons, CartesianIndex etc,
     # leaving just Int / AbstractVector of Int
     plain_inds = Base.to_indices(x, inds)
-    dx = _setindex_zero(x, dy, plain_inds...)
-    ∇getindex!(dx, dy, plain_inds...)
-    return ProjectTo(x)(dx)  # since we have x, may as well do this inside, not in rules
+    if plain_inds isa NTuple{N, Int} && T<:Number
+        # scalar indexing
+        return OneElement(dy, plain_inds, axes(x))
+    else  # some from slicing (potentially noncontigous)
+        dx = _setindex_zero(x, dy, plain_inds...)
+        ∇getindex!(dx, dy, plain_inds...)
+        return ProjectTo(x)(dx)  # since we have x, may as well do this inside, not in rules
+    end
 end
 ∇getindex(x::AbstractArray, z::AbstractZero, inds...) = z
+
+"""
+    OneElement(val, ind, axes) <: AbstractArray
+
+Extremely simple `struct` used for the gradient of scalar `getindex`.
+"""
+struct OneElement{T,N,I,A} <: AbstractArray{T,N}
+  val::T
+  ind::I
+  axes::A
+  OneElement(val::T, ind::I, axes::A) where {T<:Number, I<:NTuple{N,Int}, A<:NTuple{N,AbstractUnitRange}} where {N} = new{T,N,I,A}(val, ind, axes)
+end
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.ind, A.val, zero(T))
+# TODO: should we teach ProjectTo that OneElement is more structurally sparse than anything it intersects nonstructural zeros with?
 
 """
     _setindex_zero(x, dy, inds...)
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -85,14 +85,14 @@ function ∇getindex(x::AbstractArray{T,N}, dy, inds...) where {T,N}
     # `to_indices` removes any logical indexing, colons, CartesianIndex etc,
     # leaving just Int / AbstractVector of Int
     plain_inds = Base.to_indices(x, inds)
-    if plain_inds isa NTuple{N, Int} && T<:Number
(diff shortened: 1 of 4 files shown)


src/rulesets/Base/indexing.jl
```diff
@@ -81,16 +81,37 @@ For the `rrule` of `y = x[inds...]`, this function is roughly
...
+struct OneElement{T,N,I,A} <: AbstractArray{T,N}
+  val::T
+  ind::I
+  axes::A
+  OneElement(val::T, ind::I, axes::A) where {T<:Number, I<:NTuple{N,Int}, A<:NTuple{N,AbstractUnitRange}} where {N} = new{T,N,I,A}(val, ind, axes)
+end
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.ind, A.val, zero(T))
+# TODO: should we teach ProjectTo that OneElement is more structurally sparse than anything it intersects nonstructural zeros with?
```
oxinabox, line 113: Thoughts?

test/rulesets/Base/indexing.jl
```diff
@@ -89,7 +89,7 @@
         
         test_rrule(getindex, Symmetric(rand(3, 3)), 2, 2)
         sgrad = rrule(getindex, Symmetric(rand(3, 3)), 2, 3)[2](1.0)[2]
-        @test unthunk(sgrad) ≈ [0 0 0; 0 0 1/2; 0 1/2 0]
+        @test unthunk(sgrad) ≈ [0 0 0; 0 0 1/2; 0 1/2 0]  # We are actually getting this wrong now
```
oxinabox, line 92: It combines them to just be a `OneElement`
I think this might be a decent argument against `OneElement` at least in cases other than known dense arrays?

mcabbott: Why does this fail? I'd expect ProjectTo would make an Array here.

If we do this, quite a few `(::ProjectTo)(::OneElement)` could specialise.

src/rulesets/Base/indexing.jl
mcabbott: Not sure I follow the sentence. 

Is OneElement structurally sparse? The question can only arise for 2nd derivatives, which leads me to thin kit probably is.

oxinabox: OneElement is structually sparse yes.
And it is like something that should generally be preserved. Like how every Diagonal is a UpperTrianguar so `ProjectTo(::UpperTriangular)(::Diagonal)::Diagonal`  passes it through.
Same goes for like a OneElement that is on the diagonal

test/rulesets/Base/indexing.jl
oxinabox: We don't go through the `ProjectTo` right now Maybe we should.

oxinabox:
We should think about improving inferability so it stops making unionings over wether or not it is going to `reshape`.
One part i think is making the axes at type-param here.
But i think we also need changes in CRC


```
"""
    OneElement(val, ind, axes) <: AbstractArray

Extremely simple `struct` used for the gradient of scalar `getindex`.
"""
struct OneElement{T,N,I,A} <: AbstractArray{T,N}
  val::T
  ind::I
  OneElement(val::T, ind::I, axes) where {T<:Number, I<:NTuple{N,Int}, A<:NTuple{N,AbstractUnitRange}} where {N} = new{T,N,I, axes}(val, ind)
end
Base.size(::OneElement{<:Any,<:Any,<:Any, A}) where A = map(length, A)
Base.axes(::OneElement{<:Any,<:Any,<:Any, A}) where A = A
Base.getindex(xs::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==xs.ind, xs.val, zero(T))
```

Still this is a small union and doesn't really matter
but that is why i have to drop a lot of the inference tests

oxinabox:
Yota failure is unrelated

oxinabox:
Before
```julia
julia> x = rand(300,300);

julia> dx = OneElement(3.14, (231, 21), axes(x));

julia> @btime $x + $(collect(dx));  # Baseline
  31.978 μs (2 allocations: 703.17 KiB)

julia> @btime $x + $dx;
  61.043 μs (2 allocations: 703.17 KiB)

```

After:
```
julia> @btime $x + $(dx);
  21.056 μs (2 allocations: 703.17 KiB)
```

The orig version of this PR (and thus more or less what is in Zygote` was 2x slower than baseline, for accumulation,
with my fix it is now 30% faster than baseline.
(for a 300x300 matrix. This gets better the bigger the matrix).

For allocation no change vs what in Zygote, it remains 77,000x faster than baseline
```julia
julia> @btime collect(OneElement(3.14, (231, 21), axes($x)));  # approximate baseline
  54.806 μs (2 allocations: 703.17 KiB)

julia> @btime OneElement(3.14, (231, 21), axes($x));
  1.748 ns (0 allocations: 0 bytes)
```

I think this is worth a the type instability of going to a small union

oxinabox:
bumping @mcabbott or @ToucheSir for a review

mcabbott:
One of the things I don't like about OneElement is that it's tied to Array. Somewhere I had a prototype which stored BroadcastStyle too, so that it remembers that you indexed e.g. a CuArray. 

Maybe worth mentioning that https://github.com/JuliaArrays/FillArrays.jl/pull/235 added essentially the same struct that Zygote has to FillArrays. CR could consider getting it from there, although it's so simple that owning it is fine. I think FillArrays may overload some `mul!` methods etc, maybe worth thinking about whether any of those are desirable for CR's purposes.

Sorry not really a review, a bit swamped.

ToucheSir:
I have nothing intelligent to add on top of what Michael said, but it's cool that FillArrays has `OneElement` now. Would a port of the benchmark in https://github.com/FluxML/Zygote.jl/pull/962 help with making a decision here? I can look into that if there's interest.

oxinabox:
FillArrays is still pending in https://github.com/JuliaDiff/ChainRules.jl/issues/46

I am happy merging this as is, with other stuff to come in follow-ups.
but I do need one of you to to approve.

ToucheSir:
Happy to get this landed first. Before I approve, does anything need to be done to make CI happy?

oxinabox:
ok, now everything we expect to be passing is passing

ToucheSir (approved):

PR #718, opened 2023-05-19
Fix for julia 1.9

oxinabox:
https://github.com/JuliaLang/julia/pull/44358 broke our things on julia 1.9
this gets it back, but idk numerical linear algrebra so I am not sure it doesn't cost us anything.
Also inv is broken on GPU again.

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -342,20 +342,24 @@ function rrule(::typeof(\), A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:R
     project_B = ProjectTo(B)
 
     Y = A \ B
+    # Ever since https://github.com/JuliaLang/julia/pull/44358
+    # we need to use `pinv` rather than `/` to support both the cases of Y being scalar and array
+    # See also https://github.com/JuliaLang/julia/issues/28827 which would improve this
     function backslash_pullback(ȳ)
         Ȳ = unthunk(ȳ)
+        Ati = pinv(A')
         ∂A = @thunk begin
-            B̄ = A' \ Ȳ
+
+            B̄ = Ati * Ȳ
             Ā = -B̄ * Y'
-            Ā = add!!(Ā, (B - A * Y) * B̄' / A')
-            Ā = add!!(Ā, A' \ Y * (Ȳ' - B̄'A))
+            Ā = add!!(Ā, ((B - A * Y) * B̄') * Ati)
+            Ā = add!!(Ā, Ati * Y * (Ȳ' - B̄'A))
             project_A(Ā)
         end
-        ∂B = @thunk project_B(A' \ Ȳ)
+        ∂B = @thunk project_B(Ati * Ȳ)
         return NoTangent(), ∂A, ∂B
     end
     return Y, backslash_pullback
-
 end
 
 #####
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -343,24 +343,67 @@ function rrule(::typeof(\), A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:R
 
     Y = A \ B
 
-    Atf = factorize(A')
-    
     function backslash_pullback(ȳ)
         Ȳ = unthunk(ȳ)
+        
+        Ȳf = Ȳ
         @static if VERSION >= v"1.9"
             # Need to ensure Ȳ is an array since since https://github.com/JuliaLang/julia/pull/44358
-            Ȳ isa AbstractArray || Ȳ = [Ȳ]
+            if !isa(Ȳ, AbstractArray)
+                Ȳf = [Ȳ]
+            end
         end
-        Atf = factorize(A')
+        Yf = Y
+        @static if VERSION >= v"1.9" 
+            # Need to ensure Yf is an array since since https://github.com/JuliaLang/julia/pull/44358
+            if !isa(Y, AbstractArray)
+                Yf = [Y]
+            end
+        end
+        #@info "vars" typeof(Ȳ) typeof(Y) typeof(Yf) typeof(A) typeof(B)
         ∂A = @thunk begin
-            B̄ = Atf \ Ȳ
+            B̄ = A' \ Ȳf
             Ā = -B̄ * Y'
-            Ā = add!!(Ā, ((B - A * Y) * B̄') / Atf)
-            Ā = add!!(Ā, Atf \ Y * (Ȳ' - B̄'A))
+            t = (B - A * Y) * B̄'
+            @static if VERSION >= v"1.9" 
+                # Need to ensure t is an array since since https://github.com/JuliaLang/julia/pull/44358
+                if !isa(t, AbstractArray)
+                    t = [t]
+                end
+            end
+            Ā = add!!(Ā, t / A')
+            Ā = add!!(Ā, A' \ Yf * (Ȳ' - B̄'A))
             project_A(Ā)
         end
-        ∂B = @thunk project_B(Atf \ Ȳ)
+        ∂B = @thunk project_B(A' \ Ȳf)
         return NoTangent(), ∂A, ∂B
+    end
+    return Y, backslash_pullback
+end
+
+@static if VERSION >= v"1.9" 
+    # Need to ensure things are not scalar since since https://github.com/JuliaLang/julia/pull/44358
+    _maybe_descalar(x) = x isa AbstractArray ? x : [x]
+else
+    _maybe_descalar(x) = x
+end
+
+function rrule(A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:Real})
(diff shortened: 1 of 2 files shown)


oxinabox:
The test we need to pass is
```julia
using ChainRulesTestUtils
using Test
using ChainRules

@testset "$f" for f in (/, \)
        @testset "Matrix" begin
            for n in 3:5, m in 3:5
                A = randn(m, n)
                B = randn(m, n)
                test_rrule(f, A, B; check_inferred=false) # ChainRulesCore #407
            end
        end
        @testset "Vector" begin
            x = randn(10)
            y = randn(10)
            test_rrule(f, x, y; check_inferred=false) # ChainRulesCore #407
        end
        if f == (\)
            @testset "Matrix $f Vector" begin
                X = randn(10, 4)
                y = randn(10)
                test_rrule(f, X, y)
            end
            @testset "Vector $f Matrix" begin
                x = randn(10)
                Y = randn(10, 4)
                test_rrule(f, x, Y; output_tangent=Transpose(rand(4)))
            end
        else
            A = rand(2, 4)
            B = rand(4, 4)
            test_rrule(f, A, B; check_inferred=false) # ChainRulesCore #407
        end
    end
 ```

**Comment by oxinabox** [comment]
Oh interesting x86 (32bit) segfaults
https://github.com/JuliaDiff/ChainRules.jl/actions/runs/5025659382/jobs/9012921468?pr=718#step:6:213
definately a bug in Julia itself or in JuliaInterpretter though

**Comment by oxinabox** [comment]
@antoine-levitt you were going to review this?

File: src/rulesets/Base/arraymath.jl [hunk 1]
```diff
@@ -342,20 +342,24 @@ function rrule(::typeof(\), A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:R
...
+            B̄ = Ati * Ȳ
             Ā = -B̄ * Y'
-            Ā = add!!(Ā, (B - A * Y) * B̄' / A')
-            Ā = add!!(Ā, A' \ Y * (Ȳ' - B̄'A))
+            Ā = add!!(Ā, ((B - A * Y) * B̄') * Ati)
+            Ā = add!!(Ā, Ati * Y * (Ȳ' - B̄'A))
             project_A(Ā)
         end
-        ∂B = @thunk project_B(A' \ Ȳ)
+        ∂B = @thunk project_B(Ati * Ȳ)
```
oxinabox, line 359: I wonder if even if  we go a different path towards solving this it might not still be worth doing:

```suggestion
        Atf = factorize(A')
        ∂A = @thunk begin
            B̄ = Atf \ Ȳ
            Ā = -B̄ * Y'
            Ā = add!!(Ā, ((B - A * Y) * B̄') / Atf)
            Ā = add!!(Ā, Atf \ Y * (Ȳ' - B̄'A))
            project_A(Ā)
        end
        ∂B = @thunk project_B(Atf \ * Ȳ)
```

antoine-levitt:
Yes sorry I commented on the other issue here: https://github.com/JuliaLang/julia/pull/49915#issuecomment-1557415776. I don't think that's the way to go, because it computes explicit pseudo-inverses. Instead, you can just do something like `Ȳ isa AbstractArray || Ȳ = [Ȳ]` to get back the old behavior.

oxinabox:
Ok cool are you abble to review this PR and if it looks good now approve so i can merge?

antoine-levitt:
I don't understand why you're doing this. Can't this PR literally be `Ȳ isa AbstractArray || Ȳ = [Ȳ]`? There's no need to change anything else, is there?

antoine-levitt:
Ah, sorry, I didn't get that it was to factor only once. This is not an exact optimization, because `A \ b` has a slightly different logic than `factorize(A)\b`, but I don't think this is a problem, it should do the right thing nonetheless. So yes good to go I think.

antoine-levitt:
But
```
julia> randn(2) / factorize(randn(2,2))
ERROR: MethodError: no method matching /(::Vector{Float64}, ::LU{Float64, Tridiagonal{Float64, Vector{Float64}}, …})
```
Presumably that method needs to be added to factorize for that to work, I'll try to do a PR on julia. In the meantime, you can get by using the `\` method of factorized, which does work, and transposing.

antoine-levitt:
Yeah same problem just a bit below:
```
            Ā = add!!(Ā, A' \ Y * (Ȳ' - B̄'A))
```
That Y (but not the one above) needs to be []ified too.

oxinabox:
With that the `MethodError` is gone, but the DimensionMismatch, and the accuracy problem remains.

antoine-levitt:
What's your MWE? It looks like it's working for me. And what do you mean by accuracy?

oxinabox:
> What's your MWE? 

I am just running the tests:
To reproduce the particular ones that fail:
```julia
julia> using ChainRules, ChainRulesCore, ChainRulesTestUtils, Test

julia> @testset "$f" for f in (/, \)
               @testset "Matrix" begin
                   for n in 3:5, m in 3:5
                       A = randn(m, n)
                       B = randn(m, n)
                       test_rrule(f, A, B; check_inferred=false) # ChainRulesCore #407
                   end
               end
               @testset "Vector" begin
                   x = randn(10)
                   y = randn(10)
                   test_rrule(f, x, y; check_inferred=false) # ChainRulesCore #407
               end
               if f == (\)
                   @testset "Matrix $f Vector" begin
                       X = randn(10, 4)
                       y = randn(10)
                       test_rrule(f, X, y)
                   end
                   @testset "Vector $f Matrix" begin
                       x = randn(10)
                       Y = randn(10, 4)
                       test_rrule(f, x, Y; output_tangent=Transpose(rand(4)))
                   end
               else
                   A = rand(2, 4)
                   B = rand(4, 4)
                   test_rrule(f, A, B; check_inferred=false) # ChainRulesCore #407
               end
           end
```


You can see it failing in CI: https://github.com/JuliaDiff/ChainRules.jl/actions/runs/5143424642/jobs/9258419472?pr=718#step:6:804

By accuracy I mean: It is giving significantly numerically different results to finite differencing.
Might be numeric issues, might actually be the code is wrong now? Its a pretty big difference.

antoine-levitt:
That's just a typo, you wrote
```
        ∂A = #@thunk begin
            B̄ = A' \ Ȳ
```
which is doing `∂A = B̄ = A' \ Ȳ`, which is not what you want

oxinabox:
ah lol, in trying to simplify the problem I caused new ones

oxinabox:
Still getting errors.
I believe it is because for 
when 
```
│   typeof(Y) = Matrix{Float64} (alias for Array{Float64, 2})
│   typeof(A) = LinearAlgebra.Adjoint{Float64, Vector{Float64}}
└   typeof(B) = LinearAlgebra.Adjoint{Float64, Vector{Float64}}
```
we have that `(B - A * Y) * B̄'` is a `Float64`  (since it is `Vector * Adjoint{Vector}`).
When i assign that to a variable and wrap it in a `[t]` when it is scalar I get more errors.

See the whole complexity of this to me is really saying that, it is kinda needed that `/` accepts a scalar and a vector for matching that products of vectors and adjoint(vector)s produce scalars.

antoine-levitt:
```
function rrule(A::AbstractVecOrMat{<:Real}, B::AbstractVecOrMat{<:Real})
    Y = A \ B
    maybe_promote(x) = x isa AbstractArray ? x : [x]

    function backslash_pullback(ȳ)
        Ȳ = ȳ
        ∂A = begin
            B̄ = A' \ maybe_promote(Ȳ)
            Ā = -B̄ * Y'
            Ā += maybe_promote((B - A * Y) * B̄') / A'
            Ā += (A' \ maybe_promote(Y)) * (Ȳ' - B̄'A)
            (Ā)
        end
        ∂B = (A' \ maybe_promote(Ȳ))
        return ∂A, ∂B
    end
    return Y, backslash_pullback
end


A = randn(2)
B = randn(2)
rrule(A, B)[2](1.0)
```
works for me. 

> See the whole complexity of this to me is really saying that, it is kinda needed that / accepts a scalar and a vector for matching that products of vectors and adjoint(vector)s produce scalars.

This workaround is needed because of the interaction of vector' * vector being a scalar (which is reasonable and isn't really questionned) and vector \ vector being pinv(A)*B. I think there's a consensus for saying that vector/vector and vector\vector working was a bad leftover from matlab (where it's also a bad idea, but unavoidable since there's no distinction between vectors and 2x1 matrices there), and that it should be removed in 2.0 in favor of matrix\vector and vector/matrix (in which case, to get the old A/B for A and B vectors, you'd do A/hcat(B)), so that this issue just goes away. Again I'm pretty sure this issue is the result of an overzealous implementation of the rule to support the vector\vector usecase, which pretty much nobody uses (and nobody should use, since hopefully it will be gone in 2.0).

oxinabox:
ok, merging this. Though can revert if https://github.com/JuliaLang/julia/pull/49915 goes through.

I am going to just note that 1.6 on 64 bit is broken.
As it seems unreleated to this PR. And seems like a bug in Julia itself

PR #721, opened 2023-06-01
Avoid lowering to `Base.rest()`, as it infers poorly

staticfloat:
While it would be best to fix `Base.rest()` to be able to infer properly, at the moment the destructuring with a splat in the arguments to this `frule` end up causing inference issues in Diffractor.

### DIFF
--- a/src/rulesets/Base/fastmath_able.jl
+++ b/src/rulesets/Base/fastmath_able.jl
@@ -169,8 +169,8 @@ let
         @scalar_rule x / y (one(x) / y, -(Ω / y))
         
         ## many-arg +
-        function frule((_, Δx, Δy...), ::typeof(+), x::Number, ys::Number...)
-            +(x, ys...), +(Δx, Δy...)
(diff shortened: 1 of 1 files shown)


oxinabox:
Is `Base.tail` public API?

This is probably fine as both the nominally correct type has `Base.tail` implemented (https://github.com/JuliaDiff/ChainRulesCore.jl/blob/caf8692ca1bf5fabeb4afc86340eef9456da469c/src/tangent_types/tangent.jl#L106)
and the most commonly used type (`Tuple`) has it.

However, right now CI is broken due to not having https://github.com/JuliaDiff/ChainRules.jl/pull/721  in yet.
Hopefully i hwill have that in today and can also merge this.

It is pretty bad that this breaks inference so badly, it looks so idiomatic. Can we open an upstream issue in JuliaLang/julia about it?

oxinabox (approved):

staticfloat:
I have opened an error report and an MWE here: https://github.com/JuliaLang/julia/issues/50046

PR #724, opened 2023-06-29
Remove odd overload of calling array on an array

oxinabox:
Pointed out by @Marius311
not sure how this happened.
I blame Linux/X's  middle-click paste.

This code was unused, looks like it was part of a draft for rewriting `/` but it wasn't completed (no use of `add!!`)
So we can just delete it.

@marius311 if you want to give this a look and give it an approve, I can merge this


marius311:
Confirm that fixes the thing I had run into, thanks for the quick attention!

Issue #719, opened 2023-05-28
Errors for gradient and hessian of logabsdet of sparse matrix


ElOceanografo:
The following error:
```julia
using LinearAlgebra, SparseArrays, Zygote
Zygote.gradient(x -> logabsdet(spdiagm(x)), rand(10))
```
<details>
<summary>
ERROR: The inverse of a sparse matrix can often be dense...
</summary>
```julia
ERROR: The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please either convert your matrix to a dense matrix, e.g. by calling `Matrix` or if `A` can be factorized, use `\` on the dense identity matrix, e.g. `A \ Matrix{eltype(A)}(I, size(A)...)` restrictions of `\` on sparse lhs applies. Altenatively, `A\b` is generally preferable to `inv(A)*b`
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:35
  [2] inv(A::SparseMatrixCSC{Float64, Int64})
    @ SparseArrays C:\Users\sam.urmy\.julia\juliaup\julia-1.9.0+0.x64.w64.mingw32\share\julia\stdlib\v1.9\SparseArrays\src\linalg.jl:1448
  [3] (::Zygote.var"#833#836"{SparseMatrixCSC{Float64, Int64}})(Δ::Tuple{Float64, Float64})
    @ Zygote C:\Users\sam.urmy\.julia\packages\Zygote\HTsWj\src\lib\array.jl:379
  [4] (::Zygote.var"#3123#back#837"{Zygote.var"#833#836"{SparseMatrixCSC{Float64, Int64}}})(Δ::Tuple{Float64, Float64})
    @ Zygote C:\Users\sam.urmy\.julia\packages\ZygoteRules\OgCVT\src\adjoint.jl:71
  [5] Pullback
    @ C:\Users\sam.urmy\.julia\juliaup\julia-1.9.0+0.x64.w64.mingw32\share\julia\stdlib\v1.9\LinearAlgebra\src\generic.jl:1685 [inlined]
[...]

PR #730, opened 2023-08-09
det, logdet, and logabsdet rrules for SparseMatrixCSC

ElOceanografo:
This PR implements determinant rrules for unstructured sparse matrices. The equivalent rules for `AbstractMatrices` didn't work, since they require calculating a matrix inverse, which errors by default for sparse matrices to avoid unintentionally creating huge dense matrices. Fortunately, the adjoint for the determinant functions does not depend on all the values of the inverse matrix (see [here](https://doi.org/10.18637/jss.v070.i05), section 4.1 and Equation 8). Using the "inverse subset" algorithm, you can efficiently calculate just the sparse subset of the inverse matrix which you need. I just implemented that algorithm in this [small package](https://github.com/ElOceanografo/SparseInverseSubset.jl/tree/main), which enabled me to write the rrules here.

I put these rules and their tests in with the "structured" matrix rule sets. Would be happy to move them to new "sparse" src and test files if that fits the organizational structure better. I am also pretty new to writing chain rules, so any feedback or corrections are more than welcome!

This PR will fix https://github.com/JuliaDiff/ChainRules.jl/issues/719, and should also help fix https://github.com/TuringLang/DistributionsAD.jl/issues/89.


oxinabox:
CI is broken for this new feature on julia 1.6: 
see https://github.com/JuliaDiff/ChainRules.jl/actions/runs/5812786772/job/15758921347?pr=730#step:6:300

(It's also broken on nightly for other reasons that we are going to have to look into.
But it works for this feature on nightly.)

oxinabox:
> I put these rules and their tests in with the "structured" matrix rule sets. 

Yes, please put them in this file:
https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/SparseArrays/sparsematrix.jl
and the tests in the matching file

ElOceanografo:
The logs do not say exactly what broke in the tests for 1.6, but I suspect it is because prior to Julia 1.7 there was no method implemented for `logabsdet` of sparse matrices (see this [issue](https://github.com/JuliaLang/julia/issues/38463) and [PR](https://github.com/JuliaLang/julia/pull/40716)). 

If that's the problem, what's the best solution? A conditional statement that defines the missing [method](https://github.com/ElOceanografo/julia/blob/98651e6a4277ab3b227207ab2991ffbb92ab56d0/stdlib/SuiteSparse/src/umfpack.jl#L429) if `VERSION < v"1.7"`?

ElOceanografo:
Went ahead and added a conditional definition of the method for `logabsdet` that was introduced in Julia v1.7. A little hacky, but tests pass on 1.6 now. @oxinabox let me know if that seems like an ok solution...

oxinabox (approved):
LGTM, thanks!

PR #735, opened 2023-08-30
Allow N-dimensional arrays in `sort` rules

pxl-th:


pxl-th:
Restricted tests to 1.9+. Otherwise it does not break anything.

PR #736, opened 2023-09-03
non_differentiable unique

CarloLucibello:
Some array types for which it is safe to assume non differrentiability of `unique`. 
I don't see the test file corresponding to the `nondiff.jl` source file. It is ok to leave these untested?

The need for these rules arised in https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/333

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.54.1"
+version = "1.55.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


mcabbott (approved):
Seems fine, sad that our AD is so dumb that this is necessary...

CarloLucibello:
I don't have merge rights

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.54.1"
+version = "1.55.0"
```
mcabbott, line 3: This fails: 

> Does not meet sequential version number guideline: version 1.55.0 skips over 1.54.0

PR #738, opened 2023-09-18
Add rules needed for mutation

oxinabox:
This is the partner to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/626
and requires that PR to be merged first.


src/rulesets/Base/array.jl
```diff
@@ -86,7 +86,14 @@ _instantiate_zeros(ẋs::AbstractArray{<:AbstractArray}, xs) = ẋs
...
-    return copyto!(y, x), copyto!(ẏ, ẋ)
+    if ẏ isa AbstractZero
+        # it's allowed to have an imutable zero tangent for ẏ as long as ẋ is zero
+        # TODO should this be handled here or in the AD?
+        @assert iszero(ẋ)
+    else
+        copyto!(ẏ, ẋ)
+    end
+    return copyto!(y, x), ẏ
 end
```
oxinabox, line 97:
```suggestion
    return copyto!(y, x), copyto!(ẏ, ẋ)
end
```

PR #739, opened 2023-09-21
add frules for getfield

oxinabox:
This are taken directly from another project.
Suprising we didn't have them already

### DIFF
--- a/test/rulesets/Base/indexing.jl
+++ b/test/rulesets/Base/indexing.jl
@@ -1,9 +1,11 @@
+struct FooTwoField
+    x::Float64
+    y::Float64
+end
+ 
+
 @testset "getfield" begin
-    struct Foo
-        x::Float64
-        y::Float64
-    end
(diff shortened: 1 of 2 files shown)


test/rulesets/Base/indexing.jl
```diff
@@ -1,3 +1,17 @@
+@testset "getfield" begin
+    struct Foo
+        x::Float64
+        y::Float64
+    end
+    test_frule(getfield, Foo(1.5, 2.5), :x, check_inferred=false)
+    
+    test_frule(getfield, (; a=1.5, b=2.5), :a, check_inferred=false)
```
oxinabox, line 8: These actually infer fine as long as the `frule` gets inlined somewhere where the symbol is known at compile time.
I.e. almost all uses hopefully.

oxinabox:
Apparently these CI failures are real.
They do not occur on main, from what i can tell.

oscardssmith (approved):
This looks good to me (other than the fact that CI is unhappy)

Issue #684, opened 2022-11-11
Rules for `kron`


ToucheSir:
Zygote currently [punts this to AD](https://github.com/FluxML/Zygote.jl/blob/d39ab5905b1d6e983f59f2f4566476845b002e9d/src/lib/array.jl#L315-L325), but that trick no longer works on nightly because `kron(::AbstractVector, ::AbstractVector)` stopped calling `kron(::AbstractMatrix, ::AbstractMatrix)`.

oxinabox:
This is probably pretty fun, so I am am labelling this a good first issue.
The math is not complex, its just a little fiddy playing around with repeat and such

PR #741, opened 2023-09-25
Add rules and tests for `kron`

simsurace:
In Julia 1.9 there was an internal change in `kron` that introduced some mutation, which has made Zygote unable to differentiate `kron`. Here, we add some rules to restore that ability.

Discovered in https://github.com/JuliaGaussianProcesses/TemporalGPs.jl/pull/115

### DIFF
--- a/src/rulesets/LinearAlgebra/dense.jl
+++ b/src/rulesets/LinearAlgebra/dense.jl
@@ -399,7 +399,7 @@ end
 ##### `kron`
 #####
 
-@static if VERSION ≥ v"1.9.0"
+@static if VERSION ≥ v"1.9.0-DEV.1267"
     function frule((_, Δx, Δy), ::typeof(kron), x::AbstractVecOrMat{<:Number}, y::AbstractVecOrMat{<:Number})
         return kron(x, y), kron(Δx, y) + kron(x, Δy)
     end


devmotion:
> made Zygote unable to differentiate `kron`

Did the workaround in https://github.com/FluxML/Zygote.jl/pull/1378 not fix it? As mentioned in #684, should ideally be fixed in ChainRules nevertheless, but I'm a bit curious.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,83 @@ function rrule(
...
+
+function frule((_, Δx, Δy), ::typeof(kron), x, y)
+    return kron(x, y), kron(Δx, y) + kron(x, Δy)
+end
+
+function rrule(::typeof(kron), x::AbstractVector, y::AbstractVector)
+    function kron_pullback(z̄)
+        x̄ = zero(x)
+        ȳ = zero(y)
+        m = firstindex(z̄)
```
devmotion, line 410: Is it not needed to `unthunk` `z̄`?

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,83 @@ function rrule(
...
+        @inbounds for i in eachindex(x)
```
devmotion, line 411: Is this safe (mainly thinking about indices of `z̄`)?

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,83 @@ function rrule(
...
+            xi = x[i]
+            for k in eachindex(y)
+                x̄[i] += y[k]' * z̄[m]
+                ȳ[k] += xi' * z̄[m]
```
devmotion, line 415: This won't work with immutable arrays. I wonder if the whole rule could be rewritten using `map`/`broadcast`/`dot` or some other linalg operations that do not require preallocating the output array.

willtebbutt: This is a really good point -- @simsurace we will need this to work with StaticArrays in order to get TemporalGPs working again

simsurace:
> > made Zygote unable to differentiate `kron`
> 
> Did the workaround in [FluxML/Zygote.jl#1378](https://github.com/FluxML/Zygote.jl/pull/1378) not fix it? As mentioned in #684, should ideally be fixed in ChainRules nevertheless, but I'm a bit curious.

Thanks for commenting. I think @willtebbutt said that he will have a look at these rules later on.
I don't know if it is related, but the abovementioned fix predates Julia 1.9 by several months. I observed the breakage when upgrading Julia from 1.8 to 1.9.

src/rulesets/LinearAlgebra/dense.jl
simsurace: Probably.

src/rulesets/LinearAlgebra/dense.jl
simsurace: It could well not be safe, I'm not sure.

src/rulesets/LinearAlgebra/dense.jl
simsurace: It may be advantageous to do that, yes.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,83 @@ function rrule(
     end
     return Ω, lyap_pullback
 end
+
+#####
+##### `kron`
+#####
+
+function frule((_, Δx, Δy), ::typeof(kron), x, y)
```
devmotion, line 402: Another point: Maybe it would be safer to restrict the types of the arguments to e.g.
```suggestion
function frule((_, Δx, Δy), ::typeof(kron), x::AbstractVecOrMat{<:Number}, y::AbstractVecOrMat{<:Number})
```
or possibly even
```suggestion
function frule((_, Δx, Δy), ::typeof(kron), x::AbstractVecOrMat{<:Real}, y::AbstractVecOrMat{<:Real})
```
(same applies to the other rules below)

simsurace:
Hi all, I rewrote the rules and now all the tests pass. There is probably opportunity to optimize them, please let me know.

simsurace:
Ok, did not test on Julia 1.6. Apparently this requires special care

simsurace:
Why don't we see the full stack traces here? Is it due to using JuliaInterpreter?

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,51 @@ function rrule(
...
+#####
+
+function frule((_, Δx, Δy), ::typeof(kron), x::AbstractVecOrMat{<:Number}, y::AbstractVecOrMat{<:Number})
+    return kron(x, y), kron(Δx, y) + kron(x, Δy)
+end
+
+function rrule(::typeof(kron), x::AbstractVector{<:Number}, y::AbstractVector{<:Number})
+    function kron_pullback(z̄)
+        dz = reshape(unthunk(z̄), length(y), length(x))
+        x̄ = @thunk conj.(dz' * y)
```
sethaxen, line 409: These need a projection using `ProjectTo`

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,51 @@ function rrule(
...
+        return NoTangent(), x̄, ȳ
+    end
+    return kron(x, y), kron_pullback
+end
+
+function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractMatrix{<:Number})
+    function kron_pullback(z̄)
+        dz = reshape(unthunk(z̄), size(y, 1), size(x, 1), size(y, 2), size(x, 2))
+        x̄ = @thunk conj.(dot.(eachslice(dz, dims = (2, 4)), Ref(y)))
+        ȳ = @thunk conj.(dot.(eachslice(dz; dims = (1, 3)), Ref(x)))
```
sethaxen, line 440: Here can't you reverse the order of the arguments to `dot` to eliminate the call to `conj`?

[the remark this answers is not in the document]
simsurace: I tried different permutations...but it was tricky to find ones that work for both real and complex. Can you clarify how you'd change the order?

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,51 @@ function rrule(
...
+        ȳ = @thunk dz * conj.(x)
+        return NoTangent(), x̄, ȳ
+    end
+    return kron(x, y), kron_pullback
+end
+
+function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractVector{<:Number})
+    function kron_pullback(z̄)
+        dz = reshape(unthunk(z̄), length(y), size(x)...)
+        x̄ = @thunk conj.(dot.(eachslice(dz; dims = (2, 3)), Ref(y)))
```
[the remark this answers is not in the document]
simsurace: Yeah, makes sense. I can wrap all those rules with a version check.

src/rulesets/LinearAlgebra/dense.jl
simsurace: Ok, I'll take your word for it..

simsurace: How would we write a test that would catch this missing `ProjectTo`?

sethaxen: Probably a test using a type that has structural zeros, e.g. a one-hot vector. Not certain if a sparse type on the stdlib would work or if you'd need to add your own.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,55 @@ function rrule(
...
+            return NoTangent(), x̄, ȳ
+        end
+        return kron(x, y), kron_pullback
+    end
+
+    function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractVector{<:Number})
+        function kron_pullback(z̄)
+            dz = reshape(unthunk(z̄), length(y), size(x)...)
+            x̄ = @thunk conj.(dot.(eachslice(dz; dims = (2, 3)), Ref(y)))
+            ȳ = @thunk conj.(dot.(eachslice(dz; dims = 1), Ref(x)))
```
sethaxen, line 423:
```suggestion
            x̄ = @thunk project_x(dot.(Ref(y), eachslice(dz; dims = (2, 3))))
            ȳ = @thunk project_y(dot.(Ref(x), eachslice(dz; dims = 1)))
```

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,55 @@ function rrule(
...
+            dz = reshape(unthunk(z̄), length(y), length(x))
+            x̄ = @thunk(project_x(conj.(dz' * y)))
+            ȳ = @thunk(project_y(dz * conj.(x)))
+            return NoTangent(), x̄, ȳ
+        end
+        return kron(x, y), kron_pullback
+    end
+
+    function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractVector{<:Number})
+        function kron_pullback(z̄)
```
sethaxen, line 420:
```suggestion
        project_x = ProjectTo(x)
        project_y = ProjectTo(y)
        function kron_pullback(z̄)
```

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,55 @@ function rrule(
...
+            dz = reshape(unthunk(z̄), length(y), size(x)...)
+            x̄ = @thunk conj.(dot.(eachslice(dz; dims = (2, 3)), Ref(y)))
+            ȳ = @thunk conj.(dot.(eachslice(dz; dims = 1), Ref(x)))
+            return NoTangent(), x̄, ȳ
+        end
+        return kron(x, y), kron_pullback
+    end
+
+    function rrule(::typeof(kron), x::AbstractVector{<:Number}, y::AbstractMatrix{<:Number})
+        function kron_pullback(z̄)
```
sethaxen, line 430:
```suggestion
        project_x = ProjectTo(x)
        project_y = ProjectTo(y)
        function kron_pullback(z̄)
```

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,55 @@ function rrule(
...
+            dz = reshape(unthunk(z̄), size(y, 1), length(x), size(y, 2))
+            x̄ = @thunk conj.(dot.(eachslice(dz; dims = 2), Ref(y)))
+            ȳ = @thunk conj.(dot.(eachslice(dz; dims = (1, 3)), Ref(x)))
+            return NoTangent(), x̄, ȳ
+        end
+        return kron(x, y), kron_pullback
+    end
+
+    function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractMatrix{<:Number})
+        function kron_pullback(z̄)
```
sethaxen, line 440:
```suggestion
        project_x = ProjectTo(x)
        project_y = ProjectTo(y)
        function kron_pullback(z̄)
```

sethaxen (changes requested):

test/rulesets/LinearAlgebra/dense.jl
```diff
@@ -159,4 +159,20 @@
             test_rrule(lyap, A, C)
         end
     end
+    @testset "kron" begin
```
sethaxen, line 162:
```suggestion
    VERSION ≥ v"1.9.0" && @testset "kron" begin
```

src/rulesets/LinearAlgebra/dense.jl
sethaxen: Actually, a simple test of `ProjectTo` is to make one arg real and one arg complex. Projection will real-ify the cotangent for the real argument, and the test would otherwise fail without it.

simsurace:
Ok, I made the suggested changes and added tests to check the correct behavior of the projections. However, we have some type inference problem in the matrix-matrix case.

simsurace:
The problem is this:
```julia
julia> x = Diagonal(rand(2)); y = Diagonal(rand(2)); z, pb = rrule(kron, x, y);

julia> @code_warntype unthunk(pb(z)[2])
MethodInstance for ChainRulesCore.unthunk(::Thunk{ChainRules.var"#2318#2321"{Base.ReshapedArray{Float64, 4, Diagonal{Float64, Vector{Float64}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Diagonal{Float64, Vector{Float64}}, ProjectTo{Diagonal, NamedTuple{(:diag,), Tuple{ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}}}})
  from unthunk(x::Thunk) @ ChainRulesCore ~/.julia/packages/ChainRulesCore/0t04l/src/tangent_types/thunks.jl:204
Arguments
  #self#::Core.Const(ChainRulesCore.unthunk)
  x::Thunk{ChainRules.var"#2318#2321"{Base.ReshapedArray{Float64, 4, Diagonal{Float64, Vector{Float64}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Diagonal{Float64, Vector{Float64}}, ProjectTo{Diagonal, NamedTuple{(:diag,), Tuple{ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}}}}
Body::Any
1 ─      nothing
│   %2 = Base.getproperty(x, :f)::ChainRules.var"#2318#2321"{Base.ReshapedArray{Float64, 4, Diagonal{Float64, Vector{Float64}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Diagonal{Float64, Vector{Float64}}, ProjectTo{Diagonal, NamedTuple{(:diag,), Tuple{ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}}}}
│   %3 = (%2)()::Any
└──      return %3
```
Any ideas how to make the `unthunk`ing type-stable here?

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,64 @@ function rrule(
     end
     return Ω, lyap_pullback
 end
+
+#####
+##### `kron`
+#####
+
+@static if VERSION ≥ v"1.9.0"
```
devmotion, line 402: Was the AD problem introduced by https://github.com/JuliaLang/julia/pull/43724? Then this check could be improved to
```suggestion
@static if VERSION ≥ v"1.9.0-DEV.1267"
```

simsurace: I don't know if that's the case.

src/rulesets/LinearAlgebra/dense.jl
```diff
@@ -394,3 +394,64 @@ function rrule(
...
+        return kron(x, y), kron_pullback
+    end
+
+    function rrule(::typeof(kron), x::AbstractMatrix{<:Number}, y::AbstractMatrix{<:Number})
+        project_x = ProjectTo(x)
+        project_y = ProjectTo(y)
+        function kron_pullback(z̄)
+            dz = reshape(unthunk(z̄), size(y, 1), size(x, 1), size(y, 2), size(x, 2))
+            x̄ = @thunk(project_x(_dot_collect.(Ref(y), eachslice(dz; dims = (2, 4)))))
+            ȳ = @thunk(project_y(_dot_collect.(Ref(x), eachslice(dz; dims = (1, 3)))))
```
mcabbott, line 449: I was wondering if you have to make slices, given that kron is just reshape and `.*`. So here's an attempt to do without:

```julia
using ChainRulesCore

function pr_rule(x::AbstractMatrix{<:Number}, y::AbstractMatrix{<:Number})  # from https://github.com/JuliaDiff/ChainRules.jl/pull/741
    project_x = ProjectTo(x)
    project_y = ProjectTo(y)
    function kron_pullback(z̄)
        dz = reshape(unthunk(z̄), size(y, 1), size(x, 1), size(y, 2), size(x, 2))
        x̄ = @thunk(project_x(dot.(Ref(y), eachslice(dz; dims = (2, 4)))))
        ȳ = @thunk(project_y(dot.(Ref(x), eachslice(dz; dims = (1, 3)))))
        return NoTangent(), x̄, ȳ
    end
end

# using TensorCast
# mykron(x,y) = @cast z[(a,b), (c,d)] := x[b,d] * y[a,c]
# @pretty @cast z[(a,b), (c,d)] := x[b,d] * y[a,c]

function shape_rule(x::AbstractMatrix, y::AbstractMatrix)
    function back(dz)
        x4 = reshape(x, 1, size(x,1), 1, size(x,2))
        y4 = reshape(y, size(y,1), 1, size(y,2), 1)
        dz4 = reshape(unthunk(dz), size(y,1), size(x,1), size(y,2), size(x,2))
        dx = @thunk ProjectTo(x)(reshape(sum(dz4 .* y4, dims=(1,3)), size(x)))  # might be missing conj
        dy = @thunk ProjectTo(y)(reshape(sum(dz4 .* x4, dims=(2,4)), size(y)))
        0, dx, dy
      end
end

let x = rand(10,10), y = rand(10,10)
  b1 = pr_rule(x, y)
  b2 = shape_rule(x, y)
  z = kron(x,y)
  _, dx1, _ = @btime map(unthunk, $b1($z))
  _, dx2, _ = @btime map(unthunk, $b2($z))
  dx1 ≈ dx2
end
#   min 28.250 μs, mean 30.315 μs (4 allocations, 1.88 KiB)
#   min 25.208 μs, mean 43.459 μs (32 allocations, 159.28 KiB)
# true
```
It's a pity to allocate these big arrays `dz4 .* y4` but still seems quicker. Possibly we could use lazy broadcasting to avoid that:
```julia
bc = Broadcast.instantiate(Broadcast.broadcasted(*, [1 2 3], [4, 5]));
sum(bc)  # OK
sum(bc; dims=1) # ERROR: MethodError: no method matching reducedim_init(::typeof(identity), ::typeof(Base.add_sum), ::Base.Broadcast.Broadcasted{…}, ::Int64)
sum!([0 0 0], bc)  # ERROR: MethodError: no method matching sum!(::Matrix{Int64}, ::Base.Broadcast.Broadcasted
sum(bc; dims=1, init=0.0)  # OK, not sure if it's fast or not
```

src/rulesets/LinearAlgebra/dense.jl
sethaxen: It looks so, given that that PR added `kron!`. @devmotion I've always wondered, how does someone find out the Julia dev version for a particular commit?

devmotion: You can obtain it with the contrib/commit-name.sh script in the Julia repo: https://github.com/JuliaLang/Compat.jl#developer-tips

simsurace: I can't check it myself, but if it's obvious to you, I can approve this suggestion

src/rulesets/LinearAlgebra/dense.jl
simsurace: Does this result scale to larger arrays?

mcabbott: Result meaning speed difference? It will vary with size & machine. On very small arrays reshaping is faster.

Issues with StaticArrays will be similar at all sizes.

I think broadcasting over slices will work badly on CuArrays, and tend to make Arrays. But right now neither idea seems to work, not sure why 
```
julia> using Metal

julia> bk = pr_rule(MtlArray(rand(Float32, 3,3)), MtlArray(rand(Float32, 3,3)));

julia> bk(MtlArray(rand(Float32, 9,9)))[2] |> unthunk
ERROR: GPU compilation of MethodInstance for (::GPUArrays.var"#broadcast_kernel#26")(::Metal.mtlKernelContext, ::MtlDeviceMatrix{…}, ::Base.Broadcast.Broadcasted{…}, ::Int64) failed
KernelError: passing and using non-bitstype argument

julia> bk2 = shape_rule(MtlArray(rand(Float32, 3,3)), MtlArray(rand(Float32, 3,3)));

julia> bk2(MtlArray(rand(Float32, 9,9)))[2] |> unthunk
ERROR: could not load symbol "LLVMExtraAddPropagateJuliaAddrspaces":
dlsym(RTLD_DEFAULT, LLVMExtraAddPropagateJuliaAddrspaces): symbol not found
```

simsurace: If the `reshape` version is not strictly better than the current one, especially for large arrays, I would propose to keep the current version and put further optimizations in a separate PR.

mcabbott: A bit curious at what sizes it's slower for you?

simsurace: I haven't benchmarked anything myself yet. I will give it a go later.

ToucheSir: One way to ensure any implementation isn't excluding all GPU array types would be to toss a `@gpu` in front of the new tests, no?

PR #744, opened 2023-10-01
Copy JuliaFormatter config from ChainRulesCore.jl

hyrodium:
This PR adds JuliaFormatter config. See https://github.com/JuliaDiff/ChainRulesCore.jl/pull/482 for more information.
I haven't formatted the current source so as not to make conflicts with other PRs.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "blue"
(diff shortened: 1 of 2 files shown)


oxinabox:
I am not 100% convinced a formatter is a good idea. but we should be consistent about having one across the chainrules family of packages. So I am merging this

oxinabox:
thanks

PR #750, opened 2023-11-06
Add version bounds for stdlibs

oxinabox:
Needed to fix https://github.com/JuliaRegistries/General/pull/94682#issuecomment-1792138621

Once CI passes I am going to merge this without waiting for review since this is formulatic


oxinabox:
break in 1.9 is unrelated. Need to look into that though

PR #759, opened 2023-11-24
Allow unzip to return JLArray

oxinabox:
I do not know if this is the correct fix, this is just allowing it to return either.
Some change upstream i think in julia itself (since this test only failed in 1.x not in 1.6) changed something that cause this to now return a `JLArray` not a `ReinterprettedArray`.
But I think that's fine, good even.
But idk. @mcabbott ?

### DIFF
--- a/test/unzipped.jl
+++ b/test/unzipped.jl
@@ -90,7 +90,7 @@ using ChainRules: unzip_broadcast, unzip #, unzip_map
         @test unzip(jl([(missing,2), (missing,4), (missing,6)]))[2] == jl([2, 4, 6])
         @test unzip(jl([(1,), (3,), (5,)]))[1] == jl([1, 3, 5])
 
-        # depending on Julia version may get ReinterpretArray or may get JLArray
+        # depending on Julia/package versions, may get ReinterpretArray or JLArray
         # Either is acceptable
         @test isa(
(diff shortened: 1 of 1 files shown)


oxinabox:
Can anyone review this please?

test/unzipped.jl
```diff
@@ -87,11 +87,14 @@ using ChainRules: unzip_broadcast, unzip #, unzip_map
...
 
         @test unzip(jl([(1,2), (3,4), (5,6)])) == (jl([1, 3, 5]), jl([2, 4, 6]))
-
         @test unzip(jl([(missing,2), (missing,4), (missing,6)]))[2] == jl([2, 4, 6])
-        @test unzip(jl([(missing,2), (missing,4), (missing,6)]))[2] isa Base.ReinterpretArray
-
         @test unzip(jl([(1,), (3,), (5,)]))[1] == jl([1, 3, 5])
-        @test unzip(jl([(1,), (3,), (5,)]))[1] isa Base.ReinterpretArray
+
+        # depending on Julia version may get ReinterpretArray or may get JLArray
```
mcabbott, line 93:
```suggestion
        # depending on Julia/package versions, may get ReinterpretArray or JLArray
```

mcabbott (approved):
This seems fine... I managed to get both types by `] up`, so it's not just Julia version, but maybe something in the many GPU packages.

PR #760, opened 2023-12-10
Allow single indexing of arrays of GPU arrays

ToucheSir:
This fixes https://github.com/FluxML/Zygote.jl/issues/1470, which was caused by Zygote switching to the CR rule. I can't think of any scenarios where this wouldn't work and the current implementation would, so hopefully it's a non-controversial fix.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.58.0"
+version = "1.58.1"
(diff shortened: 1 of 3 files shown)


ToucheSir:
Test failures all look to be unrelated to this PR. Does anyone know what's going on there?

oxinabox:
the  **CI / Julia 1.6 - ubuntu-latest - x86** has been broken for ages.
It is I assume a julia bug since it is a segfault.
and I don't track Nightly failures.

PR #761, opened 2023-12-14
Properly support StaticArrays construction

avik-pal:
I don't have any strong opinions whether it stays here or in StaticArrays https://github.com/JuliaArrays/StaticArrays.jl/pull/1224, but SciMLSensitivity is not really the right place :sweat:

cc @ChrisRackauckas


oxinabox:
It's our long standing policy that ChainRules.jl doesn't take rules except for Base and stdlibs.
It has been my observation that in practice the rules tend to benifit from the expertise and understanding of the original source implementation.
It also would rapidly exceed our ability to maintain it we took the whole ecosystem.

Ergo this should be a package extension in one of the StaticArrays packages.

Issue #769, opened 2024-01-15
`frule`s for `one` are ambiguous


MasonProtter:
```julia
  MethodError: frule(::Tuple{ZeroTangent, Float64}, ::typeof(one), ::Float64) is ambiguous.
  
  Candidates:
    frule(::Tuple, ::typeof(one), var"1277")
      @ ChainRules ~/.julia/packages/ChainRules/pEOSw/src/rulesets/Base/nondiff.jl:336
    frule(::Any, ::typeof(one), x::Number)
      @ ChainRules ~/.julia/packages/ChainRules/pEOSw/src/rulesets/Base/base.jl:6

  Possible fix, define
    frule(::Tuple, ::typeof(one), ::Number
```

These methods come from
```julia
@non_differentiable one(::Any)
```
and
```julia
@scalar_rule one(x) ZeroTangent()
```
respectively. It seems like either `@non_differentiable` should not use a `::Tuple` type restriction, or `@scalar_rule` should add a `::Tuple` type restriction.

PR #770, opened 2024-01-19
Try deleting the `one(::Any)` and `zero(::Any)` non_diff methods

MasonProtter:
Attempt at alternative to https://github.com/JuliaDiff/ChainRulesCore.jl/pull/653 for fixing https://github.com/JuliaDiff/ChainRules.jl/issues/769


oxinabox:
Yes, this seems to fix the problem in Diffractor and these methods are kinda wrong anyway,
since we do need to have a zero value here, the tanget space does (in general) exist.
This matters for forward mutation support in Diffractor where `NoTangent()` gets left as is, but `ZeroTangent` gets made into a a concrete zero value if it is mutable.

(`zeros` is also wrong, I'll make another PR for that)

oxinabox:
Remaining CI issues are unrelated so merging

Issue #763, opened 2023-12-20
SparseInverseSubset.jl dependency causes issues on non-GPL Julia builds


ElOceanografo:
ChainRules depends on SparseInverseSubset.jl to be able to differentiate through determinants of sparse matrices. However, this requires the `CHOLMOD` type for sparse Cholesky decompositions from SuiteSparse.jl, which is unavailable on non-GPL Julia builds, causing errors. See [here](https://github.com/ElOceanografo/SparseInverseSubset.jl/pull/2) and [here](https://github.com/ElOceanografo/SparseInverseSubset.jl/issues/3).

I think the cleanest way to handle this would be to add conditional statements for importing SparseInverseSubset.jl, and for the `rrule`s for `logdet` and `logabsdet` of sparse matrices so that these are omitted if `Base.USE_GPL_LIBS == false`. If that sound like the right approach I will make a PR.

oxinabox:
I think that does indeed sound right
wrap it in top level
```
if Base.USE_GPL_LIBS  # don't define these if we don't have CHOLMOD from SuiteSparse.jl
    using SparseInveserSubset
    function rrule(...)
    ...
end
```

that way we will get the usual behavours as if the rrule had not being defined -- because it wouldn't have been defined.

PR #772, opened 2024-01-23
Fix bugs with SparseInverseSubset on non-GPL Julia builds

ElOceanografo:
Bug fix for https://github.com/JuliaDiff/ChainRules.jl/issues/763: the code causing errors on non-GPL Julia builds is now wrapped in a conditional block, so it will never be executed if CHOLMOD is not available.

### DIFF
--- a/src/rulesets/LinearAlgebra/structured.jl
+++ b/src/rulesets/LinearAlgebra/structured.jl
@@ -1,6 +1,5 @@
 # Structured matrices
 using LinearAlgebra: AbstractTriangular
-using SparseInverseSubset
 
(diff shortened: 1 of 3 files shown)


ElOceanografo:
One thought...would it be worth defining alternate methods for these rrules for non-GPL builds? I can imagine a case where someone writes code depending on them in an evironment where CHOLMOD is available, then tries to run it on a non-GPL build and gets surprising/mysterious errors.

oxinabox:
Lets see if anyone complains.
Noone noticed these were broken if you didn't have GPL for quite a while

PR #773, opened 2024-01-24
Use handwritten rules for `zero` and `one`

MasonProtter:
Attempt to fix the problems raised in https://github.com/SciML/NeuralPDE.jl/pull/791 that was caused by https://github.com/JuliaDiff/ChainRules.jl/pull/770

I *think* this is the right way to generalize the scalar rule to any input type, and always return a `ZeroTangent()` instead of a `NoTangent()` like the old `@non_differentiable` rule did which was quite mathematically suspect.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.59.0"
+version = "1.59.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -6,32 +6,24 @@
 
 # `zero`
 
-function frule((_, Δ1), ::typeof(zero), x)
-    var"∂f/∂x" = ZeroTangent()
-    return (zero(x), Δ1 * var"∂f/∂x")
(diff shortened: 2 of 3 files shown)


MasonProtter:
Appears to have the same failure profile as https://github.com/JuliaDiff/ChainRules.jl/pull/774

oxinabox:
Why is this not:

```
function frule((_, _), ::typeof(zero), x)
    return (zero(x), ZeroTangent())
end

function rrule(::typeof(zero), x)
    zero_pullback(_) = (NoTangent(), ZeroTangent())
    return (zero(x), zero_pullback)
end

# `one`

function frule((_, _), ::typeof(one), x)
    return (one(x), ZeroTangent())
end

function rrule(::typeof(one), x)
    one_pullback(_) = (NoTangent(), ZeroTangent())
    return (one(x), one_pullback)
end
```

This extra stuff with projection and multiplication all should just cancel away for if the tangent is `ZeroTangent()`.

I agree the `NoTangent` before was mathematically wrong, and correct is `ZeroTangent()`, though because of how they act similarly, and under Zygote identically it worked out fine.

Thank you for playing wack-a-mole with this.

MasonProtter:
> Why is this not:

Mostly because I was copy-pasting and then tweaking the output from `@scalar_rule`, I can simplify it though.

oxinabox:
test failures on 1.0 are unrelated.

oxinabox (approved):

PR #776, opened 2024-01-27
Fix `UndefVarError: sparseinv not defined` test failures

### LINKED ISSUE #763: SparseInverseSubset.jl dependency causes issues on non-GPL Julia builds

christiangnrd:
`SparseInverseSubset` was not being imported in Julia >=1.7.

Introduced by #772. Closes #763?

### DIFF
--- a/src/rulesets/SparseArrays/sparsematrix.jl
+++ b/src/rulesets/SparseArrays/sparsematrix.jl
@@ -51,7 +51,8 @@ function rrule(::typeof(findnz), v::AbstractSparseVector)
 end
 
 if Base.USE_GPL_LIBS # Don't define rrules for sparse determinants if we don't have CHOLMOD from SuiteSparse.jl
-        
+    using SparseInverseSubset
+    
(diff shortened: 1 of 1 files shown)


oxinabox:
thanks!
There are other CI failures I am going to hunt down today, but this fixes one of them

PR #779, opened 2024-02-08
Make `getindex` rule work for `AxisArrays`

simsurace:
Due to the distinction between `Base.axes` and `AxisArrays.axes`, the existing rules did not work for `AxisArrays`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "1.61.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
@@ -20,6 +21,7 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 
 [compat]
 Adapt = "3.4.0, 4"
+AxisArrays = "0.4.7"
 ChainRulesCore = "1.20"
 ChainRulesTestUtils = "1.5"
 Compat = "3.46, 4.2"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.61.0"
+version = "1.61.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -130,7 +130,7 @@ It's unfortunate to close over `x`, but `similar(typeof(x), axes(x))` doesn't
 allow `eltype(dy)`, nor does it work for many structured matrices.
 """
 _setindex_zero(x::AbstractArray{<:Number}, dy, inds::Integer...) = fill!(similar(x, typeof(dy), axes(x)), false)
-_setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy), axes(x)), false)
+_setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy)), false)
 function _setindex_zero(x::AbstractArray, dy, inds::Integer...)
     # This allows for types which don't define zero (like Vector) and types whose zero special (like Tangent),
     # but always makes an abstract type. TODO: make it infer concrete type for e.g. vectors of SVectors
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -130,7 +130,8 @@ It's unfortunate to close over `x`, but `similar(typeof(x), axes(x))` doesn't
 allow `eltype(dy)`, nor does it work for many structured matrices.
 """
 _setindex_zero(x::AbstractArray{<:Number}, dy, inds::Integer...) = fill!(similar(x, typeof(dy), axes(x)), false)
-_setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy)), false)
+_setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy), axes(x)), false)
+_setindex_zero(x::AxisArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy), AxisArrays.axes(x)), false)
 function _setindex_zero(x::AbstractArray, dy, inds::Integer...)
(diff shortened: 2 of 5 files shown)


Project.toml
```diff
@@ -1,9 +1,10 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.61.0"
+version = "1.61.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
```
devmotion, line 7: IMO ChainRules should not depend on AxisArrays.

simsurace: I don't have any insight on the merits for or against this. But what is your suggestion?

devmotion: AFAICT it has been a general policy to not accept such dependencies, see e.g. https://github.com/JuliaArrays/FillArrays.jl/pull/153#issuecomment-900188437

src/rulesets/Base/indexing.jl
```diff
@@ -131,6 +131,7 @@ allow `eltype(dy)`, nor does it work for many structured matrices.
 """
 _setindex_zero(x::AbstractArray{<:Number}, dy, inds::Integer...) = fill!(similar(x, typeof(dy), axes(x)), false)
 _setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy), axes(x)), false)
```
devmotion, line 133: Is there a reason for why these are not just
```suggestion
_setindex_zero(x::AbstractArray{<:Number}, dy, inds::Integer...) = fill!(similar(x, typeof(dy)), false)
_setindex_zero(x::AbstractArray{<:Number}, dy, inds...) = fill!(similar(x, eltype(dy)), false)
```
AFAICT this would also fix the AxisArrays problem.

simsurace: This breaks existing tests. The problem is if you don't pass the axes, then you don't get a dense array.

devmotion: Which tests are broken? The two-arg method is even advised in the Julia docs: https://docs.julialang.org/en/v1/manual/methods/#Building-a-similar-type-with-a-different-type-parameter

Project.toml
simsurace: An extension to FillArrays is also out of the question?

src/rulesets/Base/indexing.jl
mcabbott: The 3-arg one removes structured matrices like Symmetric, iirc

Project.toml
devmotion: The PR was before extensions existed, so I have been thinking for a while one should try again with an extension. I managed to get in an extension on PDMats recently, so I think it seems likely that it would be approved.

src/rulesets/Base/indexing.jl
devmotion: Maybe we should just add special cases for these? At first glance, it doesn't seem very desirable to remove structure (as the AxisArrays case shows).

mcabbott: IMO the ideal situation is for `axes` to return more information in AxisArrays, alla https://github.com/mcabbott/AxisKeys.jl/pull/6 , since the relevant properties do belong to individual axes, not to the whole (like Symmetric). But we ran out of energy to fix things.

Project.toml
mcabbott: Making that FillArrays PR into an extension there would be great.

mcabbott:
What I thought should work is `plain_inds = Base.to_indices(x, inds)`, which should turn non-indices like Colon and also Symbols etc. into actual indices. Then the gradient can remain a plain array. This doesn't work at present, but perhaps something nearby does?

If the gradient is another AxisArray (as proposed here) then one question is what meaning its axis vector have. E.g. if I differentiate `x = AxisArray(rand(3), ax = [1.1, 2.2, 3.3])` with respect to the `2.2`, you might argue the gradient representation should be ` AxisArray(zeros(3), ax = [0, 1.0, 0])`, which is in conflict with the meaning here.

simsurace:
Is this something that people do (label axes with floats)?

mcabbott:
Certainly some people want AxisArrays to be a sort of DataFrame, with labels just identifying columns.

I think the structure is more generally useful for storing anything which varies along an axis, and passing this along, never using it to replace indexing. For instance keeping a probability vector associated with eachcol(matrix). 

No idea really which is more common.

The current behaviour with Zygote is this, which I think means you won't silently get wrong answers. But may get errors if you wish to combine the two functions:
```julia
julia> gradient(x -> x[1], AxisArray([1,2,3.0], aux=[4,5,6.0]))  # natural 
([1.0, 0.0, 0.0],)

julia> gradient(x -> AxisArrays.axes(x)[1][1], AxisArray([1,2,3.0], aux=[4,5,6.0]))  # structural
((data = nothing, axes = ((val = [1.0, 0.0, 0.0],),)),)
```

AxisKeys acquired some projection rules which seem to be designed only for the first case:
```
julia> gradient(x -> x[1], KeyedArray([1,2,3.0], aux=[4,5,6.0]))[1]
1-dimensional KeyedArray(...) with keys:
↓   3-element Vector{Float64}
And data, 3-element ChainRules.OneElement{Float64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}}:
 (4.0)  1.0
 (5.0)  0.0
 (6.0)  0.0

julia> gradient(x -> axiskeys(x)[1][1], KeyedArray([1,2,3.0], aux=[4,5,6.0]))
ERROR: MethodError: (::ChainRulesCore.ProjectTo{KeyedArray, @NamedTuple{data::ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}}}}, keys::@NamedTuple{aux::Vector{Float64}}}})(::ChainRulesCore.Tangent{KeyedArray{Float64, 1, NamedDimsArray{(:aux,), Float64, 1, Vector{Float64}}, Base.RefValue{Vector{Float64}}}, @NamedTuple{data::ChainRulesCore.NoTangent, keys::ChainRulesCore.Tangent{Base.RefValue{Vector{Float64}}, @NamedTuple{x::ChainRules.OneElement{Float64, 1, Tuple{Int64}, Tuple{Base.OneTo{Int64}}}}}}}) is ambiguous.
```

simsurace:
It would seem to me that the first use case, where the index is not differentiable (strings, symbols) is the more pressing one to tackle, as it is clear what the right behavior should be (i.e. identical to indexing with integers).

Could one of the maintainers please make a final decision as to whether this PR is going to be rejected based on the additional dependency? In that case, and if we can't make #780 work, I will open a PR for an extension over at AxisArrays.

mcabbott:
CR doesn't define rules for any packages, only standard lib. This isn't a whole `rrule`, but it seems depending on a packages to define a method is pretty much the same thing. So I don't think it is an acceptable solution.

If there's an easy way to tweak a rule to play nicely that's OK, and there are some tests involving packages. But most things should be handled by packages depending on ChainRulesCore, or pkg extensions.

Making AxisArrays opt out of the rule might work, as it must ultimately index the parent array, and probably that's the right time to call this rule? I'm still a little surprised that `Base.to_indices` doesn't produce valid indices; altering that logic might be another way.

PR #780, opened 2024-02-08
Fix `getindex` rule with AxisArrays (but break existing tests?)

devmotion:
This PR tries the suggestion in https://github.com/JuliaDiff/ChainRules.jl/pull/779#discussion_r1483664720 which apparently breaks some existing tests. I wonder what exactly is broken by these changes since the two-arg `similar` is quite standard and even recommended in the Julia docs: https://docs.julialang.org/en/v1/manual/methods/#Building-a-similar-type-with-a-different-type-parameter

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "1.61.0"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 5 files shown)


simsurace:
So it is only that one test I also saw break (I did not wait for all tests to finish running), though there may be more things breaking that aren't tested. Do you think this PR is worth pursuing further?

PR #781, opened 2024-02-09
remove ambig in nondiff cumprod(::Vector{AbstractBool}) on julia 1.6

oxinabox:
This doesn't reproduce locally for me.
And doesn't show up on never julia versions.
Possibly dispatch changed subtly to prevent this.
But lets see if it fixes it on CI

### DIFF
--- a/src/rulesets/Base/nondiff.jl
+++ b/src/rulesets/Base/nondiff.jl
@@ -34,6 +34,7 @@
 @non_differentiable circshift!(::AbstractArray{Bool}, ::Any)
 @non_differentiable conj(::AbstractArray{Bool})
 @non_differentiable conj!(::AbstractArray{Bool})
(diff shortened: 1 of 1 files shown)


oxinabox:
it fixed it., Yay.
Now lets see if someone reviews

devmotion (commented):
Looks good to me - the only question I have is whether you considered merging the `rrule`s in https://github.com/JuliaDiff/ChainRules.jl/blob/b3f9f9bcf7b7c228350232944cdce019d934f4bc/src/rulesets/Base/mapreduce.jl#L328-L374 as an alternative. As far as I can tell, this would be an alternative fix for the ambiguity?

mcabbott (approved):

oxinabox:
Merging this over #784 as it is passing CI already.
Which empirically suggests it is simpler

PR #782, opened 2024-02-09
Add missing subtract rule

nmheim:
So far there was only a rule for negation, this adds a rule for subtraction (two-argument `-`).

### DIFF
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -445,7 +445,7 @@ function rrule(::typeof(-), x::AbstractArray, y::AbstractArray)
     yproj = ProjectTo(y)
     function subtract_pullback(dy_raw)
         dy = unthunk(dy_raw)  # projs will otherwise unthunk twice
-        (NoTangent(), xproj(dy), yproj(-dy))
+        return (NoTangent(), xproj(dy), yproj(-dy))
     end
     return x - y, subtract_pullback
 end
--- a/src/rulesets/Base/arraymath.jl
+++ b/src/rulesets/Base/arraymath.jl
@@ -438,11 +438,11 @@ end
 ##### Subtraction
 #####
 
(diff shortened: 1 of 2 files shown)


src/rulesets/Base/arraymath.jl
```diff
@@ -434,6 +434,16 @@ function rrule(::typeof(-), x::AbstractArray)
...
 
+#####
+##### Subtraction
+#####
+
+frule((_, Δx, Δy), ::typeof(-), x::AbstractArray, y::AbstractArray) = x-y, Δx-Δy
+
+function rrule(::typeof(-), x::AbstractArray, y::AbstractArray)
+    subtract_pullback(dy) = (NoTangent(), dy, -dy)
+    return x-y, subtract_pullback
```
oxinabox: the autoformatter is probably complaining about spaces around binary operators

```suggestion
frule((_, Δx, Δy), ::typeof(-), x::AbstractArray, y::AbstractArray) = (x - y), (Δx - Δy)

function rrule(::typeof(-), x::AbstractArray, y::AbstractArray)
    subtract_pullback(dy) = (NoTangent(), dy, -dy)
    return x - y, subtract_pullback
```

nmheim: yep, fixed already!:)

nmheim:
@oxinabox the `rrule` for multi-arg `+` has some reshape related things in the pullback which I don't understand:
https://github.com/JuliaDiff/ChainRules.jl/blob/98f893f0f74ff9d852a66064c88dde3322871341/src/rulesets/Base/arraymath.jl#L454-L462
should those be added here as well?

nmheim:
> yes, in particular it should be done with `ProjectTo` I think (c.f. #782)

Ahh, that makes sense! thank you!

oxinabox:
Shall we add a test that exercises the  `ProjectTo` path like
`test_rrule(-, [2.0, 1.0],  [3.0 + 2.0im, 4.0+5im)`
?

PR #783, opened 2024-02-09
Use ProjectTo in multiarg +

oxinabox:
I believe the `reshape(..., axes(a))` was basically a early stand before we had the more fully features ProjectTo code.


mcabbott:
Duplicate of https://github.com/JuliaDiff/ChainRules.jl/pull/578 . That one has some evil tests, which should perhaps be copied here... do they now pass?

mcabbott (dismissed):

PR #784, opened 2024-02-09
merge cumprod rules to fix ambig on julia 1.6

oxinabox:
Suggested by @devmotion  in https://github.com/JuliaDiff/ChainRules.jl/pull/781
should resolve same issue.
I am not sure which is cleaner

### DIFF
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -325,7 +325,13 @@ end
 ##### `cumprod`
 #####
 
-function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer=1)
+function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer=nothing)
+    if isnothing(dims) && ndims(x)==1
+        dims=1
+    else
+        throw(UndefKeywordError(:dims)
+    end
+
     y = cumprod(x; dims=dims)
     project_x = ProjectTo(x)
     function cumprod_pullback(dy_raw)
--- a/src/rulesets/Base/mapreduce.jl
+++ b/src/rulesets/Base/mapreduce.jl
@@ -325,52 +325,37 @@ end
 ##### `cumprod`
 #####
 
-function rrule(::typeof(cumprod), x::AbstractVector{<:Real}; dims::Integer=1)
-    y = cumprod(x; dims=dims)  # does nothing unless dims == 1
-    project_x = ProjectTo(x)
-    function cumprod_pullback_1(dy_raw)
-        dy = unthunk(dy_raw)
-        dx_thunk = InplaceableThunk(
-            dx -> if dims == 1
-                ∇cumprod!(dx, x, dy, y)
-            else
-                dx .+= dy
-            end
-            ,
-            @thunk project_x(if dims == 1
-                ∇cumprod(x, dy, y)
-            else
-                dy
-            end)
-            )
-        return (NoTangent(), dx_thunk)
-    end
-    return y, cumprod_pullback_1
-end
-
 function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer)
     y = cumprod(x; dims=dims)
     project_x = ProjectTo(x)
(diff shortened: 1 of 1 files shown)


devmotion (commented):
I think I'm slightly in favour of this PR, since it reduces the number of methods and `cumprod(::AbstractVector)` is forwarded to the keyword argument version: https://github.com/JuliaLang/julia/blob/63e95d47e692352fac5541b6ee835b573f4c4898/base/accumulate.jl#L227

Unfortunately, the diff is significantly larger than in #781 - and it requires a few additional changes to the tests since there's no `rrule(cumprod, ::AbstractVector)` anymore (so one has to set `dims=1` explicitly in these tests).

src/rulesets/Base/mapreduce.jl
```diff
@@ -325,52 +325,37 @@ end
...
-            else
-                dy
-            end)
-            )
-        return (NoTangent(), dx_thunk)
-    end
-    return y, cumprod_pullback_1
-end
-
 function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer)
```
oxinabox, line 328:
```suggestion
function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer=1)
```

src/rulesets/Base/mapreduce.jl
```diff
@@ -325,52 +325,37 @@ end
...
-                dy
-            end)
-            )
-        return (NoTangent(), dx_thunk)
-    end
-    return y, cumprod_pullback_1
-end
-
-function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer)
+function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer=1)
```
devmotion, line 328: Is it OK to put a default value for `dims` even though there's no default value in base? https://github.com/JuliaLang/julia/blob/63e95d47e692352fac5541b6ee835b573f4c4898/base/accumulate.jl#L193

oxinabox:
its actually fine for the `rrule` to always default dims to 1. (So I have made that change)
Since:
 - in the vector case this is what is wanted and is the correct default
 - in the nonvector case then a keyword argument is required by the primal, so the default will never be hit.

src/rulesets/Base/mapreduce.jl
oxinabox: yes, it is see:
https://github.com/JuliaDiff/ChainRules.jl/pull/784#issuecomment-1936107183

mcabbott:
> in the nonvector case then a keyword argument is required by the primal, so the default will never be hit.

What does this mean? The primal computation is done inside the rrule, and does use the default. 

I think that will mean `gradient(x -> sum(abs, cumprod(x)), [1 2; 3 4.])` is not an error, whereas without gradient it would be.

That said, I have no memory of why there are two rrules. Besides this issue there might have been performance concerns?

oxinabox:
> I think that will mean gradient(x -> sum(abs, cumprod(x)), [1 2; 3 4.]) is not an error, whereas without gradient it would be.

That is a point.
I had been thinking we could rely on the primal computation being correct.
But I guess we can not.

We will have to do somehting more complex like set the rrule as `dims=nothing`
then if `ndims(x)==1` set it to `dims=1`

devmotion:
> We will have to do somehting more complex like set the rrule as dims=nothing

Can't we just omit the default value and adjust the tests?

src/rulesets/Base/mapreduce.jl
oxinabox, line 328:
```suggestion
function rrule(::typeof(cumprod), x::AbstractArray{<:Real}; dims::Integer=nothing)
    if isnothing(dims) && ndims(x)==1
        dims=1
    else
        throw(UndefKeywordError(:dims)
    end

```

mcabbott:
My vote is that https://github.com/JuliaDiff/ChainRules.jl/pull/781 is much simpler. 

Besides worrying about error cases, I think this kind of change would want to be checked very carefully for type stability or performance regressions.

PR #785, opened 2024-02-12
support arbitary types for vect -- use zero_tangent for _instantiate_zeros

oxinabox:
`zero_tangent` correctly handles zeros for things that do not have `zero` implemented, by using it in `_instantitate_zeros` we can support anything being passed to `vect`

Without this rule, anything with a vector being constructed is very annoying, makes Diffractor very sad.

Not sure if should go further and also knock out some of the array cases.
Especially arrays of arrays. But that can always be a follow up

### DIFF
--- a/src/rulesets/Base/array.jl
+++ b/src/rulesets/Base/array.jl
@@ -31,7 +31,7 @@ end
 
 @non_differentiable Base.vect()
 
-function frule((_, ẋs...), ::typeof(Base.vect), xs::Number...)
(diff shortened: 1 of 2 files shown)


oxinabox:
The follow up to this would be to relax the requirement on the `frule` for `vect` to accept anything, not just `Number`s.
Without it, anything with a vector being constructed is very annoying

oxinabox:
Failure on x86 julia 1.6 is that occational bad alloc we see, but have never tracked down

oscardssmith (approved):

PR #786, opened 2024-02-20
correct test muladd so that complex type is included

not-physicist:
Generate the random numbers of the appropriate type. I assume the line should be repeated, not sure...


mcabbott (approved):
Thanks! Looks like a typo, and tests pass.

PR #793, opened 2024-05-08
Mark `hasmethod` as non-diff

DhairyaLGandhi:


ChrisRackauckas:
Failures are nightly and a linear algebra segfault. Unrelated.

PR #795, opened 2024-05-20
Add rrule for NamedTuple merge

oxinabox:
Based on https://github.com/LuxDL/Lux.jl/pull/659
but with some fixes up to use ChainRulesCore types.

Initial PR is not inferable.
I suspect i will need to rewrite this as a generated funcition to make it infer.

@Avik-pal do you want to test this branch against your use case, with it deleted from Lux?

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -291,3 +291,25 @@ function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(task_local_storage
     end
     return y, task_local_storage_pullback
 end
+
+
+####
+#### merge
+####
+
+function rrule(::typeof(merge), nt1::NamedTuple{F1}, nt2::NamedTuple{F2}) where {F1, F2}
+    y = merge(nt1, nt2)
+    function merge_pullback(dy)
+        dnt1 = Tangent{typeof(nt1)}(;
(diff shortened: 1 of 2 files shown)


oxinabox:
What i want is to do 
```julia
_getproperty_kwexpr(key) = :($key = getproperty(dy, $key))
@generated function rrule(::typeof(merge), nt1::NamedTuple{F1}, nt2::NamedTuple{F2}) where {F1, F2}
    quote
        y = merge(nt1, nt2)
        function merge_pullback(dy)
            dnt1 = Tangent{typeof(nt1)}(; $(map(_getproperty_kwexpr, setdiff(F1, F2))...))
            dnt2 = Tangent{typeof(nt2)}(; $(map(_getproperty_kwexpr, F2)...))
            return (NoTangent(), dnt1, dnt2)
        end
        merge_pullback(dy::AbstractThunk) = merge_pullback(unthunk(dy))
        merge_pullback(x::AbstractZero) = (NoTangent(), x, x)
        
        return y, merge_pullback
    end
end
```
But generated functions are not allowed to have bodys that create closures

oxinabox:
Final solution was tricky to get to but clean

oxinabox:
@Avik-pal can you review?

avik-pal:
Let me trigger the tests for SciMLSensitivity with this PR

avik-pal:
@oxinabox this should be good to go, downstream tests work

avik-pal (approved):

Issue #788, opened 2024-03-23
Make the rrule for 3-arg dot lazy 


mohamed82008:
Currently this line https://github.com/JuliaDiff/ChainRules.jl/blob/9f1817a22404259113e230bef149a54d379a660b/src/rulesets/LinearAlgebra/dense.jl#L38 leads to a loss of structure because it constructs a dense matrix from 2 vectors. We should make this a lazy outer product (like in https://github.com/SciML/LinearSolve.jl/pull/484) such that the `m x n` matrix's co-tangent only requires `O(m + n)` memory. See https://discourse.julialang.org/t/zygote-jl-how-to-get-the-gradient-of-sparse-matrix/59067/12 for the potential gain and suggested implementation. If the idea gets a green light, I will open a PR.

oxinabox:
I guess the key question is are we willing to add LazyArrays as a dependency?

gdalle:
Related: https://discourse.julialang.org/t/zygote-much-slower-than-jax-for-automatic-differentiation-of-energy/114239/5

oxinabox:
Since this seems to have come up twice in two different contexts i think we can say that this is worth it and we should add LazyArrays as a dependency.
@mohamed82008  are you able to make the PR?

mohamed82008:
I am able to yes. I will do it in the weekend unless someone wants to beat me to it. It's basically exactly this implementation https://discourse.julialang.org/t/zygote-jl-how-to-get-the-gradient-of-sparse-matrix/59067/12 plus unit tests.

albertomercurio:
Hello, are there any news about this? I would like to implement reverse diff to `dot(x, A, x)` function with `A` sparse, but at the moment it converts the `A` matrix into a dense one.

mohamed82008:
https://github.com/JuliaDiff/ChainRules.jl/pull/796

PR #796, opened 2024-05-25
make 3-arg dot rrule partially lazy 

mohamed82008:
This addresses #788. I had to remove the projection to make it work otherwise I get the following error due to a missing projection method. Projecting the lazy array to a dense array when `A` is dense partially defeats the purpose of this PR so I am leaving it up to the review process to decide what to do here. I can define a projection method if that's preferred.

```julia
julia> show(err)
1-element ExceptionStack:
MethodError: no method matching (::ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}})(::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, typeof(*), Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(*), Tuple{Float64, Vector{Float64}}}, Adjoint{Float64, Vector{Float64}}}})

Closest candidates are:
  (::ChainRulesCore.ProjectTo{T})(::ChainRulesCore.NotImplemented) where T
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/zgT0R/src/projection.jl:121
  (::ChainRulesCore.ProjectTo)(::ChainRulesCore.Thunk)
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/zgT0R/src/projection.jl:124
  (::ChainRulesCore.ProjectTo{AbstractArray})(::Number)
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/zgT0R/src/projection.jl:253
  ...

Stacktrace:
 [1] (::ChainRules.var"#1966#1970"{Adjoint{Float64, Vector{Float64}}, Float64, Vector{Float64}, ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}})()
   @ ChainRules ~/.julia/dev/ChainRules/src/rulesets/LinearAlgebra/dense.jl:39
 [2] unthunk
   @ ~/.julia/packages/ChainRulesCore/zgT0R/src/tangent_types/thunks.jl:204 [inlined]
 [3] wrap_chainrules_output
   @ ~/.julia/packages/Zygote/nsBv0/src/compiler/chainrules.jl:110 [inlined]
 [4] map
   @ ./tuple.jl:293 [inlined]
 [5] map
   @ ./tuple.jl:294 [inlined]
 [6] wrap_chainrules_output
   @ ~/.julia/packages/Zygote/nsBv0/src/compiler/chainrules.jl:111 [inlined]
 [7] ZBack
   @ ~/.julia/packages/Zygote/nsBv0/src/compiler/chainrules.jl:211 [inlined]
 [8] (::Zygote.var"#75#76"{Zygote.ZBack{ChainRules.var"#dot_pullback#1968"{Vector{Float64}, Matrix{Float64}, Vector{Float64}, Vector{Float64}, ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}}}}, ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}, ChainRulesCore.ProjectTo{AbstractArray, @NamedTuple{element::ChainRulesCore.ProjectTo{Float64, @NamedTuple{}}, axes::Tuple{Base.OneTo{Int64}}}}}}})(Δ::Float64)
   @ Zygote ~/.julia/packages/Zygote/nsBv0/src/compiler/interface.jl:91
 [9] top-level scope
   @ REPL[6]:1
```

This is related to the discussion in https://github.com/FluxML/Zygote.jl/issues/1507.


mohamed82008:
This is also technically a breaking change if we don't project.

oxinabox:
We may need to teach LazyArrays about projections, so that they project lazily?

mohamed82008:
I like the idea of lazy projections. Maybe that would also provide a way to opt out of the projection at the gradient level by calling a `get_projection_parent` function to undo the last projection.

albertomercurio:
hello, are there any updates on this? I really need to compute the gradient of `dot(x, A, y)`, with `x` and `y` dense vectors and `A` sparse matrix. It currently converts the matrix into a dense one, which is very inefficient for large matrices.

mohamed82008:
I suppose you could just define your own `mydot` function for now and define a rule following this PR.

oxinabox:
problem is it is generally required to project the tangents back down onto the tangent space once it is computed.
Explaining why is a little complex but it is probably in the docs.
But it shows up with things like a the tangent to a real number that has been multiplied by a complex number must be real,
 and structured matrixes (like diagonal matrixes) need to also follow that structure.
 
 Anyway, that operation needs to also be done lazily if we want to return a lazy array from the 3 arg dot rrule.
 
 Which can be done, but it requires overloading `ProjectTo` on `LazyArray`

PR #797, opened 2024-05-30
make `rrule` return identical pullback for `zero` as for `one`

nsajko:
Could be a minor compilation latency and/or type stability win for some uses.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -4,6 +4,11 @@
 @scalar_rule copysign(y, x) (ifelse(signbit(x)!=signbit(y), -one(y), +one(y)), NoTangent())
 @scalar_rule transpose(x) true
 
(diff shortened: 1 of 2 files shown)


nsajko:
The test failures seems quite unrelated, though serious.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.66.0"
+version = "1.66.1"
```
oxinabox, line 3: this isn't a bug fix, so should be 
```suggestion
version = "1.67.0"
```

oxinabox (approved):

oxinabox:
the 1.6 x86 thing is a transient failure that has been shopwing up on and off

PR #799, opened 2024-06-10
exclude julia 1.6 x86 from CI

oxinabox:
This has been stocastically failing more often than not. The signal to noise ratio is too low and i have just been ignoring it for like a year. It is some julia or JuliaInterpetter bug being hit


sethaxen (approved):
LGTM

PR #805, opened 2024-08-27
Replace nightly tests with pre-release tests

sethaxen:
The nightly CI runs have [failed persistently for 9 months](https://github.com/JuliaDiff/ChainRules.jl/actions/runs/7012621446). As a result, these are really not useful. More useful would be to test against pre-releases, and we should actually expect CI to pass for pre-releases. That's what this PR does.


sethaxen:
(On a personal note, since I created the JuliaNightly workflow, GitHub notifies me every single time it fails, which is at least daily, and I cannot unsubscribe from this without unsubscribing from the entire repo. Hopefully this improves things)

sethaxen:
There are some test failures on v1.11.0-rc3, specifically type-inference issues for the rrules of `sortslices(::AbstractMatrix; dims::Integer)` and `norm(::AbstractMatrix, ::Real)`.  If we merge this right now, CI will be marked as failing on main. Ideally we would fix these failures either here or shortly after merging, but after spending some time on this, it's not at all clear to me why these are no longer type-inferrable. Alternatively, we could just skip inferrability checks here.

oxinabox:
I thought we already disabled infer-ability tests on nightly.

is the problem that we are not using `CRTU.@maybe_inferred` for some tests?
Or did we not set the `CHAINRULES_TEST_INFERRED` environment variable?

PR #808, opened 2024-09-16
Fix computing of array element type for `∇eachslice`

### LINKED ISSUE #807: `∇eachslice` causes `UndefRefError` when input array contains references
https://github.com/JuliaDiff/ChainRules.jl/blob/30f9b121e69fe20f3a176e92e3ceb91f53606f8c/src/rulesets/Base/indexing.jl#L264-L282

1. `similar` makes `dx` an uninitialized array.
2. When `dys[i]` is zero, presumably the intent of `_zero_fill!(slice)` is to populate the array with zeros.
3. `_zero_fill!` on an uninitialized array calls `map!(zero, dx, dx)`
4. `map!` attempts to map `zero` onto the elements of `dx`, which are `#undef` - error

So this branch never works when the input array has references rather than inline values.

BioTurboNick:
Fixes #807

Added test as well. Two lines really belong in ChainRulesCore, not sure what the process for that is.

### DIFF
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -267,7 +267,7 @@ function ∇eachslice(dys_raw, x::AbstractArray, vd::Val{dim}) where {dim}
     if i1 === nothing  # all slices are Zero!
         return _zero_fill!(similar(x, float(eltype(x)), axes(x)))
     end
-    T = promote_type(eltype(dys[i1]), eltype(x))
+    T = promote_type(eltype.(dys)...)
     # The whole point of this gradient is that we can allocate one `dx` array:
     dx = similar(x, T, axes(x))
     for i in axes(x, dim)
@@ -282,8 +282,7 @@ function ∇eachslice(dys_raw, x::AbstractArray, vd::Val{dim}) where {dim}
 end
 ∇eachslice(dys::AbstractZero, x::AbstractArray, vd::Val{dim}) where {dim} = dys
 
-_zero_fill!(dx::AbstractArray{<:Number}) = fill!(dx, zero(eltype(dx)))
-_zero_fill!(dx::AbstractArray) = map!(zero, dx, dx)
+_zero_fill!(dx::AbstractArray) = fill!(dx, zero(eltype(dx)))
 
 function rrule(::typeof(∇eachslice), dys, x, vd::Val)
     function ∇∇eachslice(dz_raw)
--- a/src/rulesets/Base/indexing.jl
+++ b/src/rulesets/Base/indexing.jl
@@ -285,8 +285,9 @@ end
 _zero_fill!(dx::AbstractArray) = fill!(dx, zero(eltype(dx)))
 
 # Belong in ChainRulesCore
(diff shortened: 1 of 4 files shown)


BioTurboNick:
Other than the formatter, all the failures seem to be due to the fact that the changes mean Julia currently isn't fully inferring the output anymore. I'll see if there's anything to do about that.

BioTurboNick:
Switching to `Base.promote_eltype` addresses the inference issue, but that's not a public API in base.

BioTurboNick:
Huh. I'm not sure why the inference test is failing on 1.6 only. It's inferring `Union{NoTangent, Vector{Float64})` instead of just Vector{Float64}'.

src/rulesets/Base/indexing.jl
```diff
@@ -282,8 +282,12 @@ function ∇eachslice(dys_raw, x::AbstractArray, vd::Val{dim}) where {dim}
...
 ∇eachslice(dys::AbstractZero, x::AbstractArray, vd::Val{dim}) where {dim} = dys
 
-_zero_fill!(dx::AbstractArray{<:Number}) = fill!(dx, zero(eltype(dx)))
-_zero_fill!(dx::AbstractArray) = map!(zero, dx, dx)
+_zero_fill!(dx::AbstractArray) = fill!(dx, zero(eltype(dx)))
+
+# Belong in ChainRulesCore
+Base.promote_rule(T::Type{<:Number}, S::Type{<:AbstractZero}) = T
+Base.promote_rule(T::Type{<:AbstractZero}, S::Type{<:Number}) = S
+Base.eltype(::Type{NoTangent}) = NoTangent
```
simeonschaub, line 290: Can you make a PR to ChainRulesCore?

simeonschaub (changes requested):
I believe the original intent with the design of `similar` was to support cases like `Vector{Vector{Float64}}` where `zero(Vector{Float64})`. That of course doesn't work though since `similar` will just produce `#undef`s here making this fail anyways.

Looks good to me minus the methods that need to be moved to ChainRulesCore. Could you make a PR there (ideally with some unit tests), so we can tag a new version there and then change the compat here.

src/rulesets/Base/indexing.jl
BioTurboNick: Done!

simeonschaub:
Just needs to drop the changes already in https://github.com/JuliaDiff/ChainRulesCore.jl/pull/682 and bump the compat for ChainRulesCore

BioTurboNick:
done and done

simeonschaub:
Hmm, looks like the format suggestions are not working properly (Maybe because it's on a fork?) If you look at the CI logs, you can see that it complains about the formatting of line 257 in the tests. Once that's fixed this looks good to go!

devmotion:
> (Maybe because it's on a fork?) 

Yes, due to restricted permissions for forks the suggestions can only be shown in the logs but not as actual suggestions in the PR.

simeonschaub (approved):

PR #819, opened 2025-02-21
Use `_return_type` from Base, not from `Core.Compiler`

serenity4:
Starting from https://github.com/JuliaLang/julia/pull/56409, we [no longer have](https://github.com/JuliaLang/julia/blob/e026affc3ad49ac199cf22ce05aeb412d50fcb21/base/Base_compiler.jl#L305) `Base._return_type === Core.Compiler._return_type`, and therefore `Core.Compiler._return_type` calls the wrong function on nightly, triggering an error. This PR should ensure a more stable access to this binding.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ChainRules"
 uuid = "082447d4-558c-5d27-93f4-14fc19e9eca2"
-version = "1.72.2"
+version = "1.72.3"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


mcabbott:
Similar changes in https://github.com/FluxML/NNlib.jl/pull/618 used `Core.Compiler.return_type` instead of `Base._return_type`. Do you know which is better?

serenity4:
I think they are equivalent (at least they are in 1.11 and nightly), we can go with `Core.Compiler.return_type` as it reads slightly more stable without an underscore.

serenity4:
I believe test failures are unrelated, if happy with the change feel free to merge, and it'd be fantastic if a new version could be tagged as well.

mcabbott (approved):

mcabbott:
Thanks! Will tag it now.

PR #824, opened 2025-09-29
Handle Base.hypot with more than 2 arguments

moble:
Currently, there are rules for `hypot(x)` and `hypot(x, y)`, but `Base` defines `hypot(x, y, xs...)`, which is not handled by ChainRules.  The two-argument case is in `fastmath_able`, but `Base.FastMath` only supports the two-argument method, so this PR adds separate forward and reverse rules specifically for 3 arguments, with a `Vararg` on the end, to handle the remaining possibilities.  These new rules are just mild generalizations of the two-argument rules.

### DIFF
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -265,13 +265,12 @@ end
     end
 
     @testset "hypot(x, y, z, xs...)" begin
-        test_frule(hypot, 1.0, 2.0, 3.0)
-        test_rrule(hypot, 1.0, 2.0, 3.0)
-        test_frule(hypot, 1.0, 2.0, 3.0, 4.0)
-        test_rrule(hypot, 1.0, 2.0, 3.0, 4.0)
-        test_frule(hypot, 1.0+5.0im, 2.0+6.0im, 3.0+7.0im)
-        test_rrule(hypot, 1.0+5.0im, 2.0+6.0im, 3.0+7.0im)
-        test_frule(hypot, 1.0+5.0im, 2.0+6.0im, 3.0+7.0im, 4.0+8.0im)
-        test_rrule(hypot, 1.0+5.0im, 2.0+6.0im, 3.0+7.0im, 4.0+8.0im)
+        for n in (3, 4)
+            for sig in Iterators.product(ntuple(_->(Float64, ComplexF64), n)...)
+                args = randn.(sig)
+                test_frule(hypot, args...)
+                test_rrule(hypot, args...)
+            end
+        end
     end
 end


src/rulesets/Base/base.jl
```diff
@@ -119,6 +119,40 @@ function rrule(::typeof(hypot), z::Complex)
...
+function frule(
+    (_, Δx, Δy, Δz, Δxs...),
+    ::typeof(hypot),
+    x::T,
+    y::T,
+    z::T,
+    xs::Vararg{T, N}
+) where {T<:Union{Real,Complex}, N}
+    Ω = hypot(x, y, z, xs...)
+    n = ifelse(iszero(Ω), one(Ω), Ω)
```
devmotion, line 133: Generally, `one(x)` is not of the same type as `x`:
```suggestion
    n = ifelse(iszero(Ω), oneunit(Ω), Ω)
```

src/rulesets/Base/base.jl
```diff
@@ -119,6 +119,40 @@ function rrule(::typeof(hypot), z::Complex)
...
+# Note that `hypot` with two arguments has rules in `fastmath_able.jl`
+
+function frule(
+    (_, Δx, Δy, Δz, Δxs...),
+    ::typeof(hypot),
+    x::T,
+    y::T,
+    z::T,
+    xs::Vararg{T, N}
+) where {T<:Union{Real,Complex}, N}
```
devmotion, line 131: It's probably not necessary that all arguments are of the same type?
```suggestion
    x::Union{Real,Complex},
    y::Union{Real,Complex},
    z::Union{Real,Complex},
    xs::Union{Real,Complex}...,
)
```

src/rulesets/Base/base.jl
```diff
@@ -119,6 +119,40 @@ function rrule(::typeof(hypot), z::Complex)
...
+    Ω = hypot(x, y, z, xs...)
+    n = ifelse(iszero(Ω), one(Ω), Ω)
+    ∂Ωxyz = realdot(x, Δx) + realdot(y, Δy) + realdot(z, Δz)
+    ∂Ωxs = sum(realdot(xi, Δxi) for (xi, Δxi) in zip(xs, Δxs); init=zero(∂Ωxyz))
+    ∂Ω = (∂Ωxyz + ∂Ωxs) / n
```
devmotion, line 136: Arguably a bit simpler:
```suggestion
    ∂Ω = sum(map(realdot, (x, y, z, xs...), (Δx, Δy, Δz, Δxs...))) / n
```

src/rulesets/Base/base.jl
```diff
@@ -119,6 +119,40 @@ function rrule(::typeof(hypot), z::Complex)
...
+    return Ω, ∂Ω
+end
+
+function rrule(
+    ::typeof(hypot),
+    x::T,
+    y::T,
+    z::T,
+    xs::Vararg{T, N}
+) where {T<:Union{Real,Complex}, N}
```
devmotion, line 146: Same here:
```suggestion
    x::Union{Real,Complex},
    y::Union{Real,Complex},
    z::Union{Real,Complex},
    xs::Union{Real,Complex}...,
)
```

src/rulesets/Base/base.jl
```diff
@@ -119,6 +119,40 @@ function rrule(::typeof(hypot), z::Complex)
...
+    Ω = hypot(x, y, z, xs...)
+    function hypot_pullback(ΔΩ)
+        c = real(ΔΩ) / ifelse(iszero(Ω), one(Ω), Ω)
+        return (NoTangent(), c * x, c * y, c * z, (c * xi for xi in xs)...)
```
devmotion, line 150:
```suggestion
        return (NoTangent(), c * x, c * y, c * z, map(xi -> c * xi, xs)...)
```

PR #825, opened 2025-09-29
Remove old warning about map

### LINKED ISSUE #798: `rrule` for `map` for tuples is outdated with respect to JuliaLang/julia#42216
https://github.com/JuliaDiff/ChainRules.jl/blob/9dd39bdbb66819ad26d19ebbe62b76c1ede56f66/src/rulesets/Base/base.jl#L243-L273

The referenced Julia issue is resolved: JuliaLang/julia#42216

moble:
Closes #798

There already was a test that hits this, so I've just removed the error message.

### DIFF
--- a/src/rulesets/Base/base.jl
+++ b/src/rulesets/Base/base.jl
@@ -249,7 +249,7 @@ function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(map), f::F, xs::Tu
     y = map(first, hobbits)
     num_xs = Val(length(xs))
     paddings = map(x -> ntuple(Returns(NoTangent()), (length(x) - length_y)), xs)
-    @static if VERSION < v"1.10.0-alpha1"
+    @static if VERSION < v"1.10.0-DEV.1194"
         all(isempty, paddings) || @error """map(f, xs::Tuple...) does not allow mismatched lengths in Julia <1.10!
         But its `rrule` does."""
     end
--- a/test/rulesets/Base/base.jl
+++ b/test/rulesets/Base/base.jl
@@ -253,7 +253,7 @@ end
         test_rrule(map, make_two_vec, (4.0, 5.0 + 6im), check_inferred=false)
         test_rrule(map, Multiplier(rand() + im), Tuple(rand(3)), check_inferred=false)
 
-        if VERSION >= v"1.10.0-alpha1"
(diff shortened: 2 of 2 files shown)


devmotion (commented):
The fix (https://github.com/JuliaLang/julia/commit/bb3177205a5abc2f29978611eace3f287b1618dd) is only available in Julia >= 1.10.0-alpha1 but ChainRules currently supports Julia >= 1.6.0.

Thus this PR either has to update the minimum Julia version to 1.10.0 (which IMO should be done in a separate PR though since it might entail other changes such as updates to CI and removal of `VERSION` specific branches) or the check has to be removed only in more recent Julia versions.

moble:
Good point.  I've kept the check now, but wrapped it in a static version check.

src/rulesets/Base/base.jl
```diff
@@ -249,8 +249,10 @@ function rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(map), f::F, xs::Tu
     y = map(first, hobbits)
     num_xs = Val(length(xs))
     paddings = map(x -> ntuple(Returns(NoTangent()), (length(x) - length_y)), xs)
-    all(isempty, paddings) || @error """map(f, xs::Tuple...) does not allow mistmatched lengths!
-        But its `rrule` does; when JuliaLang/julia #42216 is fixed this warning should be removed."""
+    @static if VERSION < v"1.10.0-alpha1"
```
devmotion, line 252: A bit more precise:
```suggestion
    @static if VERSION < v"1.10.0-DEV.1194"
```

test/rulesets/Base/base.jl
```diff
@@ -253,8 +253,8 @@ end
         test_rrule(map, make_two_vec, (4.0, 5.0 + 6im), check_inferred=false)
         test_rrule(map, Multiplier(rand() + im), Tuple(rand(3)), check_inferred=false)
 
-        if try map(+, (1,), (2,3)); true catch e; false end
-            # True when https://github.com/JuliaLang/julia/issues/42216 has been fixed
+        if VERSION >= v"1.10.0-alpha1"
```
devmotion, line 256:
```suggestion
        if VERSION >= v"1.10.0-DEV.1194"
```

devmotion (approved):

PR #837, opened 2026-03-31
Add rules for sparse symmetric matrices: dot, lmul, rmul.

samuelsonric:
This PR adds rules for the operations

- `A * Y`
- `X * A`
- `dot(x, A, y)`

where `A` is a `Symmetric` (or `Hermitian`)-wrapped sparse matrix. The new rules are much faster then the defaults, although some combinations will perform poorly until [this issue](https://github.com/JuliaSparse/SparseArrays.jl/issues/685) is resolved.

# Example

```
using ChainRules, LinearAlgebra, SparseArrays

A = Hermitian(sprand(10000, 10000, .001), :L)

x = rand(10000)
y = rand(10000)
Δz = 0.1

_, pullback = ChainRules.rrule(dot, x, A, y)
@time map(ChainRules.unthunk, pullback(Δz));
```

- old: `0.123707 seconds (48 allocations: 769.704 MiB, 16.02% gc time)`
- new: `0.000881 seconds (23 allocations: 1.844 MiB)`


samuelsonric:
Related: https://github.com/JuliaDiff/ChainRulesCore.jl/pull/707

PR #840, opened 2026-06-14
Fix matrix-function primal type for Hermitian{<:Real} on Julia 1.12

CarloLucibello:
## Summary

Julia 1.12 changed real-valued matrix functions of `Hermitian{<:Real}` (e.g. `exp`, `log`, `sqrt`, `cos`, `sin`, ...) to return `Hermitian` instead of `Symmetric`:

```julia
julia> exp(Hermitian(ones(2,2))) |> typeof
Hermitian{Float64, Matrix{Float64}}   # was Symmetric on Julia ≤ 1.11
```

`_matfun` unconditionally wrapped the real-eltype result as `Symmetric`, so the `rrule`/`frule` primal no longer matched `f(A)`. This surfaced as a type mismatch downstream — see FluxML/Zygote.jl#1592:

```julia
julia> Zygote.pullback(exp, Hermitian(ones(2,2)))[1] |> typeof
Symmetric{Float64, Matrix{Float64}}   # should be Hermitian
```

## Fix

In `_matfun`, when the input is real, wrap the result in the same `Symmetric`/`Hermitian` type as the input **when the output is real**. The complex-valued case (e.g. `acosh` of eigenvalues `< 1`) is still wrapped as `Symmetric`, which is what Julia returns for both real `Symmetric` and `Hermitian` inputs. Behavior on Julia < 1.12 is unchanged (always `Symmetric`).

The inference assertions in the matrix-function tests are updated to expect the input wrapper type on 1.12+.

## Testing

`Pkg.test(test_args=["symmetric"])` passes on Julia 1.12 (5499 tests, 0 errors).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


devmotion:
Ref https://github.com/JuliaDiff/ChainRules.jl/pull/833

As mentioned in https://github.com/JuliaDiff/ChainRules.jl/pull/833#issuecomment-3829601001, AD doesn't guarantee that primal values are exactly identical to the values obtained without AD. So it's unclear why `Hermitian{<:Real}` would be better than `Symmetric{<:Real}`.

